PR #1, opened 2021-04-12
Initial commit

gumaerc:
#### What's this PR do?
Currently, the OCW website is built out of 2 main components.  `ocw-www` is a Hugo site with its own webpack build that contains content that builds the home page, search page, testimonials, promos, etc.  `ocw-course-hugo-theme` is a Hugo theme used to build the course sites in the website.  The problem with this is that these sites share a lot of the same styles and some templates, so changes to similar components have to be done twice in both repos.

This PR is the initial commit to a new repo combining the responsibilities of `ocw-www` and `ocw-course-hugo-theme` into a single module based Hugo theme that can be inherited by any Hugo site.

Read the updated [`README.md`](https://github.com/mitodl/ocw-hugo-theme/blob/1cd4142fc0fca34b756ed3538a7851608c494a73/README.md) file to familiarize yourself with how this works

- Clone [`ocw-www` ](https://github.com/mitodl/ocw-www) and check out the `cg/content-only` branch
- Follow the updated readme and:
  - Install dependencies and configure the site based on the instructions
  - When you create your `.env` file, set `EXTERNAL_SITE_PATH` to the path to `ocw-www` that you cloned earlier
  - Edit the `go.mod` in your `ocw-www` folder to override it to point at your local copy of `ocw-hugo-theme` by adding the following lines, replacing the paths with yours:
```
replace github.com/mitodl/ocw-hugo-theme/base-theme => /path/to/ocw-hugo-theme/base-theme

replace github.com/mitodl/ocw-hugo-theme/www => /path/to/Code/ocw-hugo-theme/www

replace github.com/mitodl/ocw-hugo-theme/course=> /path/to/Code/ocw-hugo-theme/course
```
  - Spin up the OCW home page locally using `npm start` and verify that everything looks alright when compared to https://ocwnext.odl.mit.edu/
  - Visit search and verify that everything looks like the prod site
  - Spin up a course site using `npm run start:course` and verify that the course site looks as it does on https://ocwnext.odl.mit.edu/


LICENSE
```diff
@@ -0,0 +1,37 @@
...
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+Site content Copyright (c) Gilbert Strang. 18.06 Linear Algebra. Spring 2010.
```
noisecapella, line 32: Should this be in the license file?

gumaerc: Good question, although I'm not sure.  This license is copied over from `ocw-www`.

noisecapella:
Can you change admin of this repo to `odl-engineering`?

gumaerc:
@noisecapella all set

noisecapella:
What is the use case for `EXTERNAL_SITE_PATH` if we have `www` in the same repo?

gumaerc:
@noisecapella This is an interesting question and maybe hints that I didn't name the variable well.  What that serves to do is to point to an external Hugo "site" to spin up locally using the theme.  The idea is that `ocw-www` can still exist as another repo with *only* content in it.  Check out the `cg/content-only` branch of `ocw-www` and you will see what it contains, but it should just be:

 - `content`
 - `static`
 - `config.toml`
 - `go.mod`

The idea is that the templates, JS and CSS all live in this repo and then the only thing in other repos that use this as a theme would be the barebones Hugo "site" that imports this as a module.  In the case of `ocw-www` that's the content markdown and static files related specifically to the content (testimonial images, etc.).

noisecapella:
Can we add those files to `www`? It seems like if we are checking in content there or here, we might as well check it in here to simplify things

gumaerc:
@noisecapella I think that's reasonable, as if people don't want the content when importing the `www` theme as a module, they can just omit

```
  [[module.mounts]]
    source = "content"
    target = "content"
```

from their `config.toml` file.  @Ferdi what do you think about this?

gumaerc:
For what it's worth, I tested this by adding content to my local copy of `ocw-hugo-theme` in the `www/content` folder.  The content does not show up when including the theme as a module unless you explicitly include it using the `config.toml` lines I noted above.

README.md
```diff
@@ -1 +1,184 @@
...
+- [Chrome](https://chrome.google.com/webstore/detail/moesif-origin-cors-change/digfbfaphojjndkpccljibejjbppifbc?hl=en-US#:~:text=Moesif%20Origin%20%26%20CORS%20Changer&text=This%20plugin%20allows%20you%20to%20send%20cross%2Ddomain%20requests.&text=This%20plugin%20allows%20you%20to%20send%20cross%2Ddomain%20requests%20directly,Allow%2DOrigin%20set%20to%20*.)
+- [Firefox](https://addons.mozilla.org/en-US/firefox/addon/moesif-origin-cors-changer1/)
+### Environment variables
+
+For development you'll need to set a few environment variables. The
+`.env.example` file will get you started.  Here are the variables and what they do:
+
+| Variable | Example | Description |
+| --- | --- | --- |
+| `SEARCH_API_URL` | `http://localhost:8063/api/v0/search/` | A URL to an `open-discussions` search API to fetch results from|
```
pdpinch, line 42: Why does the theme have a dependency on `SEARCH_API_URL`? I would expect this would only be used by the site that contains `/search`

README.md
```diff
@@ -1 +1,184 @@
...
+| `OCW_STUDIO_BASE_URL` | `http://localhost:8043/` | A URL of an instance of [`ocw-studio`](https://github.com/mitodl/ocw-studio) to fetch home page content from |
```
pdpinch, line 43: Why does the the theme have a dependency on `OCW_STUDIO_BASE_URL`? I would expect this would only be used by a site like ocw-www which contains the home page.

README.md
```diff
@@ -1 +1,184 @@
...
+| `EXTERNAL_SITE_PATH` | `~/Code/ocw-www/site/` | A path to a Hugo site with content (Pages, Notifications, Promos, Testimonials) to use for local development |
```
pdpinch, line 44: If I'm building a single course, would I set `EXTERNAL_SITE_PATH` to the path of the course repo?

README.md
```diff
@@ -1 +1,184 @@
...
+| `OCW_TO_HUGO_PATH` | `~/Code/ocw-to-hugo/` | A path to a local override of [`ocw-to-hugo`](https://github.com/mitodl/ocw-to-hugo), a library used to generate Hugo markdown content from parsed OCW JSON exports from the legacy site |
```
pdpinch, line 45: Why does the theme depend on ocw-to-hugo?

gumaerc: When running a course site locally, if you specify a path to `ocw-to-hugo`, the `start_course.sh` script will automatically download the parsed JSON for the course you specify and convert it to markdown to spin up the site with.  It's not necessarily required, but is a quick way to get up and running if (as a developer) you're trying to work on changes to the theme and see them reflected on an existing course.  This will change of course once we moonlight `ocw-to-hugo`.  You can still use the course theme in an external folder of content by including it in your `config.toml`, and we should add a way to do that here without using `ocw-to-hugo`.

README.md
gumaerc: As you've surmised, this is only necessary if you are working on changes to the search page and want to see search results in your locally running site.

README.md
gumaerc: In the `home.html` template in the `www` theme, the `home_course_cards.html` and `ocw_news.html` partials use the Hugo `getJSON` functionality to directly fetch these items using `OCW_STUDIO_BASE_URL`.  If it's not set, `getJSON` will throw an error during the Hugo build.  I agree that this should not be a dependency in spinning up a `www` site, and they can be ignored by adding `ignoreErrors = ["error-remote-getjson"]` to the `config.toml` file or nesting the calls to `getJSON` in a conditional.

README.md
gumaerc: Currently course sites can only be spun up using `ocw-to-hugo`, but I can add the ability to spin them up from `EXTERNAL_SITE_PATH` if `OCW_TO_HUGO_PATH` is not set.

README.md
Ferdi: This should be fixed.

README.md
gumaerc: Actually, I just tested this and you can spin up a course site without `ocw-to-hugo`.  You do need to add a couple of files to your course content folder for it to work though:

`config/_default/config.toml`:

```
baseURL = "http://localhost:3000/"
languageCode = "en-us"
title = "MIT OpenCourseWare"

[module]
  [[module.imports]]
    path = "github.com/mitodl/ocw-hugo-theme/base-theme"
  [[module.imports]]
    path = "github.com/mitodl/ocw-hugo-theme/course"
  [[module.mounts]]
    source = "assets"
    target = "assets"
  [[module.mounts]]
    source = "data"
    target = "data"
  [[module.mounts]]
    source = "layouts"
    target = "layouts"
  [[module.mounts]]
    source = "static"
    target = "static"
```

`go.mod`:
```
module github.com/mitodl/1-00-introduction-to-computers-and-engineering-problem-solving-spring-2012

go 1.16

replace github.com/mitodl/ocw-hugo-theme/base-theme => /home/gumaerc/Code/ocw-hugo-theme/base-theme

replace github.com/mitodl/ocw-hugo-theme/course => /home/gumaerc/Code/ocw-hugo-theme/course
```

After placing these files in your course content folder and running `npm start` in the theme, a local dev site will spin up pointing at that course site content.  With the way our content currently comes through legacy OCW though, if you are developing changes to the theme it's much more simple to just have it use `ocw-to-hugo` to get the content.

README.md
noisecapella: @Ferdi this change means that the site will sometimes be missing information on the frontend depending on whether ocw-studio's API call succeeded. The course cards and news items will appear and disappear seemingly at random

noisecapella:
@Ferdi @pdpinch There are a couple questions I have:
 - should we be calling the ocw-studio API for news and card content in ocw-hugo-themes, or should that happen in ocw-www? 
 - If we are calling the ocw-studio API here, should we have only markdown files in ocw-www or should those markdown files be moved into this repo as well?

README.md
Ferdi: @noisecapella I understand. I thought we were going to have a flag that would make it behave differentely in prod vs dev.  The problem I have with it is that a theme cannot depend on an external service.

Ferdi:
@noisecapella 

> should we be calling the ocw-studio API for news and card content in ocw-hugo-themes, or should that happen in ocw-www

Since we are using getJson from the templates then I think it means that the call happens in the theme 

> If we are calling the ocw-studio API here, should we have only markdown files in ocw-www or should those markdown files be moved into this repo as well

I think md files are content, not templates, so the belong in the "content" section of a hugo project. For now that content section lives in ocw-www.  

I think of it this way.  A hugo project looks like (more or less):

```
project-root
    content
    data
    themes
         my theme
```

Right now for ocw-www project, everything lives in the same repository.  The only thin this PR does is takes "my theme" out of that repository and moves it here.   Content is still there, and so would be "data" (if any)

build_scripts/start_course.sh
```diff
@@ -0,0 +1,22 @@
...
+
+set -euo pipefail
+ENV_FILE=.env
+
+if test -f "$ENV_FILE"; then
+  export $(cat .env | xargs)
+fi
+
+if [[ -z "${OCW_TEST_COURSE}" ]]; then
+  echo "OCW_TEST_COURSE not set"
```
noisecapella, line 11: We should exit with an error code here

build_scripts/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -euo pipefail
+
+ENV_FILE=.env
```
noisecapella, line 5: Can we remove `ENV_FILE` here and in the other script files, and reference `.env` directly instead? Also, I think we should error if `.env` does not exist since it should be part of a working setup

build_scripts/start_site.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -euo pipefail
+ENV_FILE=.env
+
+if test -f "$ENV_FILE"; then
+  export $(cat .env | xargs)
+fi
+
+if [[ -z "${EXTERNAL_SITE_PATH}" ]]; then
+  # Run the site without external content
```
noisecapella, line 11: I think we should error if `EXTERNAL_SITE_PATH` is not set. A developer should have both ocw-www and this repo checked out, and have the environment variables set up to point to the directory

noisecapella (changes requested):

www/config.toml
```diff
@@ -2,6 +2,9 @@ baseURL = ""
 languageCode = "en-us"
 title = "MIT OpenCourseWare"
 
+# Make sure errors with getJSON don't stop the build
+ignoreErrors = ["error-remote-getjson"]
```
noisecapella, line 6: It looks like the intention is to only do this for development: https://github.com/mitodl/ocw-hugo-themes/pull/1/files#r621530790

Is there a way to configure this option only in development environments?

noisecapella (changes requested):
I have a related question: We have both ocw-www/config.toml and ocw-hugo-themes/www/config.toml, and from this line in my output it looks like it's reading both files: `Watching for config changes in /home/george/Projects/ocw-www/config.toml, /home/george/Projects/ocw-hugo-themes/base-theme/config.toml, /home/george/Projects/ocw-hugo-themes/www/config.toml, /home/george/Projects/ocw-www/go.mod`

Should we remove one or the other, or do we need both?

noisecapella:
There's a race condition which happens when `npm start` is run, where hugo and webpack are started at the same time but hugo is looking for webpack files to create links for the javascript files. These files exist the second time hugo is started but they refer to the files created the first time webpack was started. I'm not sure it's worth holding up the PR for so I'll file an issue

base-theme/config.toml
```diff
@@ -0,0 +1,11 @@
...
+languageCode = "en-us"
+languageLang = "en"
+title = "MIT OpenCourseWare"
+relativeURLs = false
+enableRobotsTXT = true
+
+# RSS, categories and tags disabled for an easy start
+# See configuration options for more details:
+# https://gohugo.io/getting-started/configuration/#toml-configuration
+disableKinds = ["RSS", "taxonomy", "taxonomyTerm"]
```
noisecapella, line 11: There were some configuration options to adjust cache `getJSON` calls so that we have new news items fetched when builds happen. Why are those options missing here? https://github.com/mitodl/ocw-www/blob/main/site/config.toml#L13

noisecapella (changes requested):

www/config.toml
gumaerc: @Ferdi mentioned that the theme cannot depend on `ocw-studio` being up and running.  I do see your point though, that if there were an intermittent outage of `ocw-studio` and a full site build happened during that then there would be no new courses or news.  I think we can pull this off with environment based configuration as described [here](https://gohugo.io/getting-started/configuration/#configuration-directory) but then the production build would still completely stop if `ocw-studio` was unavailable.  Do we want that?

base-theme/config.toml
gumaerc: I seemed to have missed those when I was re-writing the config files so I'll add them back.  I think they belong in `ocw-www`'s configuration though and not the theme.

www/config.toml
gumaerc: > There's a race condition which happens when `npm start` is run, where hugo and webpack are started at the same time but hugo is looking for webpack files to create links for the javascript files. These files exist the second time hugo is started but they refer to the files created the first time webpack was started. I'm not sure it's worth holding up the PR for so I'll file an issue

Yea, I've experienced this and wasn't able to find a quick and easy solution but I think it's worth looking into outside of the scope of this PR.

base-theme/config.toml
gumaerc: > There's a race condition which happens when `npm start` is run, where hugo and webpack are started at the same time but hugo is looking for webpack files to create links for the javascript files. These files exist the second time hugo is started but they refer to the files created the first time webpack was started. I'm not sure it's worth holding up the PR for so I'll file an issue

Yea, I've experienced this and wasn't able to find a quick and easy solution but I think it's worth looking into outside of the scope of this PR.

gumaerc:
> There's a race condition which happens when `npm start` is run, where hugo and webpack are started at the same time but hugo is looking for webpack files to create links for the javascript files. These files exist the second time hugo is started but they refer to the files created the first time webpack was started. I'm not sure it's worth holding up the PR for so I'll file an issue

Yea, I've experienced this and wasn't able to find a quick and easy solution but I think it's worth looking into outside of the scope of this PR.

build_scripts/build.sh
gumaerc: I'm doing this for local development start scripts, but I think the build scripts need to be able to handle those env varibales just being set in the environment by salt so the .env should be optional there.

www/config.toml
gumaerc: Also @noisecapella to answer your original question, config files are actually not needed in the theme at all, although I put some examples under `example_configs`.  The config should live in the project utilizing the theme.

gumaerc:
@noisecapella I made some changes based on your feedback and it's ready for another look.

build_scripts/build.sh
noisecapella: Makes sense

noisecapella:
If we are removing `config.toml` files, should we remove the relevant files in the `example_configs` directory too?

gumaerc:
> If we are removing `config.toml` files, should we remove the relevant files in the `example_configs` directory too?

No, they are necessary for running course sites locally at the moment.  `ocw-to-hugo` should be generating config that uses the course theme here so ultimately this won't be necessary, but it is at the moment.  Also, they serve as an example of how to include the theme as a module in your project.

www/config.toml
noisecapella: I don't think we ever want the production site to look like this:
Screenshot from 2021-04-28 10-18-49

A build which failed due to an API outage from ocw-studio can be restarted, and in the meantime the site is functioning as normal. We cannot just decide not to rely on ocw-studio because we need the OCW news feed in order to create content for the cards on the frontpage, and the news items updates on a regular basis. We also rely on ocw-studio for course information for the carousel cards. The choice is to break the build or to display an error message on the production site. @Ferdi @pdpinch any thoughts?

noisecapella (approved):

Issue #17, opened 2021-05-04
site does not build properly with unset baseUrl


gumaerc:
### Steps to Reproduce

 - Clone `ocw-www` and check out the `cg/remove-dev-baseurl` branch
 - Configure `ocw-hugo-themes` to point at your local copy of `ocw-www` by setting the path to it in `EXTERNAL_SITE_PATH`
 - Run `npm run build` and inspect the output (`ocw-www/dist`)

### Expected Behavior

The site builds properly with an assumed `baseUrl` relative too the root of the site

### Actual Behavior

Links to webpack resources do not render properly

PR #18, opened 2021-05-04
check if site.BaseURL is set before trying to use it

gumaerc:
#### Pre-Flight checklist

- [x] Testing

Closes https://github.com/mitodl/ocw-hugo-themes/issues/17

#### What's this PR do?
This is related to the recent work we have done to separate "theme" and "content."  A default `baseUrl` of `http://localhost:3000` was inadvertently set in the default config for `ocw-www`, causing all builds with `baseUrl` unset to assume this was the value. 
 If you remove this value, links to webpack resources do not render properly at all.  This PR adds some conditional checks to the `webpack_url.html` and `site_root_url.html` to make sure that `site.BaseURL` is set before using it.

 - Clone `ocw-www` and check out the `cg/remove-dev-baseurl` branch
 - Configure `ocw-hugo-themes` to point at your local copy of `ocw-www` by setting the path to it in `EXTERNAL_SITE_PATH`
 - Run `npm run build` and inspect the output (`ocw-www/dist`)
 - Ensure that Webpack CSS / JS in `index.html` is properly linked

### Additional info
This `ocw-www` PR is related and defaults the `baseUrl` to `/`: https://github.com/mitodl/ocw-www/pull/112

### DIFF
--- a/base-theme/layouts/partials/site_root_url.html
+++ b/base-theme/layouts/partials/site_root_url.html
@@ -6,5 +6,5 @@
     {{- printf "%s://%s/%s" $baseUrl.Scheme $baseUrl.Host (strings.TrimPrefix "/" .) -}}
   {{- end -}}
 {{- else -}}
-  {{- . -}}
+  {{- printf "/%s" . -}}
 {{- end -}}
--- a/base-theme/layouts/partials/site_root_url.html
+++ b/base-theme/layouts/partials/site_root_url.html
@@ -1,4 +1,4 @@
-{{- if isset site "BaseURL" -}}
+{{- if not (eq site.BaseURL "") -}}
   {{- $baseUrl := urls.Parse site.BaseURL -}}
   {{- if or (eq $baseUrl.Scheme nil) (eq $baseUrl.Host nil) -}}
     {{- printf "%s/%s" (strings.TrimSuffix "/" site.BaseURL) (strings.TrimPrefix "/" .) -}}
@@ -6,5 +6,5 @@
     {{- printf "%s://%s/%s" $baseUrl.Scheme $baseUrl.Host (strings.TrimPrefix "/" .) -}}
   {{- end -}}
 {{- else -}}
-{{- . -}}
+  {{- . -}}
 {{- end -}}
--- a/base-theme/layouts/partials/site_root_url.html
+++ b/base-theme/layouts/partials/site_root_url.html
@@ -1,6 +1,10 @@
-{{- $baseUrl := urls.Parse site.BaseURL -}}
-{{- if or (eq $baseUrl.Scheme nil) (eq $baseUrl.Host nil) -}}
+{{- if isset site "BaseURL" -}}
+  {{- $baseUrl := urls.Parse site.BaseURL -}}
+  {{- if or (eq $baseUrl.Scheme nil) (eq $baseUrl.Host nil) -}}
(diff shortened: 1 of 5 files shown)


base-theme/layouts/partials/site_root_url.html
```diff
@@ -1,6 +1,10 @@
-{{- $baseUrl := urls.Parse site.BaseURL -}}
-{{- if or (eq $baseUrl.Scheme nil) (eq $baseUrl.Host nil) -}}
+{{- if not (eq site.BaseURL "") -}}
```
noisecapella, line 1: Can we use [errorf](https://gohugo.io/functions/errorf/) to error if `baseURL` is missing or empty?

noisecapella (changes requested):

base-theme/layouts/partials/site_root_url.html
gumaerc: @noisecapella An empty `baseURL` is actually acceptable, in that case it's basically assumed that it is `/`.  This check is in place because `urls.Parse` throws a hard error if it's passed a string that it cannot disassemble.

gumaerc:
@noisecapella I left some clarification above about why I don't think we should generate an error if `baseURL` is unset, because that's actually an acceptable condition.  However, I did change up the else statement in `site_root_url.html` to prefix anything passed in with `/` to guarantee it's relative to the root of the site.

gumaerc:
@noisecapella Okay so after thinking about this some more, I think I've come up with something that makes sense.  I added some comments to `site_root_url.html` to explain the logic but basically:

 - If `site.BaseURL` is blank or starts with a `/`, simply return the string passed in relative to `/`
 - if `site.BaseURL` starts with `http`, attempt to disassemble the URL and return the string relative to the reconstructed root URL of the site
 - Otherwise, it's an invalid `site.BaseURL` and we should error
 
What do you think?

noisecapella:
For this case: "If site.BaseURL is blank or starts with a /, simply return the string passed in relative to /", can we error? If the user wants to have a base url of `/`, they should specify `baseURL = "/"` in their `config.toml`. By erroring when there is a missing configuration value, we help people by pointing the problem out to them right away.

gumaerc:
@noisecapella In Hugo (and in general with static site generators) you're supposed to be able to leave the `BaseURL` property blank.  It's not a required argument.  This makes the resulting site "portable" in that you can redeploy it anywhere you want and all URLs will just be relative to `/` by default.

gumaerc:
@noisecapella I made some changes to this and the associated `ocw-www` PR, let me know what you think

gumaerc:
@noisecapella Just realized I didn't link the associated `ocw-www` PR that updates the config to remove the dev `baseUrl`, so I linked that in the description just now.

noisecapella:
Is `site_root_url.html` meant to cut out the path part of `baseURL`? I have `baseUrl = "http://localhost.org/courses/18-01-single-variable-calculus-fall-2006/"` in my config and I added a line to `baseof.html`: `{{ partial "site_root_url.html" "/end.html" }}`, but the output is `http://localhost.org/end.html`. Just checking to make sure this behavior is intentional

gumaerc:
@noisecapella Yes, that's its exact purpose.  In the case of course builds on the live server, we pass in a `baseUrl` with something like `https://ocwnext.mit.edu/courses/6-334-power-electronics-spring-2007/` (for example) and the course theme is built and deposited at the root of the site.  So in that case, the URL it needs to get is something like `https://ocwnext.mit.edu/course.css`.

noisecapella (approved):

Issue #20, opened 2021-05-05
pdfjs not working with injected course builds


gumaerc:
### Steps to Reproduce

Bulid the OCW Next website with a subset of example courses:
 - Clone `ocw-www`
 - Configure this repo to point at `ocw-www` by setting `EXTERNAL_SITE_PATH` to the path to it
 - Run `npm run build`
 - Clone `ocw-to-hugo` and follow the readme to download and convert the courses in `example_courses.json` to `private/output`
 - Set the following environment variables, adjusting the paths to match your system:
```
OCW_TO_HUGO_OUTPUT_DIR=/path/to/ocw-to-hugo/private/output
COURSE_OUTPUT_DIR=/path/to/ocw-www/dist/courses
COURSE_BASE_URL=/courses
```
 - Run `./build_scripts/build_all_courses.sh`
 - Inspect the output of the build at `ocw-www/dist`

### Expected Behavior

PDFJS's web viewer is deployed as static files to the site root at `/pdfjs` and nowhere else

### Actual Behavior

PDFJS is not deployed at the site root, and is instead deployed alongside every single course build

PR #21, opened 2021-05-05
fix pdfjs

gumaerc:
#### Pre-Flight checklist

- [x] Testing

Closes https://github.com/mitodl/ocw-hugo-themes/issues/20

#### What's this PR do?
Currently, the PDFJS web viewer is deposited as static JS / HTML / CSS under `course/static/pdfjs`.  This has the side effect in the full site build of including a copy of these files with every single course build, which is not the intention.  PDFJS should only be copied over once to the root of the site.  This PR moves the static files over to `www/static/pdfjs` (including the `build` folder, which was missing) and these files are now only included as part of a course build if you start a course locally for development using `npm run start:course`.  The static files are simply copied into the staging area using:

```
  mkdir -p private/ocw-to-hugo-output/$OCW_TEST_COURSE/static/pdfjs
  cp -r www/static/pdfjs/* private/ocw-to-hugo-output/$OCW_TEST_COURSE/static/pdfjs
```
This way the PDFJS artifacts are included at the root of the OCW Next full site bulid so they can be referenced by courses and a copy is not included along with every course being built and you can still test the PDFJS viewer pages when running a single course site locally

 - Follow the "steps to reproduce" in [this](https://github.com/mitodl/ocw-hugo-themes/issues/20) issue and ensure that copies of the PDFJS artifacts are not deposited into courses
 - Follow the readme to learn how to start up a single course using `npm run start:course` and verify that the PDFJS viewer still functions


abeglova (commented):
LGTM

PR #27, opened 2021-05-11
Accessibility changes for carousel

noisecapella:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

Fixes https://github.com/mitodl/ocw-www/issues/100

#### What's this PR do?
Makes some accessibility changes to the carousel:
 - fade in and out instead of scroll
 - left and right buttons at the side
 - no autoscrolling

Try out the promo carousel on the homepage. You may want to checkout https://github.com/mitodl/ocw-www/pull/115 for ocw-www first to have alt text for the images. 
 - Nothing should animate automatically. When you trigger an animation, it should end after a couple seconds and stay inanimate until you trigger it explicitly again. 
 - You should see a fade transition effect instead of seeing cards animate to the side.
 - You should be able to tab between the left and right buttons and press enter to activate them. You should not be able to activate the carousel indicator buttons at the bottom with a keyboard. I wasn't able to get this to work effectively, and the bootstrap example code does not do so either.
 - On mobile, you should be able to swipe the carousel and see a similar fade transition. It is awkward but you can do this in the responsive view in Firefox, and there's likely a similar feature in Chrome.
 - Mobile transitions are turned off so you should not see a fade at all. It should just instantly switch to the next or previous card.

#### Screenshots
Desktop:
Screenshot from 2021-05-11 09-25-35

Mobile:
Screenshot from 2021-05-11 09-25-19

### DIFF
--- a/www/assets/css/home.scss
+++ b/www/assets/css/home.scss
@@ -458,7 +458,24 @@ $promo-carousel-height: 200px;
 
 .promo-carousel {
   max-height: $promo-carousel-height;
+  margin: 0 auto;
   margin-bottom: 30px;
+  max-width: $max-content-width;
+
+  a {
+    cursor: pointer;
+    .material-icons {
+      font-size: 60px;
+      font-weight: 200;
+      color: $text_gray;
+    }
+  }
+
+  @include media-breakpoint-down(sm) {
+    .carousel-item {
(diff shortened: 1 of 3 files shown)


annagav (approved):
Looks good 👍

PR #47, opened 2021-05-17
allow launching of an externally converted course

gumaerc:
#### Pre-Flight checklist
- [x] Testing

None

#### What's this PR do?
The `npm run start:course` command currently only supports downloading data from S3.  This PR allows you to optionally bypass this for working with data locally converted in `ocw-to-hugo` manually.  This is useful for development if you are working on testing a PR with a specific branch of `ocw-to-hugo`.

 - Clone `ocw-to-hugo`, read the readme if you've never used it before and convert any number of courses
 - See the updated readme and example env, set your variables to point at your local copy of `ocw-to-hugo` and set `DOWNLOAD` to `0`
 - Set `OCW_TEST_COURSE` to one of the courses you downloaded in `ocw-to-hugo`
 - Run `npm run start:course`
 - Verify that you don't see any log entries saying that a course is being downloaded, and ensure that the course site is available at http://localhost:3000


build_scripts/start_course.sh
```diff
@@ -9,20 +9,30 @@ else
   exit 1
 fi
 
+# If the DOWNLOAD variable isn't set, default it to 0
```
noisecapella, line 12: It says 1 in the if block

build_scripts/start_course.sh
```diff
@@ -9,20 +9,30 @@ else
...
+    cp -r $OCW_TO_HUGO_OUTPUT_DIR/$OCW_TEST_COURSE private/ocw-to-hugo-output/
+  fi
   cp example_configs/example.course.go.mod private/ocw-to-hugo-output/$OCW_TEST_COURSE/go.mod
   cp example_configs/example.course.config.toml private/ocw-to-hugo-output/$OCW_TEST_COURSE/config/_default/config.toml
   mkdir -p private/ocw-to-hugo-output/$OCW_TEST_COURSE/static/pdfjs
   cp -r www/static/pdfjs/* private/ocw-to-hugo-output/$OCW_TEST_COURSE/static/pdfjs
   cd private/ocw-to-hugo-output/$OCW_TEST_COURSE
   hugo mod get -u
-  hugo server -p 3000 --bind 0.0.0.0 --renderToDisk
+  hugo server -p 3000 --bind 0.0.0.0 --renderToDisk --gc
```
noisecapella, line 37: What does `--gc` do here? Should we have this flag in other scripts too?

noisecapella:
Functionality works great

build_scripts/start_course.sh
gumaerc: It forces garbage collection and was just something I was using to troubleshoot.  I don't think it's necessary, as my issue was elsewhere.  This kind of all sprang from testing out the "other versions" functionality.  Since a reverse lookup needs to be made on `master_subject` if you only convert one course in `ocw-to-hugo` and run that locally, then `ocw-to-hugo` sees the master subject set on the course and generates the other versions section, but no other versions appear because `ocw-to-hugo` doesn't have the context of the other courses.  Being able to point to an already converted external course would allow you to convert a batch of courses and then point at one specifically from that batch to spin up so things like the master subject reverse lookup function properly.

build_scripts/start_course.sh
gumaerc: Good catch, the code here is correct and the comment is not.  At least I think it makes sense that if someone new were working on this, they might not have an external copy of `ocw-to-hugo` so download using the dependency version should be the default.

noisecapella (approved):

PR #48, opened 2021-05-18
Update some headers to remove accessibility warning

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Part of https://github.com/mitodl/ocw-www/issues/103

#### What's this PR do?
Changes some header tags to divs to workaround searchlight warnings

Page should look identical to what it looks like on main

### DIFF
--- a/www/layouts/home.html
+++ b/www/layouts/home.html
@@ -98,7 +98,7 @@
       <div class="d-flex flex-column justify-content-center">
         <h2>GIVE NOW</h2>
         <h2>Your Donation Makes a Difference</h2>
-        <h4 class="font-weight-normal">MIT OpenCourseWare is supported through the generosity of people like you, who believe that unlocking knowledge can empower minds.</h4>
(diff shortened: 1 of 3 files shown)


www/layouts/home.html
```diff
@@ -98,7 +98,7 @@ <h2>Next Generation OpenCourseWare</h2>
       <div class="d-flex flex-column justify-content-center">
         <h2>GIVE NOW</h2>
         <h2>Your Donation Makes a Difference</h2>
```
pdpinch, line 100: Are we allowed to have two `<h2>` tags?

noisecapella: It looks like they only discourage multiple header-element tags for `h1`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

I think we are all over the place with our header element tags so it may be best to just convert them all to span or div tags with the equivalent styling so the design doesn't change. I can do that in a future PR

mbertrand (approved):
👍

PR #74, opened 2021-05-27
Add search role and mark search area with aria-live

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Part of https://github.com/mitodl/ocw-www/issues/107

#### What's this PR do?
Mark search section as search landmark and set aria-live=polite so screenreaders notice changes to the feed

I'm not sure how to test the accessibility changes since I don't have a screenreader except the plugin for Chrome which seems low quality. When the page refreshes it only reads off the number of search results. Getting to scroll through the search results is also difficult. But you can look at the source and verify that the HTML has been changed, and read the accessibility docs on the aria [search](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role) role and [aria-live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) regions.

The color in search result cards has been changed to make greys darker for better contrast, and I also made a similar change to home page course cards.

#### Screenshots

Screenshot from 2021-05-28 11-38-05

Screenshot from 2021-05-28 11-26-00

### DIFF
--- a/www/assets/css/home.scss
+++ b/www/assets/css/home.scss
@@ -346,14 +346,14 @@ $searchbox-height: 50px;
 
     .course-level {
       text-transform: uppercase;
-      color: $medium-gray;
+      color: $dark-gray;
     }
 
     .course-card-content {
       font-size: 12px;
 
       .card-label {
-        color: $medium-gray;
+        color: $dark-gray;
       }
     }
   }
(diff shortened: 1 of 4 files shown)


annagav (approved):
👍

pdpinch (commented):
👍

PR #113, opened 2021-06-16
Fix course info expander

### LINKED ISSUE #104: Course info expander isn't collapsing
### Steps to Reproduce

Go to https://ocwnext.odl.mit.edu/courses/18-01-single-variable-calculus-fall-2006/ and click the "Course Info" header. You should be able to toggle it open and closed but it will only open.

I don't see any console warnings or errors, but it's very likely the problem is in `course_expander.js`

In this screenshot, the caret should be pointing to the right but it's pointing down instead.
Screenshot from 2021-01-27 16-41-27

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Fixes #104 

#### What's this PR do?
Fixes the course info expander

Start the server on course `6-00sc-introduction-to-computer-science-and-programming-spring-2011`. You should be able to toggle the "Course info >" link and also the "Read More" link. Both links should toggle correctly to expand the course info or course description sections.

### DIFF
--- a/course/layouts/course/course_home.html
+++ b/course/layouts/course/course_home.html
@@ -24,13 +24,13 @@
         </div>
       </div>
       {{ partial "mobile_nav_toggle.html" . }}
-      <div class="col-lg-8 course-description {{ if $shouldCollapseContent }}collapsed{{ end }}">
+      <div class="col-lg-8 course-description expand-container {{ if $shouldCollapseContent }}collapsed{{ end }}">
         <h4 class="font-weight-bold course-description-title">Course Description</h4>
         <div class="mb-1 description">{{- .Content -}}</div>
         <div class="mb-3">
           {{ if $shouldCollapseContent }}
             <button class="expand-link" aria-expanded="false">
-              <span class="text">Read More</span>
+              <span class="read-more-text">Read More</span>
             </button>
           {{ end }}
         </div>
--- a/course/assets/course.js
+++ b/course/assets/course.js
@@ -12,5 +12,5 @@ import { initCourseInfoExpander } from "./js/course_expander"
 $(document).ready(() => {
   initPdfViewers()
   initDesktopCourseInfoToggle()
-  initCourseInfoExpander()
+  initCourseInfoExpander(document)
 })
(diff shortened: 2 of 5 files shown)


pdpinch:
Any idea why this regressed? Is there anything we can do avoid it happening again?

noisecapella:
I don't know why it regressed. We weren't specific enough with our CSS classes before so that should be addressed in this PR. We didn't have any unit tests for this before and I added a bunch. However this is a hugo template so we can test the javascript but we don't have an easy way of making sure HTML changes don't break that over time.

noisecapella:
I think it would be worth the effort to integrate hugo templates into our test framework to do integration tests even with the added complexity since so many end users will see these pages. But it's a nontrivial amount of work similar to our selenium tests in micromasters, though I don't think we would actually need to use selenium

gumaerc (approved):
:+1: This mostly looks good, I noted one issue that could either be fixed here on in another PR relating to accessibility.

course/layouts/course/course_home.html
```diff
@@ -24,13 +24,13 @@
         </div>
       </div>
       {{ partial "mobile_nav_toggle.html" . }}
-      <div class="col-lg-8 course-description {{ if $shouldCollapseContent }}collapsed{{ end }}">
+      <div class="col-lg-8 course-description expand-container {{ if $shouldCollapseContent }}collapsed{{ end }}">
         <h4 class="font-weight-bold course-description-title">Course Description</h4>
         <div class="mb-1 description">{{- .Content -}}</div>
         <div class="mb-3">
           {{ if $shouldCollapseContent }}
             <button class="expand-link" aria-expanded="false">
```
gumaerc, line 32: When tabbing through the elements in this page, the Read More link becomes focused but does not receive an outline.  The enter button works to toggle the expander, but it's not clear that "Read More" is selected.

noisecapella: I couldn't figure out how to fix this via CSS so it's probably better to address in a separate PR. However I'm not sure it's an accessibility issue since a screenreader would still be able to focus even if the outline isn't rendering

PR #119, opened 2021-06-17
fix infinite scroll issue

### LINKED ISSUE #114: infinite scroll doesn't display all results
### Steps to Reproduce

Visit https://ocwnext.odl.mit.edu/search/?q=%22quantum+mechanics%22 and continuously scroll to the bottom.  Open an inspector and compare the last `article` element's `aria-labelledby` property to the total amonut of results listed at the top of the page

### Expected Behavior

The page renders all of the results that are supposed to be returned by the search terms

### Actual Behavior

An inconsistent amount of search results are actually rendered

### Screenshot or Screencast

image

aliceriot:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #114

#### What's this PR do?

This fixes the infinite scroll issue we were seeing by applying two fixes:

- ensuring that the state update for new search result data is always applied as an update to the most up-to-date data
- fixing the logic for the `loaded` prop passed to `useCourseSearch` so that only one request may be made at a time

This ensures that we display all the search results and that they are displayed in the right order. It also ensures we're not putting unnecessary, bursty load on the backend by firing off like 15 search requests at the same time 😃 

It seems the easiest way to test this if you don't have a fully set-up open discussions index locally is to point your ocw-hugo-themes repo at the RC search.

You can do that by setting:

```
SEARCH_API_URL=$DISCUSSIONS_RC/api/v0/search/
```

Then you'll face a CORS issue. You can either use a browser extension to turn off CORS temporarily, or you can pass a command-line flag to Chrome to open it without CORS. Probably not a good idea to then do anything else in this window at all!

The command to do that looks like:

```
/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --disable-web-security --user-data-dir=~/chromeTemp/
```

on macOS, and on Linux you can probably just do something like

```
google-chrome --disable-web-security --user-data-dir=~/chromeTemp/
```

I think this would probably work with Chromium too.

Screen Shot 2021-06-17 at 9 57 31 AM
Screen Shot 2021-06-17 at 9 58 18 AM

### DIFF
--- a/www/assets/js/components/SearchPage.js
+++ b/www/assets/js/components/SearchPage.js
@@ -34,7 +34,7 @@ export default function SearchPage() {
   const [suggestions, setSuggestions] = useState([])
   const [total, setTotal] = useState(0)
   const [completedInitialLoad, setCompletedInitialLoad] = useState(false)
-  const [busy, setBusy] = useState(false)
+  const [requestInFlight, setRequestInFlight] = useState(false)
 
   const runSearch = useCallback(
     async (text, activeFacets, from, sort) => {
@@ -43,7 +43,7 @@ export default function SearchPage() {
         activeFacets["type"] = [LR_TYPE_COURSE]
       }
 
-      setBusy(true)
+      setRequestInFlight(true)
       const newResults = await search({
         text,
         from,
@@ -51,7 +51,7 @@ export default function SearchPage() {
         size: SEARCH_PAGE_SIZE,
         sort: sort
       })
-      setBusy(false)
+      setRequestInFlight(false)
 
       const { suggest } = newResults
       if (!emptyOrNil(suggest) && !emptyOrNil(text)) {
@@ -76,7 +76,7 @@ export default function SearchPage() {
 
       setSearchFacets(new Map(Object.entries(newResults.aggregations ?? {})))
 
-      setSearchResults(
+      setSearchResults(results =>
         from === 0 ?
           newResults.hits.hits :
           [...results, ...newResults.hits.hits]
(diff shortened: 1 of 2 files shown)


www/assets/js/components/SearchPage.js
```diff
@@ -76,21 +76,23 @@ export default function SearchPage() {
 
       setSearchFacets(new Map(Object.entries(newResults.aggregations ?? {})))
 
-      setSearchResults(
+      setSearchResults(results =>
```
noisecapella, line 79: That's interesting, I wish they would highlight the need for this in the docs for handling previous state

noisecapella (changes requested):
If I make a change to a facet or to the sort dropdown it shows a loading indicator indefinitely

aliceriot:
aha ok well you can tell I tested thoroughly :P

I'll figure out what's up and ping you

aliceriot:
ok I think that should be fixed - I forgot that we reset `completedInitialLoad` to `false` after clearing the search because of a facet change.

www/assets/js/components/SearchPage.js
aliceriot: yeah...they discuss it in the docs but don't really clarify how helpful it is for avoiding race conditions for these kinds of things. It also means if you're passing the state update function in to a `useCallback` function you can avoid also passing the current value as a dependency, which I think would in a lot of cases drastically cut down on the number of unnecessary re-allocations of that function (which will increase memory usage and pressure on the GC and whatnot).

noisecapella (approved):
Looks good now :+1:

Issue #125, opened 2021-06-28
use new resource collection in www theme


gumaerc:
In https://github.com/mitodl/ocw-hugo-projects/pull/8, we set up the `ocw-www` studio config to have a collection of resources with uploaded files and metadata.  Now that we are moving to source `ocw-www` content from `ocw-studio`, we will need to update the way we source image URLs for promos and testimonials to reflect this change.  Currently, in the static version of `ocw-www` at https://github.com/mitodl/ocw-www we simply store the image URL under the `image` front matter property and the alt text in `image_alt`.  This needs to be updated to reflect the structure output by `ocw-studio` here: https://github.mit.edu/ocw-content-rc/ocw-www_eb8b70174c714752a91dbe730af76769

PR #126, opened 2021-06-28
adapt to use ocw-studio generated ocw-www content

gumaerc:
#### Pre-Flight checklist

- [x] Testing

Closes https://github.com/mitodl/ocw-hugo-themes/issues/125

#### What's this PR do?
In mitodl/ocw-hugo-projects#8, we set up the `ocw-www` studio config to have a collection of resources with uploaded files and metadata. Now that we are moving to source `ocw-www` content from `ocw-studio`, we will need to update the way we source image URLs for promos and testimonials to reflect this change. Currently, in the static version of `ocw-www` at https://github.com/mitodl/ocw-www we simply store the image URL under the image front matter property and the alt text in image_alt.

This PR adds the `resource_url.html` partial to the `www` theme.  It takes two arguments, the first being `filetype` and second being `uid`.  The partial retrieves a resource by first filtering the resources by `filetype` and then matching on `uid`.  It then returns the URL for the image minus the hostname and schema, leaving a relative URL.  `ocw-studio` currently writes in fully qualified S3 URLs into the `file` field, and these need to be stripped down to just be a relative URL in the final output as it will be served from a CDN.  This partial is then used in the various locations that promo and testimonial images are displayed throughout our templates.

 - If you have never run this project before, run `yarn install`
 - Run `npm run start:site:webpack`
 - Clone https://github.com/mitodl/ocw-hugo-projects locally
 - Clone https://github.mit.edu/ocw-content-rc/ocw-www_eb8b70174c714752a91dbe730af76769 locally and open a terminal there
 - Run `hugo server --config /path/to/ocw-hugo-projects/ocw-www/config.yaml --themesDir /path/to/ocw-hugo-themes`, replacing `/path/to` with your path to these respective repos
 - Browse to http://localhost:1313 and check the URLs on the images for promos and testimonials at:
   - http://localhost:1313
   - http://localhost:1313/testimonials/
   - http://localhost:1313/testimonials/clinton-blackburn/
 - Note that the images won't load locally as they are being stripped of their hostname down to relative URLs for use behind the CDN.  What you should expect to see for each URL is a pattern of `/ocw-www/filename`.  Check the markdown source for any given testimonial or promo and make sure that the URL written into the outputted HTML is essentially the `file` property from the markdown minus the schema and domain (https://ol-ocw-studio-app-qa.s3.amazonaws.com/)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -43,6 +43,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile --ignore-engines --prefer-offline
 
+      - name: Clear ocw-www directory
+        run: (cd ..; rm -rf ocw-www) 
+
       - name: Clone ocw-www for content
         run: (cd ..; git clone git@github.mit.edu:ocw-content-rc/ocw-www_eb8b70174c714752a91dbe730af76769.git ocw-www)
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -53,7 +53,7 @@ jobs:
         run: npm run build:webpack
 
       - name: Run hugo build
-        run: (cd ../ocw-www; hugo --config ../ocw-hugo-projects/ocw-www/config.yaml --themesDir ../ocw-hugo-themes)
+        run: (cd ../ocw-www; hugo --config ../ocw-hugo-projects/ocw-www/config.yaml --themesDir ../ocw-hugo-themes; ls -lah public)
         env:
           OCW_STUDIO_BASE_URL: ${{ secrets.OCW_STUDIO_BASE_URL }}
           SEARCH_API_URL: ${{ secrets.SEARCH_API_URL }}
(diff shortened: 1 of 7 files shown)


mbertrand (commented):
LGTM.  Seems like it should work fine on RC, but will these changes break production ocw-next?  Currently there is no production git organization, git site repo syncing is not enabled on ocw-studio production, etc.

mbertrand:
PS The lighthouse test failed:
```
Error: Error building site: failed to render pages: render of "page" failed: "/home/runner/work/ocw-hugo-themes/ocw-hugo-themes/www/layouts/testimonials/single.html:12:89": execute of template failed: template: testimonials/single.html:12:89: executing "main" at <$testimonial.Params.image.content>: can't evaluate field content in type string
```

but it deployed fine for me locally.

gumaerc:
@mbertrand should be all set

mbertrand (approved):
👍

Issue #161, opened 2021-08-25
home course cards display course level improperly


gumaerc:
### Steps to Reproduce

Spin up or build `ocw-www`

### Expected Behavior

The home course cards display a level for each course

### Actual Behavior

The course cards show a string rendered map of the level and a search URL for the level

pdpinch:
do we need some tests in the studio API to prevent this from happening again? And/or an API version number?

gumaerc:
@pdpinch Really this has little to do with the API itself.  The `metadata` property on a given course is essentially a copy of what `ocw-to-hugo` is outputting into the course data template (`data/course.json`) in the case of imported courses.  When a change is made in `ocw-to-hugo` currently, the practice is that a new version is published to npm and then you make a PR in `ocw-hugo-themes` that makes the necessary template changes as well as upping the `ocw-to-hugo` version in `package.json`.  This ensures that when a new release is created, RC and subsequently Prod are using the correct version of `ocw-to-hugo` to produce buildable content.  Since the `ocw-studio` API sources its data from the Django DB which currently gets course information by importing `ocw-to-hugo` content output by the build, the API doesn't import the updated information in each environment until an OCW Next site build is done and the new `ocw-to-hugo` data is uploaded to S3 and subequently imported into studio with `import_ocw_course_sites`.  In this case, `home_course_cards.html` is sourcing its data from the API.  You could make the necessary change in the `ocw-hugo-themes` PR I mentioned earlier where you update the `ocw-to-hugo` version, but then when you released that change the home course cards would be broken because the updated data hasn't yet been imported into `ocw-studio`.  So, that is a bit of a "chicken and egg" problem.  In addition to that, we have this issue with unpublished courses lingering in `ocw-studio` after they are unpublished and removed from the `ocw-to-hugo` S3 bucket which leaves the unpublished courses with the outdated metadata structure.  That issue will be fixed by https://github.com/mitodl/ocw-studio/pull/526 which I just merged.  As far as the API / template discrepancy, the best solution I can think of at the moment is a getting a second PR ready on the template that accesses the API that is ready to merge as soon as the new data is deployed and `import_ocw_course_sites` is run.

PR #165, opened 2021-08-31
use level text and search url

gumaerc:
#### Pre-Flight checklist

- [ ] Testing

Closes https://github.com/mitodl/ocw-hugo-themes/issues/161

#### What's this PR do?
In https://github.com/mitodl/ocw-to-hugo/pull/244, the `level` property was changed to an object containing the text of a level as well as a search URL.  The `home_course_cards.html` template was never adjusted to match this, and instead of displaying the level, it's displaying a string representation of the level object.  This PR adjusts the template to generate a link to the search URL using the data contained in the object.

 - Read the readme if you've never spun up `ocw-wvw` before using `ocw-hugo-themes`, you'll need to clone https://github.com/mitodl/ocw-www and point to it using `EXTERNAL_SITE_PATH`
 - Spin up `ocw-www` using `npm start`
 - Verify that the `level` links on the course home cards are clickable links that take you to `/search` with the appropriate query string (note that images will not work locally)

### DIFF
--- a/www/layouts/partials/home_course_cards.html
+++ b/www/layouts/partials/home_course_cards.html
@@ -34,7 +34,7 @@
                     </a>
                     <div class="course-card-content pt-1 px-3 pb-3">
                       <div class="course-level">
-                        {{ $courseData.primary_course_number }} | {{ $courseData.level }}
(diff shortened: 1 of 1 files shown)


abeglova (approved):
👍

Issue #170, opened 2021-09-07
no new courses found


gumaerc:
### Steps to Reproduce

Build `ocw-www`

### Expected Behavior

A list of new courses from the `ocw-studio` API is displayed near the top of the home page

### Actual Behavior

It finds no courses and instead displays "No new courses found"

PR #171, opened 2021-09-07
ocw-course not course

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/170

#### What's this PR do?
Recently in `ocw-studio`, we implemented the use of imported `WebsiteStarter` config files from the [`ocw-hugo-projects`](https://github.com/mitodl/ocw-hugo-projects) repo.  The name of the course site starter there is `ocw-course`, so courses imported into `ocw-studio` (and subsequently delivered by the API) are of type `ocw-course`.  Previously the `ocw-studio` import process used a starter checked into the `ocw-studio` repo simply called `course`.  This PR updates the API request on `home_course_cards.html` to specify the `ocw-course` type rather than `course` so that courses are returned in the results.

 - Read the readme before if you have never spun up `ocw-www` locally
 - Clone https://github.com/mitodl/ocw-www
 - In your `.env`, make sure `EXTERNAL_SITE_PATH` is set to your path to `ocw-www` and make sure `OCW_STUDIO_BASE_URL` is set to http://ocw-studio-rc.odl.mit.edu and `OCW_IMPORT_STARTER_SLUG` is set to `ocw-course`
 - Run `npm start`
 - Visit http://localhost:3000/ and ensure that courses are displayed in the new courses section, noting that images will not work because of the current `coursemedia` prefix being assigned


www/layouts/partials/home_course_cards.html
```diff
@@ -1,7 +1,8 @@
 <div class="new-courses course-cards standard-width container mx-auto mt-3">
 {{ $studioBaseUrl := getenv "OCW_STUDIO_BASE_URL" }}
-{{ if $studioBaseUrl }}
-{{  with getJSON (print (strings.TrimSuffix "/" $studioBaseUrl) "/api/websites/?type=course")  }}
+{{ $courseStarterSlug := getenv "OCW_IMPORT_STARTER_SLUG" }}
```
noisecapella, line 3: Are the environment variables defined for ocw-studio the same environment variables defined for ocw-hugo-themes?

gumaerc: Good point, there will need to be a `salt-ops` PR to set these since the current OCW Next site build server is unique.  Either way, I think we should keep the name consistent.

noisecapella: Can we just pick a constant and use that everywhere? I'm not sure why this was made an environment variable

gumaerc: When we started development of `ocw-studio`, a "course" starter was added directly to the repo in the `localdev` folder, but it ended up being used for more than just "local development."  It's used as a fixture essentially in every deployed instance of `ocw-studio` so even in RC and Prod we end up with a `course` starter already existing.  The starter in `ocw-hugo-projects` was named `ocw-course` as to not conflict.  They can't be named the same thing, so it was made configurable so you can use either the local version checked into `ocw-studio` or the imported version from `ocw-hugo-projects`.  I made a PR to set the value on the OCW Next build server here: https://github.com/mitodl/salt-ops/pull/1456

gumaerc:
@noisecapella The `OCW_STUDIO_STARTER_SLUG` variable is set to `course` on CI because of what I described above with the repo version of the course starter being called `course` in `ocw-studio`.  A large amount of unit tests in `ocw-studio` were built around importing the starter checked into the repository and using that.  The use case for it being an env variable here too would be if you were debugging starter changes locally and needed to test the API against a site that was created with your local `course` starter.  In general I agree that there should really only be one course starter, but the decentralization of the course starter files in `ocw-hugo-projects` makes that difficult because `ocw-studio` needs to maintain its own version for unit testing.  When I originally implemented the `OCW_IMPORT_STARTER_SLUG` env variable in `ocw-studio`, it was because adding a step to pull in the starter configs from `ocw-hugo-projects` instead of using the local ones in the repo was not desirable because of adding that dependency to `ocw-studio`.

www/layouts/partials/home_course_cards.html
```diff
@@ -1,7 +1,8 @@
...
+{{ if and $studioBaseUrl $courseStarterSlug }}
+{{  with getJSON (print (strings.TrimSuffix "/" $studioBaseUrl) "/api/websites/?type=" $courseStarterSlug)  }}
```
noisecapella, line 5: Can we use [querify](https://gohugo.io/functions/querify/) here to ensure the course slug is escaped properly?

noisecapella (changes requested):

gumaerc:
@noisecapella all set

noisecapella (approved):

PR #180, opened 2021-09-09
allow comments in .env, put .env sourcing in one place

### LINKED ISSUE #178: allow commenting-out lines in `.env`
### Steps to Reproduce

Add a `#` to a line in `.env` to comment out a variable.

### Expected Behavior

With a lot of tools that handle `.env` files like we do here a line starting with `#` will be treated as a comment, so I expected I'd be able to do that here too.

### Actual Behavior

`npm start` results in an error:

```
build_scripts/prep_external_site.sh: line 6: export: `#': not a valid identifier
```

This is a result of doing

```bash
export $(cat .env | xargs)
```

I think we can safely change this to the following:

```bash
export $(cat .env | grep -v '#' | xargs)
```

the `-v` flag inverts grep's normal behavior so that it only includes lines which _don't_ match the provided string. I tested out locally replacing each instance of `$(cat .env | xargs)` this way and it seems to work fine. There's an argument we should define a file with a single bash function in it for this functionality as well.

alicewriteswrongs:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes #178 

#### What's this PR do?

two things:

- allow use of `#` to comment out lines in a `.env` file
- write a single function for sourcing `.env` which can be called in all of our bash scripts in this repo

I think try a few things:

- ensure all functionality works as normal with your standard `.env` file
- add a commented line to it and ensure it still works as well
- delete `.env` and ensure that build scripts still fail on `env file not found` NOT on trying to access a variable. i.e. if you want add an `echo` statement just after the line where we call `source_env` in one of the build scripts and ensure that that line doesn't execute.

those are the main things I think.

### DIFF
--- a/build_scripts/build.sh
+++ b/build_scripts/build.sh
@@ -2,9 +2,8 @@
 
 set -euo pipefail
 
-if test -f ".env"; then
-  export $(cat .env | xargs)
-fi
+source build_scripts/common.sh
+load_env
 
 if [[ -z "${EXTERNAL_SITE_PATH}" ]]; then
   echo "EXTERNAL_SITE_PATH not set"
--- a/build_scripts/build_all_courses.sh
+++ b/build_scripts/build_all_courses.sh
@@ -2,9 +2,8 @@
 
 set -euo pipefail
 
-if test -f ".env"; then
-  export $(cat .env | xargs)
-fi
+source build_scripts/common.sh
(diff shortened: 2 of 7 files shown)


gumaerc (approved):
:+1:

PR #193, opened 2021-09-21
Update template to use newer format for topics

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Part of https://github.com/mitodl/ocw-studio/issues/466
Related to https://github.com/mitodl/ocw-to-hugo/pull/370

#### What's this PR do?
Topics are now stored using the hierarchical select layout which is just a list of lists. This adds code to the template to render topics from that format.

In general you shouldn't notice any significant changes. Try converting these courses with https://github.com/mitodl/ocw-to-hugo/pull/370 and view these courses:
 - `21g-065-japanese-literature-and-cinema-fall-2013` has multiple top level topics
 - `18-06-linear-algebra-spring-2010` has only top level topics
 - `17-263-american-elections-fall-2020` has multiple subtopics sharing the same topic

Compare each to what is on production now. The course home page should look identical. Click one of the section pages and view the course info. The same topics should be expanded or hidden, and the same subtopics and specialities should be listed. If you hover over the links it should show a search link which is almost identical to what is on production, except that spaces are replaced with `+`, which are equivalent in query strings.


abeglova (approved):
LGTM

Issue #211, opened 2021-10-01
course images don't load with ocw-to-hugo 1.31.1 content


gumaerc:
### Steps to Reproduce

Convert a course using `ocw-to-hugo` version 1.31.1 or greater and spin that course up using the theme

### Expected Behavior

The course site renders and the course image is visible

### Actual Behavior

The site renders but the course image is broken

PR #212, opened 2021-10-01
course images from resources

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/211

#### What's this PR do?
In https://github.com/mitodl/ocw-to-hugo/pull/378, we altered the rendering of `course.json` to produce an `ocw-studio`-like reference to the course image resource content by UUID.  This PR updates the version of `ocw-to-hugo` in `package.json` and adjusts the theme to source images from the content in this manner:

 - `resource_url.html` was moved to `base-theme`
 - `resource_url.html` was set up to look for a URL in `.Params.file` first, but fall back to `.Params.file_location` because this is where `ocw-to-hugo` currently puts this property in converted legacy resources
 - An optional `RESOURCE_BASE_URL` env variable was added that will prefix resource URLs rendered in this partial with the value supplied. This way, when you are developing a course site locally, you can manually set the S3 base URL written in front of resources so images load correctly. The value can also just be left blank or unset and relative URLs will be written in instead.

 - Read the readme if you have never spun up a course site before
 - Ensure the following values are set in your `.env`:
   -`RESOURCE_BASE_URL=https://open-learning-course-data-rc.s3.amazonaws.com/`
  - `OCW_TO_HUGO_PATH` should be blank or unset
  - `OCW_TEST_COURSE` to any course id
 - Run `yarn install`
 - Run `npm run start:course` and visit http://localhost:3000
 - Ensure that the course image renders

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ There are some environment variables you can use to configure optional funcional
 | `SEARCH_API_URL` | `http://localhost:8063/api/v0/search/` | A URL to an `open-discussions` search API to fetch results from|
 | `OCW_STUDIO_BASE_URL` | `http://localhost:8043/` | A URL of an instance of [`ocw-studio`](https://github.com/mitodl/ocw-studio) to fetch home page content from |
 | `STATIC_API_BASE_URL` | `http://ocwnext.odl.mit.edu/` | A URL of a deployed Hugo site with a static JSON API to query against |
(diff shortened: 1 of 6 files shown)


alicewriteswrongs (approved):
everything seems to be working correctly!

Issue #210, opened 2021-10-01
Show caption text under video resources


abeglova:
Show caption text underneath videos. The caption corresponding to the current point in the video should be highlighted. Also, the user should be able to select a caption and have the video update to that point.

Mockup:
https://projects.invisionapp.com/share/X8107I6ESWT3#/screens/444868933_OCW_Course_Video_Page_02-12-2021_Page_Info_Exanded

PR #228, opened 2021-10-14
show transcripts under video

abeglova:
#### Pre-Flight checklist

Screen Shot 2021-10-14 at 7 46 23 PM
Screen Shot 2021-10-14 at 7 44 58 PM

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/ocw-hugo-themes/issues/210

#### What's this PR do?
This pr adds a div with caption text under videos. The caption corresponding to the current point in the video should be highlighted. Also, the user should be able to select a caption and have the video update to that point.

Run `npm run start:course` with a course that has (non-inline) lecture videos for example set `OCW_TEST_COURSE=15-401-finance-theory-i-fall-2008` in your .env file.

Go to a lecture video. Verify that you see a box with captions and the captions look good

### DIFF
--- a/course/layouts/partials/video.html
+++ b/course/layouts/partials/video.html
@@ -1,10 +1,27 @@
 {{ $youtubeKey := .Params.video_metadata.youtube_id }}
 {{ $captionsLocation := .Params.video_files.video_captions_file }}
-{{ $transcriptLocation := .Params.video_files.video_transcript_file }}
+{{ $transcriptPdfLocation := .Params.video_files.video_transcript_file }}
 
 <div class="video-page">
   <div class="description">
     {{ .Params.about_this_resource_text | safeHTML }}
   </div>
-  {{ partial "youtube_player.html" (dict "youtubeKey" $youtubeKey "captionsLocation" $captionsLocation "transcriptLocation" $transcriptLocation) }}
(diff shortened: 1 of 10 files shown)


course/layouts/partials/video.html
```diff
@@ -7,4 +7,21 @@
     {{ .Params.about_this_resource_text | safeHTML }}
   </div>
   {{ partial "youtube_player.html" (dict "youtubeKey" $youtubeKey "captionsLocation" $captionsLocation "transcriptLocation" $transcriptLocation) }}
+  {{if $captionsLocation}}
```
gumaerc, line 10: Shouldn't this be `{{ if $transcriptLocation }}`?

gumaerc (changes requested):
This looks and functions great, just one question about an if statement:

course/layouts/partials/video.html
abeglova: No,` transcriptLocation` is the location of the downloadable pdf transcript.

`captionsLocation` is the location of the  .vtt captions file which is used by videojs and this plugin to display captions

abeglova:
I made the update. Back to you @gumaerc

course/layouts/partials/video.html
gumaerc: A comment would have been fine, but that works too

Issue #87, opened 2021-06-04
use themeDir instead of Hugo Modules


gumaerc:
The Hugo Module system offers a lot of functionality for merging components into your site build, but it has some drawbacks that have been complicating our build process locally and in production.  Rather than inheriting our themes as Hugo modules, they should simply be specified using the `themesDir` and `theme` directives.

PR #231, opened 2021-10-18
overhaul local dev / package scripts / documentation

gumaerc:
#### Pre-Flight checklist

- [ ] Testing

Closes https://github.com/mitodl/ocw-hugo-themes/issues/87

#### What's this PR do?
Over the course of working on https://github.com/mitodl/ocw-studio/issues/568, we've altered the way the theme sources and renders metadata and content to match the way `ocw-studio` publishes sites.  This PR makes some further adjustments to the theme to make it easier to build courses locally that have been published by `ocw-studio`.  The use of Hugo modules has also been removed and replaced with the new strategy of running `hugo` from the content directory, pointing to the relevant config from `ocw-hugo-projects` with `--config` and pointing to `ocw-hugo-themes` with `--themesDir`.  When running sites locally, `concurrently` (https://github.com/open-cli-tools/concurrently) is now being used to manage killing off both the Hugo and Webpack processes when you hit ctrl + c to stop the local dev server.

 - Read through the new readme thoroughly and make sure it makes sense: https://github.com/mitodl/ocw-hugo-themes/tree/cg/overhaul-local-dev
 - Following the instructions in the new readme:
   - Spin up `ocw-www` locally
   - Spin up a course site locally
 - Verify that `concurrently` is working properly, which should automatically close both the Webpack dev server and the Hugo server when you press ctrl + c in your terminal.  After closing the dev servers, you should be able to run `ps aux | grep hugo` and not see any lingering processes.

References to `ocw-to-hugo` have been removed as we are nearing production readiness with `ocw-studio`.  Since `ocw-to-hugo` converted courses are now being imported into `ocw-studio` and published in the same way the rest of the `ocw-studio` courses are, spinning up courses directly from `ocw-to-hugo` is mostly irrelevant.  Spinning up courses directly generated by `ocw-to-hugo` can still be done, but will need to be done manually.

### DIFF
--- a/package_scripts/start_course.sh
+++ b/package_scripts/start_course.sh
@@ -6,10 +6,19 @@ source package_scripts/common.sh
 load_env --require-dot-env
 
 THEMES_PATH=$(pwd)
+STATIC_API_BASE_URL=${STATIC_API_BASE_URL:-}
+RESOURCE_BASE_URL=${RESOURCE_BASE_URL:-}
 COURSE_HUGO_CONFIG_PATH=${COURSE_HUGO_CONFIG_PATH:-}
 COURSE_CONTENT_PATH=${COURSE_CONTENT_PATH:-}
 OCW_TEST_COURSE=${OCW_TEST_COURSE:-}
 
+# If the base URL variables aren't set, assign defaults
+if [[ -z "${STATIC_API_BASE_URL}" ]]; then
+  export STATIC_API_BASE_URL="https://ocw-live-qa.global.ssl.fastly.net/"
+fi
+if [[ -z "${STATIC_API_BASE_URL}" ]]; then
+  export RESOURCE_BASE_URL="https://ol-ocw-studio-app-qa.s3.amazonaws.com/"
+fi
 if [[ -z "${COURSE_HUGO_CONFIG_PATH}" ]]; then
   # Path to Hugo config not set, fetch default from Github
   clone_or_pull_repo git@github.com:mitodl/ocw-hugo-projects.git private/ocw-hugo-projects
@@ -33,4 +42,9 @@ fi
 
 cd $COURSE_CONTENT_PATH
 # Run hugo server
-hugo server -p 3000 --bind 0.0.0.0 --config $COURSE_HUGO_CONFIG_PATH --themesDir $THEMES_PATH --renderToDisk
+hugo server \
+  -p 3000 \
+  --bind 0.0.0.0 \
+  --config $COURSE_HUGO_CONFIG_PATH \
+  --themesDir $THEMES_PATH \
+  --renderToDisk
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,191 +1,154 @@
...
+
+ - Clone the `ocw-www` content repo: https://github.mit.edu/ocw-content-rc/ocw-www
+ - Clone `ocw-hugo-projects` to obtain the relevant configuration file: https://github.com/mitocw/ocw-hugo-projects
+ - Set the following environment variables in your `.env` file, replacing `/path/to` with your path to the repos indicated:
+   - `WWW_HUGO_CONFIG_PATH=/path/to/ocw-hugo-projects/ocw-www/config.yaml`
+   - `WWW_CONTENT_PATH=/path/to/ocw-content-rc/ocw-www/`
+ - Optionally set these environment variables as well, depending on the functionality you need to work on:
+   - `SEARCH_API_URL=https://discussions-rc.odl.mit.edu/api/v0/search/` (for testing search functionality)
+   - `OCW_STUDIO_BASE_URL=http://ocw-studio-rc.odl.mit.edu/` (for testing `ocw-studio` API functionality, such as the "new courses" section rendered by `www/layouts/partials/home_course_cards.html`)
+   - `RESOURCE_BASE_URL=https://open-learning-course-data-rc.s3.amazonaws.com/` (if you need to test the loading of resources from S3 or some other CDN)
```
alicewriteswrongs, line 125: what would you think about automating these setup steps? I would _love_ to have a `dev.sh` script or similar which automatically handled the happy path for all these things so that testing out a styling change locally was really straightforward (we could then provide overrides via env variables or something for things which might need to change)

alicewriteswrongs (changes requested):
one initial question

README.md
gumaerc: You mean like some test content to spin up `www` or `course` sites without any external content?  I think that could be done.  We'd have to create and maintain a "baseline" set of content for each theme but I think that's probably a good idea anyway and could be used for some basic assertion of the output as well.  I'll work on that.

alicewriteswrongs: I don't know that test content is necessary - but just something where there is a zero-config way to get the site going by running a single script, perhaps just by automating cloning the repos mentioned here, setting the right environment variables, etc. One of the things I've felt has been "friction-ey" about spinning up the ocw-hugo-themes for development is that the dev setup takes a number of steps, so if you haven't been working on it for a bit it is a bit error prone and can end up taking up a chunk of time if you do something slightly wrong in a difficult-to-detect way. So if there are, as outlined here in the README, a set of relatively concrete, repeatable steps for spinning up the ocw-www site, it would be great I think to automate those if we can.

Kind of like the equivalent of how for running ocw-studio locally in development we really only support using `docker-compose`, and, aside from environment variables which are somewhat messy to configure, bringing up the whole application and it's environment is just `docker-compose build && docker-compose up`. It would be great to just be able to do something like `./ocw-www-dev.sh` and have something similar happen.

gumaerc:
@alicewriteswrongs I added some functionality that will by default clone the necessary repos you need for spinning up `www` or `course`.  Read the readme changes and try it out and let me know what you think.

PR #249, opened 2021-10-29
Adjust level, term to new ocw-to-hugo format

noisecapella:
#### Pre-Flight checklist

- [x] Testing

Related to https://github.com/mitodl/ocw-to-hugo/pull/398

#### What's this PR do?
`level` was changed to an array so this PR changes it to output one level link per item in the array. Additionally, `term` was changed to not include a year anymore, and `year` was added with the corresponding year. 

Convert a course with the latest ocw-to-hugo and view a course with this PR. The level in the course info on the course home page should look the same as before, with a link to the search page. The text next to "As Taught In" should be a semester and a year, looking exactly like it did before

### DIFF
--- a/course/layouts/partials/course_info.html
+++ b/course/layouts/partials/course_info.html
@@ -47,7 +47,7 @@
       <tr>
         <td class="pl-0 text-nowrap">As Taught In:</td>
         <td>
-          {{ $courseData.term }}
+          {{ $courseData.term }} {{ $courseData.year }}
         </td>
       </tr>
       <tr>
--- a/course/layouts/partials/course_info.html
+++ b/course/layouts/partials/course_info.html
@@ -47,17 +47,24 @@
       <tr>
         <td class="pl-0 text-nowrap">As Taught In:</td>
         <td>
-          {{ $courseData.term }} {{ $courseData.year }}
+          {{ $courseData.term }} {{ if isset $courseData "year" }}{{ $courseData.year }}{{ end }}
         </td>
       </tr>
       <tr>
         <td class="pl-0 pb-3">Level:</td>
         <td>
-          {{ range $courseData.level }}
-            {{ $levelSearchUrl := partial "get_search_url.html" (dict "key" "level" "value" .) }}
+          {{ if eq (printf "%T" $courseData.level) "string" }}
+            {{ $levelSearchUrl := partial "get_search_url.html" (dict "key" "level" "value" $courseData.level) }}
             <a href="{{ $levelSearchUrl }}" class="course-info-level">
-              {{ . }}
-            </a><br />
+              {{ $courseData.level }}
+            </a>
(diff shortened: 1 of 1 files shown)


gumaerc (dismissed):
:+1: LGTM

gumaerc:
Actually, a caveat here.  After approving this I tested it with an `ocw-studio` published course and the build failed:

```
[start:course:hugo] Error: Error building site: failed to render pages: render of "page" failed: execute of template failed: template: pages/single.html:14:7: executing "pages/single.html" at <partialCached "mobile_course_info.html" .>: error calling partialCached: execute of template failed: template: partials/mobile_course_info.html:5:5: executing "partials/mobile_course_info.html" at <partial "course_info.html" (dict "context" . "inPanel" true)>: error calling partial: "/home/gumaerc/Code/ocw-hugo-themes/course/layouts/partials/course_info.html:56:30": execute of template failed: template: partials/course_info.html:56:30: executing "partials/course_info.html" at <$courseData.level>: range can't iterate over
```
It seems we need to make changes to the `ocw-course` starter in `ocw-hugo-projects` to match what was done here.  If those changes aren't made before this PR is merged and pushed to RC, builds there will start to fail.

course/layouts/partials/course_info.html
```diff
@@ -47,16 +47,18 @@ <h4 class="course-info-title font-weight-bold">Course Info</h4>
...
         </td>
       </tr>
       <tr>
         <td class="pl-0 pb-3">Level:</td>
         <td>
-          {{ $levelSearchUrl := partial "get_search_url.html" (dict "key" "level" "value" $courseData.level) }}
-          <a href="{{ $levelSearchUrl }}" class="course-info-level">
-            {{ $courseData.level }}
-          </a>
+          {{ range $courseData.level }}
```
gumaerc, line 56: We should check if `$courseData.level` is an array here and only iterate it if that's the case.

course/layouts/partials/course_info.html
```diff
@@ -47,16 +47,18 @@ <h4 class="course-info-title font-weight-bold">Course Info</h4>
       <tr>
         <td class="pl-0 text-nowrap">As Taught In:</td>
         <td>
-          {{ $courseData.term }}
+          {{ $courseData.term }} {{ $courseData.year }}
```
gumaerc, line 50: We should check `if isset $courseData "year"` before trying to access it here

gumaerc (changes requested):
In order to be compatible with courses published from `ocw-studio` currently, we should make this compatible with both new and old versions of the data.

noisecapella:
@gumaerc all set

gumaerc (approved):
:+1: LGTM

It seems we don't actually need to check access on year.  I had it unset in my test course and didn't get any errors.

Issue #215, opened 2021-10-06
Quizzes aren't being rendered correctly


abeglova:
An example of a page that isn't being rendered properly:
https://ocwnext.odl.mit.edu/courses/14-01sc-principles-of-microeconomics-fall-2011/pages/unit-1-supply-and-demand/introduction-to-microeconomics/

PR #290, opened 2021-11-30
multiple choice

abeglova:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested
Screen Shot 2021-11-30 at 3 04 32 PM
Screen Shot 2021-11-30 at 3 04 58 PM
Screen Shot 2021-11-30 at 3 05 16 PM

closes https://github.com/mitodl/ocw-hugo-themes/issues/215

#### What's this PR do?
This PR fixes multiple choice questions for legacy courses

Select the branch from https://github.com/mitodl/ocw-to-hugo/pull/418 in ocw-to-hugo

Set example_courses.json in ocw-to-hugo to
```
{
    "courses": [
		"6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016",
		"14-01sc-principles-of-microeconomics-fall-2011",
    ]
}
```
Run
```
node . -i private/input -o private/output -c course_json_examples/example_courses.json --download
```
in ocw-to-hugo.

In ocw-hugo-themes,
Set your .env to
```
OCW_TEST_COURSE=6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016
COURSE_CONTENT_PATH=/ocw-to-hugo/private/output
DOWNLOAD=0
```
Run
```
npm run start:course
```
Go to a page with multiple choice questions and verify that both the "Check" and the "Show Solution" buttons work

Set
OCW_TEST_COURSE=14-01sc-principles-of-microeconomics-fall-2011

Again, go to a page with multiple choice questions and verify that both the "Check" and the "Show Solution" buttons work

Code block rendering and videos are currently broken, that's not related to these changes

### DIFF
--- a/course/assets/course.js
+++ b/course/assets/course.js
@@ -11,6 +11,11 @@ import { initDesktopCourseInfoToggle } from "./js/course_info_toggle"
 import { initDivToggle } from "./js/div_toggle"
 import { initCourseInfoExpander } from "./js/course_expander"
 import { initVideoTranscriptTrack } from "./js/video_transcript_track"
+import {
+  clearSolution,
+  checkAnswer,
+  showSolution
+} from "./js/quiz_multiple_choice"
 
 $(document).ready(() => {
   initPdfViewers()
@@ -18,4 +23,7 @@ $(document).ready(() => {
   initCourseInfoExpander(document)
   initVideoTranscriptTrack()
   initDivToggle()
+  clearSolution()
+  checkAnswer()
+  showSolution()
 })
--- a/course/assets/css/course.scss
+++ b/course/assets/css/course.scss
@@ -37,11 +37,36 @@
   }
 }
 
-div[class^="toggle"]:not(.toggle-visible) {
+div[class^="toggle"]:not(.toggle-visible),
+span[class^="toggle"]:not(.toggle-visible) {
   display: none;
 }
 
-div[class^="reveal"] {
+div[class^="reveal"],
+.multiple-choice-check-button,
+.multiple-choice-show-button {
   color: #115f83;
   cursor: pointer;
+}
+
+.correctness-icon-correct {
+  color: green;
+}
+
+.correctness-icon-wrong {
+  color: red;
+}
+
+.correctness-icon {
+  position: relative;
+  top: 6px;
+}
+
+.multiple-choice-radio {
+  margin-top: 11px;
+  margin-bottom: 11px;
+}
+
+.multiple_choice_buttons {
+  margin: 10px;
 }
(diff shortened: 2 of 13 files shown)


abeglova:
I don't think the test error is related to my pr

course/assets/js/multiple_choice.js
```diff
@@ -0,0 +1,35 @@
...
+
+      for (const solution of solutions) {
+        solution.classList.remove("toggle-visible")
+      }
+    })
+  }
+}
+
+export const checkAnswer = () => {
+  const buttons = document.getElementsByClassName(
```
gumaerc, line 17: Similarly to above, I think we can avoid declaring the extra constant and loop here and instead just use:

```
  Array.from(
    document.getElementsByClassName("multiple-choice-check-button")
  ).forEach(button => {
    button.addEventListener("click", () => {
      Array.from(
        button
          .closest(".multiple-choice-question")
          .getElementsByClassName("multiple-choice-radio")
      ).forEach(input => {
        if (input.checked) {
          const solution = input
            .closest("div")
            .getElementsByClassName("correctness-icon")[0]
          solution.classList.add("toggle-visible")
        }
      })
    })
  })
```

course/layouts/shortcodes/choice.html
```diff
@@ -0,0 +1,5 @@
+{{ $correct := .Get "isCorrect" }}
```
gumaerc, line 1: I have some concerns about how generic the name of this and the "choices" shortcode are.  If we were to implement a more generic "multiple choice" type shortcode then we might want to use these names.  Since these refer specifically to the quizzes, could we name them something like `quiz_choice` and `quiz_choices`?  Similarly, I think `multiple-choice-question` could be named `quiz_question`.  I think we should be using underscore separators in the filenames.  I realize that video gallery is `video-gallery.html` but it should be `video_gallery.html` and I need to make an issue to change that.  It can be kind of tricky to rename these after the fact because so much content has to be republished, so it's best to really have good names before committing anything.  Also note that there can be partials and shortcodes with the same name, since shortcodes are only called from within markdown content and partials are only called inside of layouts.

gumaerc (changes requested):
This works well, but I have a minor code optimization to suggest as well as some naming concerns.  Also, I'm not sure if we have designs for how these quizzes are supposed to look / function but I did notice something about the behavior that is maybe unintentional.  If you select an answer and then click "check" to reveal whether or not the answer was correct, you get the x or the check mark, but then I noticed that if you click "show solution," the icon disappears.  If we don't have designs for this or aren't sure on the intention feel free to make an issue and put this off into another PR or something.

abeglova:
back to you @gumaerc, thanks for the review

gumaerc (approved):
:+1:

PR #326, opened 2022-01-06
delete a package-lock.json that snuck in here somehow

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

none

#### What's this PR do?

It looks like in #290 a `package-lock.json` file (the lockfile used by `npm`) was accidentally committed. Since we use yarn to install having this around doesn't really _hurt_, but we don't want to confuse ourselves!

everything should work as normal


gumaerc (approved):
:+1:

Issue #280, opened 2021-11-22
ocw-www: course collections


pdpinch:
As an OCW staff member, I'd like to be able to create and publish a collection of courses so I can, for example, publish a list of Open Learning Library courses. 

#### Designs and Mockups

Similar to the video gallery design at List of documents (PDFs) from https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/461095857/preview?scrollOffset=497.20001220703125 but with course thumbnails instead of video thumbnails. 

#### Related issues

- depends on adding course collections to project: https://github.com/mitodl/ocw-hugo-projects/issues/101

alicewriteswrongs:
The UI for this is very similar to the UI we show for the search results, so I'm going to look into what's required to re-use that here.

PR #335, opened 2022-01-11
switch to building w/ typescript

### LINKED ISSUE #324: convert search page and other JS code to Typescript
We probably want to do this at some point because

- our [course-search-utils](https://github.com/mitodl/course-search-utils) library is built with typescript, so it would be nice to take advantage of the type safety over here
- we have relatively little code here so a conversion isn't very arduous
- the code here is pure JS w/out any flow annotations, so we can do a gradual conversion.

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #324 

#### What's this PR do?

This PR sets up `ts-loader` and converts most of the JS files over to Typescript (especially all the files for the frontend search). I tried not to touch any logic at all in doing so.

I wanted to have this in place for working on #280.

Some files still have a number of type errors, but these are silenced with a `@ts-nocheck` at the top of the file. We should circle back and remove these over time.

Everything should work as before.


gumaerc (approved):
:+1: I pretty thoroughly went through everything, tested building all courses and all the features within the course theme as well as the search page, etc.  Everything worked as expected.  One question though, is there anything we want to add to the readme that developers might want to know in regards to the typescript configuration?

alicewriteswrongs:
That's a good point - I'll add some documentation to the README about just how to typecheck the project and whatnot

PR #358, opened 2022-01-20
Minor readme updates

ChristopherChudzicki:
I believe `npm run start:course` requires a course to have metadata created, which my test course did not initially. So added that to the README.

Also, updated the `.env.example` for `COURSE_CONTENT_PATH` to be consistent with the README.


alicewriteswrongs (approved):
lgtm!

Issue #86, opened 2021-06-03
No Learning Resource Types section should be displayed if there are none


noisecapella:
### Steps to Reproduce

You can see this on https://ocwnext-rc.odl.mit.edu/courses/11-405-just-money-banking-as-if-society-mattered-spring-2021/. If there are no features we should omit the section

![Screenshot_20210603-150233](https://user-images.githubusercontent.com/863262/120698240-c67cf980-c47c-11eb-9a55-e4321861ef35.png)

MAbdurrehman12:
@annagav Are you working on this or can I do it?

annagav:
@MAbdurrehman12 yes, you can take it. Sorry I forgot to unassign myself

MAbdurrehman12:
@pdpinch This ticket was opened by George, but I'm unable to find **course features** and unable to reproduce the issue. However, a similar issue is there for **learning resource types**, maybe course features has been renamed but I'm not sure.
Anyways, I have created a PR for learning resource types.

Please let me know if you have any thoughts on this.

pdpinch:
Thank you @MAbdurrehman12, I forgot the page section was changed to "Learning Resource Types" some time ago.

PR #362, opened 2022-01-21
fix: hide resource types when count is 0

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/86
**Note:** Ticket was created for **course features**, but I was unable to find course features, maybe it has been fixed or renamed, I'm not sure. Since the issue was also for **learning resource types**, so I fixed that instead.
For reference, please view [this comment](https://github.com/mitodl/ocw-hugo-themes/issues/86#issuecomment-1018337756)

#### What's this PR do?
Does not render learning resource types when it's count is 0

- Go to any course which has learning resource types:
    - Verify that the section is shown in home page and other pages.
- Go to any course which has **no** learning resource types:
    - Verify that the section is **not** shown in home page and other pages.    

Before: 
image

image

After: 
image

image

### DIFF
--- a/course/layouts/partials/learning_resource_types.html
+++ b/course/layouts/partials/learning_resource_types.html
@@ -1,14 +1,17 @@
 {{ $courseData := .context.Site.Data.course }}
 {{ $inPanel := .inPanel }}
-<div class="course-home-section w-100 {{ if $inPanel }}in-panel{{ else }}bg-light{{ end }}">
-  <div class="container px-0 mx-0">
-    <h4 class="course-info-title font-weight-bold">Learning Resource Types</h4>
-    <div class="row">
-      {{ range $courseData.learning_resource_types }}
-        <div class="{{ if $inPanel }}col-12{{ else }}col-6 col-xl-3{{ end }}">
-          {{ partial "learning_resource_type.html" . }}
-        </div>
-      {{ end }}
+
+{{if gt (len $courseData.learning_resource_types) 0}}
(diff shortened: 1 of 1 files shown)


course/layouts/partials/learning_resource_types.html
```diff
@@ -1,14 +1,17 @@
...
+        {{ range $courseData.learning_resource_types }}
+          <div class="{{ if $inPanel }}col-12{{ else }}col-6 col-xl-3{{ end }}">
+            {{ partial "learning_resource_type.html" . }}
+          </div>
+        {{ end }}
+      </div>
     </div>
   </div>
-</div>
+{{end}}
```
Wassaf-Shahzad, line 17: Looking good just one small change.
we just need to wrap this in another if condition or use the `and `operator to check if the  `$courseData.learning_resource_types` exists.
That way if  someone removes the learning_resource_type from course.json it will still work

Wassaf-Shahzad (changes requested):

course/layouts/partials/learning_resource_types.html
MAbdurrehman12: I'm not sure whether this is required as if $courseData.learning_resource_types does not exist then it will break in other places too. Even if we have to place a check for checking existence then I guess it should be placed in other files too and for other attributes too for better consistency. 
Maybe this can be done somewhere which is more generic or maybe we can do it here.
Perhaps @gumaerc can share his thoughts?

Wassaf-Shahzad (approved):

Issue #361, opened 2022-01-21
Document title doesn't wrap


pdpinch:
### Steps to Reproduce

1. Visit https://ocwnext-rc.odl.mit.edu/courses/6-811-principles-and-practice-of-assistive-technology-fall-2014/resources/mit6_811f14_octdeliverbles/

### Expected Behavior

- Document title should wrap, within the main content area
- If the title is too greater than 3 lines, we can use expand / collapse widget from course home page

### Actual Behavior

- document title keeps scrolling to the right 

### Screenshot or Screencast

![image](https://user-images.githubusercontent.com/430126/150468432-97864ee4-7472-4956-ab06-50c0da64b494.png)

### Out of scope

- I think this document template should also have a section header. In this case "Assignments"

PR #364, opened 2022-01-24
feat: expand/collapse in document title

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/361

#### What's this PR do?
**Note:** Initially this PR was opened to fix the `issue of title being overflown: title was not wrapped in it's div, it flowed to right outside it's section` but as per discussion in this [thread](https://github.com/mitodl/ocw-hugo-themes/pull/364#discussion_r791214738), this issue should be addressed upstream hence now this PR only adds expand/collapse in document title 

- Expand/collapse shows up if title covers more than 3 lines. 

- Verify that document title remains in it's section regardless of it's length/size. It is not flowed outside like shown in the issue.
- Verify that on increasing title length, expand/collapse shows up which correctly works with a smooth transition.
- Please repeat the above step for:
    1. all sections where title section is being rendered.
    2. multiple courses.
    3. on multiple browsers.

image

image

image

### DIFF
--- a/base-theme/assets/index.ts
+++ b/base-theme/assets/index.ts
@@ -27,11 +27,11 @@ window.Popper = Popper;
 
 $(document).ready(() => {
   // hacky coming-soon popover
-  document.querySelectorAll(".coming-soon").forEach((el) => {
+  document.querySelectorAll(".coming-soon").forEach(el => {
     tippy(el, {
       content: "Coming soon!",
       trigger: "click",
-      placement: "top",
+      placement: "top"
     });
   });
 
--- a/base-theme/assets/index.ts
+++ b/base-theme/assets/index.ts
@@ -1,29 +1,29 @@
-import "../../node_modules/tippy.js/dist/tippy.css";
+import "../../node_modules/tippy.js/dist/tippy.css"
 
-import "./css/main.scss";
-import "video.js/dist/video-js.css";
+import "./css/main.scss"
+import "video.js/dist/video-js.css"
 
-import "bootstrap";
-import Popper from "popper.js";
-import tippy from "tippy.js";
-import "shifty";
-import "hammerjs";
-import "imagesloaded";
-import "screenfull";
+import "bootstrap"
+import Popper from "popper.js"
+import tippy from "tippy.js"
+import "shifty"
+import "hammerjs"
+import "imagesloaded"
+import "screenfull"
 
-import { initSentry } from "./js/sentry";
+import { initSentry } from "./js/sentry"
 
 export interface OCWWindow extends Window {
-  $: JQueryStatic;
-  jQuery: JQueryStatic;
-  Popper: typeof Popper;
+  $: JQueryStatic
+  jQuery: JQueryStatic
+  Popper: typeof Popper
 }
 
-declare let window: OCWWindow;
+declare let window: OCWWindow
 
-window.jQuery = $;
-window.$ = $;
-window.Popper = Popper;
+window.jQuery = $
+window.$ = $
+window.Popper = Popper
 
 $(document).ready(() => {
   // hacky coming-soon popover
@@ -32,17 +32,17 @@ $(document).ready(() => {
       content: "Coming soon!",
       trigger: "click",
       placement: "top"
-    });
-  });
+    })
+  })
 
-  initSentry();
-});
+  initSentry()
+})
 
 // &nbsp; causes text to overflow and it doesn't wrap inside a div
 // not sure if this is a good approach but I couldn't ANY Hugo template function
 // with which we can get rid of &nbsp;
 // not wrapping this code in function for better efficiency and to avoid glitch for user
-const elements = document.getElementsByClassName("remove-nbsp");
+const elements = document.getElementsByClassName("remove-nbsp")
 for (let i = 0; i < elements.length; i++) {
-  elements[i].innerHTML = elements[i].innerHTML.replace(/&nbsp;/g, " ");
+  elements[i].innerHTML = elements[i].innerHTML.replace(/&nbsp;/g, " ")
 }
--- a/base-theme/assets/index.ts
+++ b/base-theme/assets/index.ts
@@ -37,12 +37,3 @@ $(document).ready(() => {
 
   initSentry()
 })
-
-// &nbsp; causes text to overflow and it doesn't wrap inside a div
-// not sure if this is a good approach but I couldn't ANY Hugo template function
-// with which we can get rid of &nbsp;
-// not wrapping this code in function for better efficiency and to avoid glitch for user
-const elements = document.getElementsByClassName("remove-nbsp")
-for (let i = 0; i < elements.length; i++) {
-  elements[i].innerHTML = elements[i].innerHTML.replace(/&nbsp;/g, " ")
-}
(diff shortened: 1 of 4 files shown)


base-theme/assets/index.ts
```diff
@@ -37,3 +37,12 @@ $(document).ready(() => {
 
   initSentry()
 })
+
+// &nbsp; causes text to overflow and it doesn't wrap inside a div
+// not sure if this is a good approach but I couldn't ANY Hugo template function
+// with which we can get rid of &nbsp;
+// not wrapping this code in function for better efficiency and to avoid glitch for user
+const elements = document.getElementsByClassName("remove-nbsp")
```
gumaerc, line 45: This seems more like it's a problem upstream with `ocw-to-hugo`.  I checked out the `ocw-to-hugo` output for the course indicated in the issue, and the `title` property on the indicated page is this:

`title: "Review\_of\_October\_Course\_Requirements\_and\_Deliverables"`

For some reason, `\_` is being inserted into the title where there should just be a space.  This is likely creating the HTML character entity `&nbsp;` and it should probably be addressed there and not after the fact here in the theme.

gumaerc (changes requested):
I tested the wrapping and expand / collapse functionality and it works great, but I think the issue you're seeing with `&nbsp;` in titles should be addressed upstream:

base-theme/assets/index.ts
MAbdurrehman12: @gumaerc Yeah, you're right, this should not be addressed here. 
Thank you for looking into it.

I was also wondering that from where `\_` is being appended, it seems to be in titles and some descriptions too, not sure if it occurs somewhere else too. 
Do you have any idea from where this `\_` is coming from in the first place? If not, then I can try finding it out. What do you suggest? 

For now, if you have addressed this in `ocw-to-hugo` via regex then I'll remove the custom js solution from here and keep the expand/collapse as it is. Correct?

MAbdurrehman12:
@gumaerc @pdpinch I have removed the custom js code as discussed [here](https://github.com/mitodl/ocw-hugo-themes/pull/364#discussion_r791401247), now this PR is only left with expand/collapse feature. I'll also update the PR title for better clarity.

base-theme/assets/index.ts
pdpinch: There's 20 years of data coming out of the legacy OCW system. If you give me an example course and file name (or URL) I can look into it and speculate where those unicode characters are coming from, but it probably doesn't matter. 

If it's easy to fix them up with a regex in the migration, great. If not, we can ask the OCW team to fix them when they see them. It might be helpful to know how many of these there are to inform the decision of automation vs. manual.

pdpinch:
Does this handle the case where the Title is missing or empty? I've heard that hugo doesn't have great error handling. 

The legacy data may have missing or null titles. It would also be good to avoid courses failing to publish to the staging from OCW studio just because they are missing a title.

gumaerc (approved):
:+1: LGTM

@pdpinch In this case, it's not necessary to check that `.Title` exists.  `.Title` is a standard property of the Hugo `Page` object, so even if you don't set this property in the front matter it is populated on the Page object as a blank string.

@MAbdurrehman12 In the future when running `len` or any other function on metadata whether it comes from the page front matter, the course data template, `getJSON` or any other metadata source we should be using a pattern to sanity check the data first.  `.Title` is an exception because of what I described above, but if it were another custom property that were to return `nil` for any reason, it would trigger an error.  Any error that happens during a Hugo build stops the whole build, so we have to be careful about checking data.  For instance, say you were trying to get the length of a front matter property called `caption`.  I would do this:

```
{{ $caption := .Params.caption | default "" }}
{{ if gt (len $caption) 75 }}
etc...
{{ end }}
```

The Hugo [`default`](https://gohugo.io/functions/default/) function allows you to specify a fallback value in the event that the value you're trying to source from is `nil` or empty.  Another approach is rebinding the context using [`with`](https://gohugo.io/functions/with/):

```
{{ with .Params.caption }}
{{ if gt (len .) 75 }}
etc...
{{ else }}
something different if .Params.caption is nil or empty
{{ end }}
{{ end }}
```

The point is to only create errors where it is necessary to stop the whole site build, and like I said it's fine to access core Hugo properties like `.Title` without checking them first but others do need to be checked.

FYI I merged [this PR](https://github.com/mitodl/ocw-to-hugo/pull/435) in `ocw-to-hugo` to handle the irregular whitespace characters.  Once there is an `ocw-to-hugo` release and the new data runs through the whole pipeline deployment process then the issue will be fixed.

base-theme/assets/index.ts
MAbdurrehman12: @pdpinch Yeah, let's fix them up via regex, which I think @gumaerc has already done for titles. 
Although after seeing a bit more into it, it looks like this issue is in multiple places and courses. For [example](https://github.mit.edu/raw/ocw-content-rc/21g.049-spring-2017/main/content/resources/21g-049s17-th.md?token=AAAFYX4CZ22R46M6V7YFQFLB7I34A). So I think it will be really hard to identify and do it manually so I guess we are good with fixing via regex.

MAbdurrehman12:
Thanks for the clarification @gumaerc, initially I thought that the data coming to hugo is already validated and does not need to be checked but you explained it pretty well 👍 

ps: What do you suggest? Should we merge this PR right away or wait for the `ocw-to-hugo` release so the issue is fixed before.

MAbdurrehman12:
@gumaerc Couple of days back, [this PR](https://github.com/mitodl/ocw-hugo-themes/pull/362/files) was merged in which hugo `len` function was used on an attribute but I did not sanity check it, as a default value was already being given [here in ocw-to-hugo](https://github.com/mitodl/ocw-to-hugo/blob/f11ee8585fdd73bf4cdcae346d54b5d6e8fcb975/src/data_template_generators.js#L26).

Should I place a check for it or is it fine. What do you suggest?

gumaerc:
> @gumaerc Couple of days back, [this PR](https://github.com/mitodl/ocw-hugo-themes/pull/362/files) was merged in which hugo `len` function was used on an attribute but I did not sanity check it, as a default value was already being given [here in ocw-to-hugo](https://github.com/mitodl/ocw-to-hugo/blob/f11ee8585fdd73bf4cdcae346d54b5d6e8fcb975/src/data_template_generators.js#L26).
> 
> Should I place a check for it or is it fine. What do you suggest?

I would open another PR and add a check there.  Something to keep in mind is that while currently `ocw-to-hugo` is the source of a majority of the data, soon sites will be authored using `ocw-studio`.  It's possible that sites could get published with missing data in the `draft` branch for preview, so we have to assume that any data can be missing.

MAbdurrehman12:
@gumaerc Alright. I'll create a PR with the check.
Thanks
btw should we merge this PR or wait for ocw-to-hugo release which has a fix you did?

gumaerc:
@MAbdurrehman12 You can feel free to merge this.  This PR will add the expand / collapse functionality to the page title and whenever the `ocw-to-hugo` PR gets released those non-breaking spaces will be removed, they're separate changes.

MAbdurrehman12:
> ocw-to-hugo

@gumaerc Is your `ocw-to-hugo` PR released? As I'm unable to test expand/collapse in title, introduced via this PR

PR #380, opened 2022-01-26
fix aria-live situation

### LINKED ISSUE #378: `aria-live` should only wrap the number of search results, not the full list of all results
for the search app, we currently wrap a _lot_ of page content with a `div` tag that has `aria-live="polite"`. This causes screenreaders to read out updates to the content within the tag. So while we correctly are tagging content that can update and which will be of use to someone using a screenreader, the amount of content that will need to be read out for most searches is very large with this approach.

Instead of wrapping the facet UI and the search results w/ `aria-live="polite"`, we should wrap just the results count and let the user tab through the results list to have course titles and so on be read out.

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #378 

#### What's this PR do?

The issue has a bit of information, but basically we had mis-used `aria-live` in a way that made the search page pretty terrible for using a screenreader. In short, all of the changes within a div marked `aria-live="polite"` are read out when they change, and we were wrapping both the search facets and the search results with that, so every time the results changed the screenreader would read out all that stuff.

Instead, a recommendation for dynamic search UIs like ours is to only wrap the result count with `aria-live="polite"`, so that when the search results update there is a notification to the user but they aren't deluged with a ton of stuff.

Optimally, try out the search page using a screenreader and confirm that the setup is better now (i.e. when you run a search you should just hear "43 results" or whatever, not a bunch of nonsense).

### DIFF
--- a/www/assets/js/components/SearchPage.test.tsx
+++ b/www/assets/js/components/SearchPage.test.tsx
@@ -10,7 +10,6 @@ import {
   LR_TYPE_COURSE,
   LR_TYPE_RESOURCEFILE
 } from "@mitodl/course-search-utils/dist/constants"
-import FilterableFacet from "./FilterableFacet"
 
 import SearchPage, { SEARCH_PAGE_SIZE } from "./SearchPage"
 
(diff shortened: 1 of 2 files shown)


abeglova (approved):
LGTM

PR #381, opened 2022-01-26
fix up role="search" situation

### LINKED ISSUE #379: search role should be set on the form which wraps search input
If you read the documentation on MDN for `role="search"` it is actually much more narrow than how we are currently (incorrectly) using it: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/search_role

We are using it right now more like a section, to wrap the entire portion of the page containing the search results:

Screen Shot 2022-01-26 at 10 50 53 AM

Instead, it should be set on a `form` tag which contains an `input` tag with `type="search"`.

- [ ] remove `role="search"` from the `div` that it is currently applied to
- [ ] set `role="search"` on the `form` tag which wraps our search box
- [ ] test out the usage w/ a screenreader on both the home page and the search page

alicewriteswrongs:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

closes #379 

#### What's this PR do?

- sets `role="search"` on the `form` tag which wraps the search input
- sets `type="search"` on the `input` tag itself

these two changes should signal to a screenreader user that the search box is, well, a search box, and not just a random text field. setting `role="search"` on the form should also allow that to be recognized as a functional region by screenreaders.

You can just confirm that the change makes sense, and possibly try out in a screenreader if you want.

Note that I removed the existing (erroneous) `role="search"` in #380

### DIFF
--- a/www/assets/js/components/SearchBox.test.tsx
+++ b/www/assets/js/components/SearchBox.test.tsx
@@ -14,8 +14,11 @@ describe("SearchBox component", () => {
     const wrapper = render({ onChange, value, onSubmit })
     expect(wrapper.find("input").prop("onChange")).toEqual(onChange)
     expect(wrapper.find("input").prop("value")).toBe(value)
+    expect(wrapper.find("input").prop("type")).toBe("search")
 
     expect(wrapper.find("form").prop("onSubmit")).toEqual(onSubmit)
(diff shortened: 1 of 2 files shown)


abeglova (approved):
👍

Issue #396, opened 2022-01-31
CHP image caption has expand widget even when all text is visible 


pdpinch:
### Steps to Reproduce

1. navigate to https://ocw-published.odl.mit.edu/courses/1-017-computing-and-data-analysis-for-environmental-applications-fall-2003/ (in Firefox) or https://ocwnext-rc.odl.mit.edu/courses/1-017-computing-and-data-analysis-for-environmental-applications-fall-2003/

### Expected Behavior

- no expand widget, because the entire caption fits in the available area

### Actual Behavior

- there is an expand widget. Clicking on it expands the box, but doesn't reveal any more text.

### Screenshot or Screencast

<img width="331" alt="image" src="https://user-images.githubusercontent.com/430126/151806646-ffd341dc-81e9-466c-ad86-e97d08ac572f.png">

MAbdurrehman12:
This issue is resolved [here](https://ocwnext-rc.odl.mit.edu/courses/1-017-computing-and-data-analysis-for-environmental-applications-fall-2003/) but since the content can have HTML elements which are counted in hugo `len` function but not shown to the user, so it is possible that this issue occurs for any course.
This can be resolved, later on, by a custom JS solution as discussed [here](https://github.com/mitodl/ocw-hugo-themes/pull/407)

PR #407, opened 2022-02-02
fix: increased value for expand widget for course image description

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/396

#### What's this PR do?
- Increases length value for course caption comparison
- Adds some bottom margin when the expand widget is not showing so the caption does not stick/or come too close with the border

- Go to any course homepage
- Ensure that expand/collapse is only shown and working when the course image caption is **not** adjusted in the available area.
- Please repeat this for maximum courses or change the caption in course content

image

image

image

### DIFF
--- a/course/layouts/home.html
+++ b/course/layouts/home.html
@@ -30,7 +30,8 @@
                   <div class="homepage-content d-flex">
                     <div class="container-fluid pt-0 pb-4">
                       <div class="row px-3 pb-2 justify-content-between">
-                        {{ $shouldCollapseImageDescription := gt (len $courseImageMetadata.Params.image_metadata.caption) 130 }}
+                        {{- $imageCaption := $courseImageMetadata.Params.image_metadata.caption -}}
+                        {{ $shouldCollapseImageDescription := gt (len $imageCaption) 130 }}
                         <div class="col-lg-4 d-none d-lg-block pl-0">
                           <div class="d-flex flex-column image-with-caption">
                             <img class="course-image" src="{{ $courseImageUrl }}"
@@ -39,7 +40,7 @@
                             <div id="course-image-description" class="{{ if $shouldCollapseImageDescription }}collapse{{ else }} mb-3 {{ end }}" aria-expanded="false">
                               <div class="caption pt-3 px-3">
                                 <span>
-                                  {{ $courseImageMetadata.Params.image_metadata.caption | .RenderString }}
+                                  {{ $imageCaption | .RenderString }}
                                 </span>
                               </div>
                             </div>
--- a/course/layouts/home.html
+++ b/course/layouts/home.html
@@ -30,13 +30,13 @@
                   <div class="homepage-content d-flex">
                     <div class="container-fluid pt-0 pb-4">
                       <div class="row px-3 pb-2 justify-content-between">
-                        {{ $shouldCollapseImageDescription := gt (len $courseImageMetadata.Params.image_metadata.caption) 100 }}
+                        {{ $shouldCollapseImageDescription := gt (len $courseImageMetadata.Params.image_metadata.caption) 130 }}
                         <div class="col-lg-4 d-none d-lg-block pl-0">
                           <div class="d-flex flex-column image-with-caption">
                             <img class="course-image" src="{{ $courseImageUrl }}"
                               alt="{{ index $courseImageMetadata.Params.image_metadata "image-alt" }}"
                             />
-                            <div id="course-image-description" class="{{ if $shouldCollapseImageDescription }}collapse{{ end }}" aria-expanded="false">
+                            <div id="course-image-description" class="{{ if $shouldCollapseImageDescription }}collapse{{ else }} mb-3 {{ end }}" aria-expanded="false">
                               <div class="caption pt-3 px-3">
                                 <span>
                                   {{ $courseImageMetadata.Params.image_metadata.caption | .RenderString }}


course/layouts/home.html
```diff
@@ -30,13 +30,13 @@
                   <div class="homepage-content d-flex">
                     <div class="container-fluid pt-0 pb-4">
                       <div class="row px-3 pb-2 justify-content-between">
-                        {{ $shouldCollapseImageDescription := gt (len $courseImageMetadata.Params.image_metadata.caption) 100 }}
+                        {{ $shouldCollapseImageDescription := gt (len $courseImageMetadata.Params.image_metadata.caption) 130 }}
                         <div class="col-lg-4 d-none d-lg-block pl-0">
```
MAbdurrehman12, line 34: Bug finding: Some `caption` also have HTML in them (usually anchor tag) which is rendered by `.RenderString` and not shown to users but it is included in `len` function, hence caption length becomes greater and expand/collapse is shown but the entire caption fits in the available area and this issue occurs. (`len` function is not accurate on `.RenderString` due to whitespaces in between so comparison length value has been increased but I'd love to hear if there's any other more appropriate solution)

@gumaerc Would you like to share some thoughts?

gumaerc: You might be able to pass the rendered string to a variable first before assessing the length.  Something like:

```
{{ $renderedCaption := $courseImageMetadata.Params.image_metadata.caption }}
{{ $shouldCollapseImageDescription := gt (len $renderedCaption) 130 }}
```

MAbdurrehman12: > You might be able to pass the rendered string to a variable first before assessing the length. Something like:

Okay, I'll update the PR.

> This strikes me as something that should maybe be solved in the frontend though anyway, rather than during templating. Is there some kind of JS / CSS solution we can use to respond to containers with a certain class going over a certain height and add expand collapse that way?

Yeah, We can do it with JS as well, but a generic solution will be required which is applicable to any section which requires expand/collapse, but I guess, the current solution will be more efficient than the JS one.

MAbdurrehman12: Although if it's not possible to get the raw length of what would be the visible text in the browser after rendering and if the frequency of this issue is more then I guess we should consider something else.

btw I also tried something like this: 
`(len caption | .RenderString)`
but it also not accurate as it seemed like that link html has been replaced by whitespaces

MAbdurrehman12:
@gumaerc shall we proceed by merging with this PR or would you suggest some alternative JS solution?

course/layouts/home.html
gumaerc: `.RenderString` tells Hugo to generate HTML from a given Markdown string at build time, so that would probably be even more verbose.  Ultimately trying to determine the height based on character length is difficult because really that's up to the browser.  At render time, CSS could change properties like the font size or line height which would affect the height.  Ultimately, I think the only truly reliable way to do this is with JS client side by measuring the height.

gumaerc:
@MAbdurrehman12 I did some more research on this and it looks like you can use the Hugo [countwords](https://gohugo.io/functions/countwords/) function.  I did a brief test adding the following code to a template:

```
<h1>Count the words: {{ "Count the words" | countwords }}</h1>
<h1>Count the [words](https://www.google.com): {{ "Count the [words](https://www.google.com)" | countwords }}</h1>
```

...and got this on the page:
image

So, it looks like `countwords` ignores links and you should be able to safely use that to estimate how many "words" are in a string.  Of course, a "word" could be any number of characters long, but it ignores links at least.  I would give that a shot, and if it ends up being hard to make accurate a JS based solution that looks at `clientHeight` of the container element at runtime in the browser might be necessary.

MAbdurrehman12:
@gumaerc Alright. 
Thanks for looking into all this.

This is a good suggestion, though I think I tried the `count words` approach but can't remember the issue which I faced, anyways, I'll give it another shot, and let's see how things goes. 
JS solution will require some work to be done but if accuracy becomes a problem then we will have to go with it.

MAbdurrehman12:
@gumaerc So I replaced the `len` function with `count words` and changed the threshold value and height accordingly, but it looks like this issue has become harder to handle due to the inaccuracy as the content is completely dynamic, content with more small words like `a`, `the`, `,`, `&` etc meets the threshold but they are adjusted in the available area, leaving behind an unnecessary expand/collapse widget.

I guess, `count words` can be suitable for the content which has a greater length, unlike image caption, as estimation can work there.
What do you suggest now? 

I think replacing all this with a JS solution will certainly require some work to be done so we can proceed with it if this issue is occurring for multiple courses. I'm also not sure about the priority of this issue.
Perhaps, if @pdpinch could share his thoughts too?

gumaerc:
@MAbdurrehman12 I had a feeling this would be the case.  Even if we could get an accurate character count of the text to be rendered in the browser, I don't think character count in general is a reliable way to approach this because of potential variability in height because of potential changes in font size / weight / line height / margin / padding / etc.  The way I see it we have 3 options:

 - Client side JS solution involving reacting to a length check on the `innerText` property of the caption container.  This might work, but could also be not so ideal because of the drawbacks of a character count approach I described above as well as potential layout thrashing due to having to apply the expand collapse after the text is rendered.
 - Client side JS solution involving measuring `clientHeight` of the container after rendering, of which a potential drawback is layout thrashing by applying the expand / collapse after the DOM is rendered.  The benefit of this solution over the character solution is that if it's reacting to the container height, changes to the style of the text inside wouldn't affect the height at which the expand / collapse is triggered.
 - Using a library like http://marcj.github.io/css-element-queries/: I had looked into this library when writing `partial_collapse_list.html` but didn't end up using it because it seemed like a heavy handed solution for a fairly simple problem.  I also remember having some technical issues with it, but you may want to try it yourself and see if it works out.  The benefit of using a solution like this is that we could write media query type CSS targeted at specific elements rather than just the whole window.

gumaerc (approved):
As @pdpinch said @MAbdurrehman12, we should just merge this as-is for now.  Later on we can re-visit a more advanced solution.

PR #410, opened 2022-02-03
Wassaf/404 add search icon

### LINKED ISSUE #404: All page headers should have search icon
... except for the home page and the search page, where the search icon would be redundant in the header. 

Course pages already have the search icon in the header, for example, https://ocwnext-rc.odl.mit.edu/courses/16-885j-aircraft-systems-engineering-fall-2004/

#### Acceptance Criteria:

- [ ] about page
- [ ] educator page
- [ ] contact page
- [ ] newsletter page
- [ ] testimonial pages
- [ ] privacy and terms of service page, e.g. https://ocw-draft-qa.global.ssl.fastly.net/pages/privacy-and-terms-of-use/

#### Out of Scope

Course pages already have the search icon in the header, for example, https://ocwnext-rc.odl.mit.edu/courses/16-885j-aircraft-systems-engineering-fall-2004/

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

Closes #404

#### What's this PR do?
- adds search icon to required pages

- Verify all pages have search icon
- Verify home page and search page dont have search icon

Screenshot 2022-02-03 at 1 28 33 PM

### DIFF
--- a/base-theme/layouts/partials/header.html
+++ b/base-theme/layouts/partials/header.html
@@ -61,7 +61,7 @@
       </div>
     </div>
     <div class="right">
-      {{ block "extraheader" . }}{{ end }}
+        {{ block "extraheader" . }}{{ end }}
       <a
         class="d-flex give-button-header py-2 px-3"
         href="https://giving.mit.edu/give/to/ocw/?utm_source=ocw&utm_medium=homepage_banner&utm_campaign=nextgen_home"
--- a/base-theme/layouts/partials/extraheader.html
+++ b/base-theme/layouts/partials/extraheader.html
@@ -1,0 +1,8 @@
+{{ define "extraheader" }}
+  {{ $renderSearchIcon := index .Params "renderSearchIcon" | default true}}
+  {{ if $renderSearchIcon}}
+    <a class="nav-link search-icon pr-6" href="/search">
+      <i class="material-icons">search</i>
+    </a>
+  {{end}}
+{{ end }}
--- a/course/layouts/partials/extraheader.html
+++ b/course/layouts/partials/extraheader.html
@@ -1,5 +1,0 @@
-{{ define "extraheader" }}
-<a class="nav-link search-icon pr-6" href="/search">
-  <i class="material-icons">search</i>
-</a>
-{{ end }}
--- a/www/content/_index.md
+++ b/www/content/_index.md
@@ -1,5 +1,3 @@
 ---
-title: Search
-type: search
 renderSearchIcon: false
 ---
--- a/www/layouts/home.html
+++ b/www/layouts/home.html
@@ -1,6 +1,9 @@
 {{ define "main" }}
 <div id="home-header">
-{{ block "header" . }}{{ partial "header" . }}{{ end }}
+  {{ block "header" . }}
+  {{ $renderSearchIcon := index .Params "renderSearchIcon" | default true }}
+  {{ partialCached "header" . $renderSearchIcon }}
+  {{ end }}
 </div>
 <div class="d-flex flex-column align-items-center home-banner" style="background-image: url(/images/homepage_hero.jpg);">
   <div class="banner-contents-container mx-auto d-flex justify-content-start justify-content-md-around w-100 align-items-center h-100">
--- a/www/layouts/search/section.html
+++ b/www/layouts/search/section.html
@@ -1,8 +1,9 @@
 {{ define "main" }}
 <div class="search-wrapper">
   {{ block "header" . }}
-  {{ partialCached "header" . }}
-  {{end}}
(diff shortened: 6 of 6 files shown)


pdpinch:
I'd like to get @gumaerc's input on this approach before it's merged.

mbertrand (approved):
Looks great 👍

gumaerc (changes requested):
I think we should go with a different approach here.  First of all, partials like this that will likely return the same thing every time should be called with `partialCached`. 
 Secondly, removing the `.` as an argument means the `header.html` partial does not receive a context object.  This means that the partial template will not have access to the calling page or any of the site wide data.  This also creates a situation where the header partial will be receiving inconsistent arguments.  If you search `ocw-hugo-themes` for `{{ partialCached "header" . }}` you'll see that there are a lot of places where it's called this way.  A better approach here would be to store this variable in the front matter of pages where it needs to be set.  For instance, in `www/content/search/_index.md` you could set:

``` 
---
title: Search
type: search
notRenderSearchIcon: true
---
```

Then, at `www/content` you could create an `_index.md` file with:

```
---
notRenderSearchIcon: true
---
```

The code `{{if ne (isset . "notRenderSearchIcon") true}}` also does not actually check the value of this variable.  It only checks for the existence of it.  Regardless, the search icon may be the only thing in `extraheader` right now, but we might want to add more to this partial in the future and then this flag would hide that as well.  Instead, if we preserve passing only the context variable, in `extraheader.html` we can do something like this in combination with the changes I described above:

```
{{ define "extraheader" }}
{{ $notRenderSearchIcon := index .Params "notRenderSearchIcon" | default false }}
{{ if not $notRenderSearchIcon }}
<a class="nav-link search-icon pr-6" href="/search">
  <i class="material-icons">search</i>
</a>
{{ end }}
{{ end }}
```

This way, any particular type of page can be marked to not render the search icon in the content itself and you can default it to false easily.

Wassaf-Shahzad:
> I think we should go with a different approach here. First of all, partials like this that will likely return the same thing every time should be called with `partialCached`. Secondly, removing the `.` as an argument means the `header.html` partial does not receive a context object. This means that the partial template will not have access to the calling page or any of the site wide data. This also creates a situation where the header partial will be receiving inconsistent arguments. If you search `ocw-hugo-themes` for `{{ partialCached "header" . }}` you'll see that there are a lot of places where it's called this way. A better approach here would be to store this variable in the front matter of pages where it needs to be set. For instance, in `www/content/search/_index.md` you could set:
> 
> ```
> ---
> title: Search
> type: search
> notRenderSearchIcon: true
> ---
> ```
> 
> Then, at `www/content` you could create an `_index.md` file with:
> 
> ```
> ---
> notRenderSearchIcon: true
> ---
> ```
> 
> The code `{{if ne (isset . "notRenderSearchIcon") true}}` also does not actually check the value of this variable. It only checks for the existence of it. Regardless, the search icon may be the only thing in `extraheader` right now, but we might want to add more to this partial in the future and then this flag would hide that as well. Instead, if we preserve passing only the context variable, in `extraheader.html` we can do something like this in combination with the changes I described above:
> 
> ```
> {{ define "extraheader" }}
> {{ $notRenderSearchIcon := index .Params "notRenderSearchIcon" | default false }}
> {{ if not $notRenderSearchIcon }}
> <a class="nav-link search-icon pr-6" href="/search">
>   <i class="material-icons">search</i>
> </a>
> {{ end }}
> {{ end }}
> ```
> 
> This way, any particular type of page can be marked to not render the search icon in the content itself and you can default it to false easily.

I went over this approach and discovered a few things. Caching a partial usually saves its first instance and renders that one every where else. Which causes problems in rendering the `extraheader` dynamically since if you visit the home page first and the Header is rendered without  `extraheader`, every subsequent page will render that particular  cached header ie the one without the `extraheader` part.
To overcome this issue we need to make variants of the header (or any cached partial) depending on the `renderSearchIcon` Param (I renamed notRenderSearchIcon since it was not very reader friendly ).
Which lead me to the realisation that rather then manually adding header and footer in every new `section` (Contact, About,Newsletter etc) we should abstract the  include statements  in a block and reuse them rather then redefining them multiple times in other pages.
Your thoughts @gumaerc ?

base-theme/layouts/partials/header.html
```diff
@@ -61,7 +63,9 @@
       </div>
     </div>
     <div class="right">
-      {{ block "extraheader" . }}{{ end }}
+      {{if ne (isset . "notRenderSearchIcon") true}}
```
gumaerc, line 66: This can be reverted too.

gumaerc (approved):
:+1: LGTM

Issue #302, opened 2021-12-14
Resource link on a line by itself does not render with a line break. 


pdpinch:
### Steps to Reproduce

1. Create a page in studio with multiple resources on separate lines, like https://ocw-studio-rc.odl.mit.edu/sites/peters-test-2021-09-29-1/type/page/

<img width="805" alt="image" src="https://user-images.githubusercontent.com/430126/146078601-c25faca7-c1d3-4b89-864a-3c26e1f11acf.png">

2. Publish it. This will create three resource short codes, each on their own line, like https://github.mit.edu/ocw-content-rc/pinch-test-2021-09-29-01/blob/main/content/pages/resource-links-line-breaks.md

<img width="822" alt="image" src="https://user-images.githubusercontent.com/430126/146078620-06832925-63fd-4fe4-ac4f-d4207803e17b.png">

3. Examine the published output. The three links are run together on one line: https://ocw-draft-qa.global.ssl.fastly.net/courses/peters-test-2021-09-29-1/pages/resource-links-line-breaks/

<img width="555" alt="image" src="https://user-images.githubusercontent.com/430126/146078698-b8fd4c8c-0e6b-40f6-b3e5-67383396d9e0.png">

### Expected Behavior

The published output should look more like the WYSIWYG editor, with line breaks between each source code. 

### Actual Behavior

It seems like the short code somehow swallows the `\n` between each item.

alicewriteswrongs:
This occurs because of the logic that Hugo uses to decide when content in Markdown should be wrapped in a `<p>` tag and when it should not. When shortcodes are on a line by themselves, without any other text content, Hugo assumes that they should not be wrapped in a `<p>` tag, which generally makes sense because shortcodes are often used to render block-level content (like iframes or the like) which shouldn't be inserted into a paragraph or otherwise treated like text.

In our case, however, our `resource_link` shortcode renders an anchor tag, which can be treated like inline (non-block) content just fine. But I think from Hugo's perspective our use-case is an edge-case / exception. From my research on the Hugo forums and issues and so on there is not, currently, a way to tell Hugo that the rendered content of a shortcode should be treated as inline (text) content in this case.

A workaround would be to tell content authors that if they want something like this, then the best way to do it is to use a list, which will be properly formatted as they expect.

pdpinch:
@gumaerc is addressing this with #485

pdpinch:
@gumaerc you merged #485. I think this can be closed?

[...]

PR #418, opened 2022-02-07
fix: line break on resource links

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

- Closes https://github.com/mitodl/ocw-hugo-themes/issues/302

#### What's this PR do?
- Adds display block class to resource link shortcode so it renders with a link break
- Note: TBH I'm not confident with this solution as it will also cause the resource_links in a text/paragraph to be displayed in a new line which might look a bit odd. Something like this maybe: 
image

- Repeat the steps mentioned in the relevant ticket above and verify that the resource links have a line break and they do not render.
- Also verify that the resource links in paragraphs/texts do not look odd as they will be rendered in a new line.

image

### DIFF
--- a/course/layouts/shortcodes/resource_link.html
+++ b/course/layouts/shortcodes/resource_link.html
@@ -1,7 +1,5 @@
 {{- $uuid := index .Params 0 -}}
 {{- $title := index .Params 1 -}}
 {{- range where $.Site.Pages "Params.uid" $uuid -}}
-    <p class="m-0 p-0">
-        <a href="{{- .Permalink -}}">{{ $title }}</a>
-    </p>    
+    <a class="d-block" href="{{- .Permalink -}}">{{ $title }}</a>
 {{- end -}}
--- a/course/layouts/shortcodes/resource_link.html
+++ b/course/layouts/shortcodes/resource_link.html
@@ -1,5 +1,7 @@
 {{- $uuid := index .Params 0 -}}
 {{- $title := index .Params 1 -}}
 {{- range where $.Site.Pages "Params.uid" $uuid -}}
-    <a href="{{- .Permalink -}}">{{ $title }}</a>
+    <p class="m-0 p-0">
+        <a href="{{- .Permalink -}}">{{ $title }}</a>
+    </p>    
 {{- end -}}


course/layouts/shortcodes/resource_link.html
```diff
@@ -1,5 +1,5 @@
 {{- $uuid := index .Params 0 -}}
 {{- $title := index .Params 1 -}}
 {{- range where $.Site.Pages "Params.uid" $uuid -}}
-    <a href="{{- .Permalink -}}">{{ $title }}</a>
+    <a class="d-block" href="{{- .Permalink -}}">{{ $title }}</a>
```
MAbdurrehman12, line 4: tbh, I'm not liking this solution as it will cause the resource links in paragraphs/texts to be displayed in an entirely new line. 
I spent some time but couldn't figure out something better. Maybe we can make some customizations in the editor to render a new line **only** when resource links are placed/created in different lines.
Perhaps, @gumaerc @alicewriteswrongs can share their thoughts

alicewriteswrongs: yes I think you're right, we don't want to introduce a solution to one line-break issue that creates another one

gumaerc: I agree, we shouldn't be setting these links to be block level elements because it will cause unintended line breaks when the goal here is to solve unintentionally _not_ inserting line breaks.  @alicewriteswrongs noted on the original issue that a potential short term solution would be to instruct content editors to put a resource link inside a list if it's by itself, even if it's only one link.  This would work, and the only way I can think of that might be better would be if we can somehow recognize that a resource link is on a line by itself in CKEditor's turndown rule, maybe by looking at adjacent children of the node or something.  If we could do that, we could add an optional boolean argument to `resource_link.html` called `inline` or something that defaults to `true`.  If we could catch that the resource link is being rendered on a line by itself, we could set that to `false` and then intentionally render a line break in that case?

MAbdurrehman12: @gumaerc Yeah, I also thought about this approach but there were some use-cases in which this approach might not be suitable.

pdpinch:
@gumaerc figured out another way to do this, with a combination of #485 and https://github.com/mitodl/ocw-studio/pull/1067

PR #422, opened 2022-02-07
add simple subscript, superscript shortcodes

ChristopherChudzicki:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

#### What's this PR do?
Adds`{{< sup "some value" >}}` and `{{< sub "some value" >}}` shortcodes to the base theme.

Run this repo with a local course and try adding the new shortcodes to some markdown files. For example:
```
E=mc{{< sup "2" >}} everyone likes that.

quotes dog{{< sub "**wo\"of**" >}} woof bark

The nuts and bolts of preparing a Business Plan and Starting a New Venture Plan was explored in this 25{{< sup "th" >}}  annual 
```

These do not 
1. handle nesting
2. or markdown (e.g., bold) inside the super/subscript

value right now. Let me know if you have suggestions/feedback about that. In the `ocw-to-hugo` sample courses, neither of these seems common. Possibly (2) could be handled with `| markdownify`... (1) might require an opening/closing shortcode? Unsure. 

Screen Shot 2022-02-07 at 5 06 08 PM


alicewriteswrongs:
on your questions, I think doing 1. would require having an opening and closing shortcode, like what we do for the tables. The reason we have to do that is that when rendering markdown via `markdownify` Hugo doesn't render shortcodes, but just does normal markdown rendering, so `markdownify` works great if you've only got built-in markdown syntax but won't work if you have shortcodes. On 2. you should be able to support basic markdown syntax via `| markdownify`.

MAbdurrehman12 (approved):
Looks good 👍 
Apart from point# 1 in background context which doesn't seems to be a common use-case, though I'm not sure I'm able to pick what Alice stated

PR #445, opened 2022-02-16
fixed video urls not linking to archive.org

### LINKED ISSUE #443: video download links failing
### Steps to Reproduce

1. Visit https://ocwnext.odl.mit.edu/courses/22-01-introduction-to-nuclear-engineering-and-ionizing-radiation-fall-2016/resources/radiation-history-to-the-present2014understanding-the-discovery-of-the-neutron/
2. Click download video button

### Expected Behavior

- According to https://github.com/mitodl/ocw-hugo-themes/issues/145#issuecomment-953880005 we should be linking to the Internet Archive (archive.org) to download the videos. 

### Actual Behavior

- The link appears to be a root relative path

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing

Closes #443 

#### What's this PR do?
Adds conditional in resource_link partial to skip processing links to archive.org. 

- Open video lectures for course `22.01-fall-2016`
- Click `Download Video` and verify they it opens archive.org download page.
- Verify videos from s3 bucket behave as it is

### DIFF
--- a/base-theme/layouts/partials/resource_url.html
+++ b/base-theme/layouts/partials/resource_url.html
@@ -6,13 +6,9 @@
 
   The RESOURCE_BASE_URL env variable is prefixed before the URL. That way for local
   development, you can set this to the S3 URL you expect the resources to be available at.
-
-  If the URL is for archive.org then we skip the above mentioneds steps.
 */}}
-{{- if not (hasPrefix $url "https://archive.org") -}}
-  {{- $prefix := getenv "RESOURCE_BASE_URL" | default "" -}}
-  {{- $url := printf "%s%s" (strings.TrimSuffix "/" $prefix) (urls.Parse $url).Path -}}
-{{- end -}}
+{{- $prefix := getenv "RESOURCE_BASE_URL" | default "" -}}
+{{- $url := printf "%s%s" (strings.TrimSuffix "/" $prefix) (urls.Parse $url).Path -}}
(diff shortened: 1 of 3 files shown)


MAbdurrehman12 (approved):
Looks good but I may not be familiar with all the use-cases/contexts so not sure if this approach is fine for all the use-cases hence another review from someone else will be good, which you have already requested.

I'm also assuming that all the video urls of legacy courses will always have this `https://archive.org` prefix.
I think it will also be good to move this constant to a more generic place or in `env` as you suggested.

pdpinch:
@abeglova can you review this too?

abeglova:
I don't think it makes sense to make this change in partials/resource_url.html. This partial should just be used for internal urls

 I think it makes a lot more sense to make the update in `course/layouts/partials/video.html`. The link there should use `partial "resource_url" ` if the video is stored on s3 and just link to the video directly if the video is on archive.com

Wassaf-Shahzad:
> I don't think it makes sense to make this change in partials/resource_url.html. This partial should just be used for internal urls
> 
> I think it makes a lot more sense to make the update in `course/layouts/partials/video.html`. The link there should use `partial "resource_url" ` if the video is stored on s3 and just link to the video directly if the video is on archive.com

I added the conditional in the partial as video_embed also uses archive links but I will move it.

abeglova (approved):
LGTM

PR #457, opened 2022-02-21
added optional and related resources tab

### LINKED ISSUE #454: Rendering related resource, optional tab sections of video player page
Some videos pages on the legacy OCW site include optional "related reource" or "optional tab" content. We recently added this content to the markdown output from studio. 

legacy video page with related resource: 
https://ocw.mit.edu/courses/mathematics/18-06-linear-algebra-spring-2010/video-lectures/lecture-4-factorization-into-a-lu/
markdown for this video: https://github.mit.edu/ocw-content-rc/18.06-spring-2010/blame/main/content/resources/lecture-4-factorization-into-a-lu.md#L11

legacy video page with an optional tab, labelled "slides"
https://ocw.mit.edu/courses/nuclear-engineering/22-033-nuclear-systems-design-project-fall-2011/projects/core-sub-task-presentation-ii/
markdown for this video: https://github.mit.edu/ocw-content-rc/22.033-fall-2011/blame/main/content/resources/biofuels-sub-task-presentation-ii.md#L7-L8

#### Designs and Mockups

image

#### Acceptance Criteria:

- [ ] When there is content, add a "related resources" section below the Transcript, with a similar UX and with the label "Related Resources". It should default to closed, but can be opened to view the content.  
- [ ] When there is content, add an optional section below the video, using the label specified in the page markdown. It should default to closed, but can be opened to view the content. 
- [ ] All sections should default to closed, including the transcript.

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

Closes #454 

#### What's this PR do?
 Adds a `related resources` and `optional tab` in video page.

Verify the following
 - When there is content, add a "related resources" section below the Transcript, with a similar UX and with the label "Related Resources". It should default to closed, but can be opened to view the content.
 - When there is content, add an optional section below the video, using the label specified in the page markdown. It should default to closed, but can be opened to view the content.
 - All tabs remain close on page load
 
(Optional)
Screenshot 2022-02-21 at 6 23 49 PM
Screenshot 2022-02-21 at 6 24 03 PM
Screenshot 2022-02-21 at 6 25 06 PM
Screenshot 2022-02-21 at 6 25 19 PM

### DIFF
--- a/course/layouts/partials/video.html
+++ b/course/layouts/partials/video.html
@@ -2,6 +2,9 @@
 {{ $captionsLocation := .Params.video_files.video_captions_file }}
 {{ $transcriptPdfLocation := .Params.video_files.video_transcript_file }}
 {{ $downloadLink := .Params.file }}
+{{ $relatedResourses := .Params.related_resources_text | default "" }}
+{{ $optionalTabTitle := .Params.optional_tab_title | default "" }}
+{{ $optionalTabContent := .Params.optional_text | default ""}}
 {{ if (eq $downloadLink nil) }}
 {{ $downloadLink = .Params.video_files.archive_url }}
 {{ else }}
@@ -28,7 +31,7 @@
 
   {{if $captionsLocation}}
   <div class="transcript-box">
-  	<div class="transcript-header">
+  	<div class="tab-header">
 	   	<span>
 	       Transcript
 	    </span>
@@ -40,8 +43,48 @@
 	      <i class="material-icons md-18"></i>
 	    </a>
 	</div>
-	<div class="transcript collapse show"></div>
+	<div class="transcript collapse"></div>
   </div>
-  <div class="transcript-footer"></div>
+  <div class="tab-footer"></div>
+  {{end}}
+  {{if $relatedResourses }}
+  <div class="related-resource-box">
+  	<div class="tab-header">
+	   	<span>
+	       Releated Resources
+	    </span>
+	    <a class="transcript-section-toggle course-nav-section-toggle"
+	      href=".related-resource"
+	      data-toggle="collapse"
+	      aria-controls="related-resource"
+	      aria-expanded="">
+	      <i class="material-icons md-18"></i>
+	    </a>
+	</div>
+	<div class="related-resource  tab-content collapse">
+		{{ .Page.RenderString $relatedResourses}}
+	</div>
+  </div>
+  <div class="tab-footer"></div>
+  {{end}}
+  {{if and $optionalTabTitle $optionalTabContent }}
+  <div class="related-resource-box">
+  	<div class="tab-header">
+	   	<span>
+	       {{ $optionalTabTitle }}
+	    </span>
+	    <a class="transcript-section-toggle course-nav-section-toggle"
+	      href=".related-resource"
+	      data-toggle="collapse"
+	      aria-controls="related-resource"
+	      aria-expanded="">
+	      <i class="material-icons md-18"></i>
+	    </a>
+	</div>
+	<div class="related-resource  tab-content collapse">
+		{{ .Page.RenderString $optionalTabContent}}
+	</div>
+  </div>
+  <div class="tab-footer"></div>
   {{end}}
 </div>
--- a/course/layouts/partials/video.html
+++ b/course/layouts/partials/video.html
@@ -29,7 +29,7 @@
 	{{end}}
 	</div>
 
-  {{if $captionsLocation}}
+  {{ if $captionsLocation}}
   <div class="transcript-box">
   	<div class="tab-header">
 	   	<span>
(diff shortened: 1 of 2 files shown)


course/layouts/partials/video.html
```diff
@@ -2,6 +2,9 @@
 {{ $captionsLocation := .Params.video_files.video_captions_file }}
 {{ $transcriptPdfLocation := .Params.video_files.video_transcript_file }}
 {{ $downloadLink := .Params.file }}
+{{ $relatedResourses := .Params.related_resources_text }}
```
MAbdurrehman12, line 5: I wonder why there's no need to give a default value here as the code does not break if this variable does not exist. Maybe these variables are defined in the `front matter`?

MAbdurrehman12 (approved):
Looks good! 

Few things in mind:

1- If, in future, these tab increases then maybe we can make an array and loop over it to avoid html code duplication.
2- For [course 18.06-spring-2010](https://github.mit.edu/ocw-content-rc/18.06-spring-2010/blame/main/content/resources/lecture-4-factorization-into-a-lu.md#L11) the links of `readings` and `table of content` in `related resource` tab are broken. Since it is `[Readings](pages/readings)` so it considers the current URL as base URL and takes to `*baseurl*/resources/lecture-33-left-and-right-inverses-pseudoinverse/pages/readings` where as the correct URL should be `*baseurl*/pages/readings/`. I assume this should be fixed from studio so I'm approving this PR, but I have a doubt about it so it's good to get it clarified before merging this PR

course/layouts/partials/video.html
Wassaf-Shahzad: - The reason we cannot loop over them is mainly both of them behave in somewhat different  ways since related resources has a fixed title but a dynamic content while optional resources have a dynamic title and dynamic content.
The better way to remove duplication would be to create a partial which takes this values as parameters and call them here.
-  These variables are defined in front matter and  are somewhat guaranteed, if not then we would have to add the check for every variable used here

MAbdurrehman12: - Both, dynamic and fixed content can be handled with array and loop but creating a partial is a more relevant and better approach here I guess.
- Yeah, they exist in the front matter but even if a variable does not exist `(e.g: .Params.anything_here)`, the code still does not break. I wonder why? 😄

pdpinch:
@gumaerc can you weigh in here, specifically on:
1. what should we do about links in the tab content? My understanding is that most of the tab content is links
2. based on the data we currently have in production, what do we need to do to protect against missing variables?

gumaerc:
@pdpinch @Wassaf-Shahzad @MAbdurrehman12 :

1. The links in `related_resources_text` are being resolved properly to the relative page they're supposed to point to from `resolveuid` links in the parsed JSON.  This is a situation though where in the past we would have prepended the links with `{{< baseurl >}}`, but we can't do that here (or use any shortcodes) because the markdown is being written into front matter. See [here](https://github.com/gohugoio/hugo/issues/6703) for more info on that.  Most likely, we will have to do some kind of transformation of these links on the `ocw-studio` side, at least until we can support shortcodes in front matter.  Previously, the only time we ran into this situation was with the course description, and we left the links as-is because the only place the course description is only rendered on the course home page.  In this case though, the links are rendered on a sub-page so as you've seen they don't work.  What we might have to end up doing in the short term is re-write the links to be root-relative.  For example, instead of `pages/readings/` we would have studio write in `/courses/18-06-linear-algebra-spring-2010/pages/readings`.  This would break the links if you were running the course locally for development, but we could work around that.  Hopefully at some point rendering shortcodes with `.RenderString` is implemented.
2. As far as variable checking goes. as you've noticed here you can assign undefined properties of `.Params` to a variable, i.e. `$shouldNotExist := .Params.should_not_exist` doesn't cause an error.  Even rendering that property with `{{ .Params.should_not_exist }}` will not cause an error.  What _will_ cause an error is trying to pass that value to functions like `len` or trying to pass the value to a partial.  Because of this, we should be sanity checking all of our variables by default because if someone else makes a PR in the future to try and pass these variables into a partial or run a standard Hugo function against them, there could be build-breaking errors if they are not set.

pdpinch:
@Wassaf-Shahzad if this is ready for review again, please change the label and assign it back to @gumaerc 

I'm eager to see this go out, so I think we'll have to live with the linking issues that Carey mentioned in #1 above in https://github.com/mitodl/ocw-hugo-themes/pull/457#issuecomment-1048976746

course/layouts/partials/video.html
```diff
@@ -28,7 +31,7 @@
 
   {{if $captionsLocation}}
```
gumaerc, line 32: In this and other places where you're calling template logic, the standard I've been following is to leave 1 space on either side of the curly braces.  For example, `{{ if $captionsLocation }}`.  I'm not sure how necessary this is, but in every sample of Hugo template code I've ever seen this pattern is always followed.  Again I'm not sure exactly why or if it even matters, but I follow this pattern because it's what I see in every example I've ever looked at.

gumaerc (approved):
The defaults on variable assignment look good, and as @pdpinch said we'll just have to live with the linking issues for now.  One small nitpick though:

PR #468, opened 2022-02-22
Fixed page_title partial caching the Title

### LINKED ISSUE #458: Incorrect heading on ocw-www pages
### Steps to Reproduce

1. Make an edit to a page, like https://ocw-studio-rc.odl.mit.edu/sites/ocw-www/type/pages/
2. publish the site to draft
3. Look at the published version of the page you changes, for example https://ocw-draft-qa.global.ssl.fastly.net/pages/welcome-to-the-nextgen-ocw-beta-site/
4. Look at the published version of another page you didn't change, for example https://ocw-draft-qa.global.ssl.fastly.net/pages/privacy-and-terms-of-use/

### Expected Behavior

- Each page should have the correct H1 in the blue banner. 

### Actual Behavior

- Both pages have the same H1, and the value seems unpredicable
- Note that each page has the correct HTML title 
- The titles appear to be correct in the markdown, for example: https://github.mit.edu/ocw-content-rc/ocw-www/blob/main/content/pages/privacy-and-terms-of-use.md

Wassaf-Shahzad:
#### Pre-Flight checklist
- [x] Testing

Closes #458 

#### What's this PR do?
Adds variant of page-title partial based on title of current page

This PR can be tested by the following way
- Start ocw-www
- Visit '/pages/privacy-and-terms-of-use/' and verify title is `Privacy and Terms of Use`
- Visit `/pages/welcome-to-the-nextgen-ocw-beta-site/` and verify title is `Welcome to the NextGen OCW beta site!`

### DIFF
--- a/www/layouts/pages/single.html
+++ b/www/layouts/pages/single.html
@@ -2,7 +2,7 @@
 <div class="page-single">
   {{ block "header" . }}
   {{ partialCached "header" . }}
-  {{ partialCached "page_title" . }}
+  {{ partial "page_title" . }}
   {{end}}
   <div class="container standard-width mx-auto mt-5">    
     {{ .Content }}


www/layouts/pages/single.html
```diff
@@ -2,7 +2,7 @@
 <div class="page-single">
   {{ block "header" . }}
   {{ partialCached "header" . }}
-  {{ partialCached "page_title" . }}
+  {{ partialCached "page_title" . .Title}}
```
alicewriteswrongs, line 5: I suspect we're not likely to render the `"page_title"` partial with the same `.Title` more than once - maybe it would be better to just remove `partialCached` entirely?

alicewriteswrongs (changes requested):

www/layouts/pages/single.html
Wassaf-Shahzad: Yeah i dont think the it improves the performance that much and it would also probably costs more in memory then it saves on performance.

gumaerc: @Wassaf-Shahzad in the very beginning of this project, we were trying to build all course sites as a single Hugo build so using `partialCached` was much more important.  Nowadays each course build is run individually, so caching partials is less effective.  I think it's okay to remove it, especially where it's causing issues like this.

alicewriteswrongs: my guess is that for most of our partials the caching is going to give us only very, very marginal build time increases given the overhead of all the stuff that's happening other than the Hugo build itself

@gumaerc this makes me think, back when we were doing the full site build we ended up using `partialCached` to render the course nav which, iirc, prevented us from setting class names for highlighting current nav items in HTML, so we had to lean on some JS instead to do that once the page had loaded in the user's browser. Should we revisit that perhaps? It is nice to have a pure-HTML nav-highlighting thing because then users don't need to use JS in order for the nav to work for them

alicewriteswrongs (approved):
lgtm!

www/layouts/pages/single.html
gumaerc: @alicewriteswrongs Yea, that sounds about right.  Nowadays the Hugo build's performance is the least of our concerns.

Issue #472, opened 2022-02-23
Inconsistent font size


pdpinch:
### Steps to Reproduce

1. See the section "Exercises and Homework" on https://ocwnext.odl.mit.edu/courses/18-s997-introduction-to-matlab-programming-fall-2011/pages/syllabus/
2. The font size in the paragraph text ("Throughout the course...") is larger than the font size in the bulleted list ("Exercises are intended to...")

1. Also, see the section "Reading" on https://ocwnext.odl.mit.edu/courses/3-091sc-introduction-to-solid-state-chemistry-fall-2010/pages/bonding-and-molecules/9-drawing-lewis-structures/
2. The font size in the paragraph text "Section 3" is larger than the font size in the table cell, "Ionization energies; electron affinities;"

### Expected Behavior

All this text should be the same size.

PR #309, opened 2021-12-30
feat: Educator Page

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
https://github.com/mitodl/ocw-hugo-themes/issues/307

#### What's this PR do?
Educator's page created which can be accessed from homepage, by clicking "Educators start here" 

#### How should this be manually tested?

- This PR is Deployed [here](https://ocw-hugo-themes-pr-309--ocw-next.netlify.app/educator/)
- Clicking "Educators Start Here" in homepage should redirect to this educator page.
- One long, scrolling page
- on-page nav should stay sticky, at top of viewport when user scrolls down 
- Clicking on item in on-page nav should smooth-scroll down to the appropriate section, not jump.
- Verify that all links work fine
- Should be a part of ocw-www theme
- Compare with design: https://invis.io/7Y12598HHF6E#/462148864_Educators_Start_Here
- Proofread content: https://docs.google.com/document/d/1DTD0cHXB_n3UH1HZwyD8adFOfrgxfY_zSzY5pTTyKyE
- As much as possible, styles should follow existing styles (e.g. fonts & colors), especially those added in About Us Page #303
- Test the responsiveness of the page for various screen sizes, including:
    - 1920×1080
    - 1366×768 
    - 768×1024 
    - 414×816 
    - 375×812
- Verify that the embedded video play
- Test the page on multiple browsers"
- Any other instructions, mentioned in relevant ticket

#### Screenshots (if appropriate)
Laptop screen: 

educator page SS

Mobile screen:

mobile SS

PR #474, opened 2022-02-24
fix: cleaning font-sizes, replacing px with rem

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/472

#### What's this PR do?
- Cleans up almost all font sizes.
- Root `font size` set to `16px`
- Font sizes variables created (`font-lg`, `font-md`) with `REM` values.
- Above created variables are used/re-used wherever required.
- Few font-sizes with `16px` value have been removed as they will inherit root's font size which is already `16px`.
- Few other tweaks in CSS to maintain font-sizes uniformity across the app (though there are still places which can be improved too)

- This might not sound good for the tester :) but all the **sections** on all the **pages** of **www** and **courses** need to be tested and verified that `font-size` content is clean, smooth, and consistent. 
- Verify that all the content of **www** and **courses** is looking fine and easy to read with respect to consistent `font-size`.
- [Preview netlify app here](https://ocw-hugo-themes-pr-474--ocw-next.netlify.app/) which can be compared with [ocwnext RC](https://ocwnext-rc.odl.mit.edu/)

### DIFF
--- a/base-theme/assets/css/main.scss
+++ b/base-theme/assets/css/main.scss
@@ -69,6 +69,18 @@ a {
   }
 }
 
+p {
+  font-size: $normal-text-font-size;
+}
+
+ul li {
+  font-size: $normal-text-font-size;
+}
+
+table {
+  font-size: $normal-text-font-size;
+}
+
 h1 {
   font-size: 36px;
   color: $black;
@@ -116,10 +128,6 @@ a + h3 {
   h3 {
     font-size: 18px;
   }
-}
-
-p {
-  font-size: 18px;
 }
 
 @include media-breakpoint-down(xs) {
@@ -370,7 +378,7 @@ article.content {
 }
 
 .course-title-section {
-  font-size: 18px;
+  font-size: $normal-text-font-size;
 }
 
 .course-section-title-container {
--- a/base-theme/assets/css/main.scss
+++ b/base-theme/assets/css/main.scss
@@ -45,7 +45,7 @@ a.coming-soon {
 
 body {
   font-family: "Helvetica";
-  font-size: 16px;
+  font-size: $font-normal;
 }
 
 .underline {
@@ -69,33 +69,21 @@ a {
   }
 }
 
-p {
-  font-size: $normal-text-font-size;
-}
-
-li {
-  font-size: $normal-text-font-size;
-}
-
-table {
-  font-size: $normal-text-font-size;
-}
-
 h1 {
-  font-size: 36px;
+  font-size: $heading-one-font-lg;
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 32px;
+    font-size: $heading-one-font-xs;
   }
 }
 
 h2 {
-  font-size: 28px;
+  font-size: $heading-two-font-lg;
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 24px;
+    font-size: $heading-two-font-xs;
   }
 
   &.branded {
@@ -109,14 +97,8 @@ h2 {
   }
 }
 
-@include media-breakpoint-down(xs) {
-  h2 {
-    font-size: 24px;
-  }
-}
-
 h3 {
-  font-size: 24px;
+  font-size: $heading-three-font-lg;
   color: $black;
 }
 
@@ -126,13 +108,7 @@ a + h3 {
 
 @include media-breakpoint-down(xs) {
   h3 {
-    font-size: 18px;
-  }
-}
-
-@include media-breakpoint-down(xs) {
-  p {
-    font-size: 16px;
+    font-size: $heading-three-font-xs;
   }
 }
 
@@ -297,6 +273,24 @@ table {
 .material-icons {
   @include icon-styles;
   font-family: "Material Icons";
+  font-weight: normal;
+  font-style: normal;
+  font-size: $material-icon-font-normal;
+  display: inline-block;
+  line-height: 1;
+  text-transform: none;
+  letter-spacing: normal;
+  word-wrap: normal;
+  white-space: nowrap;
+  direction: ltr;
+
+  /* Support for all WebKit browsers. */
+  -webkit-font-smoothing: antialiased;
+  /* Support for Safari and Chrome. */
+  text-rendering: optimizeLegibility;
+
+  /* Support for Firefox. */
+  -moz-osx-font-smoothing: grayscale;
 }
 
 .material-icons-round {
@@ -368,17 +362,9 @@ article.content {
 
       td {
         border: none;
-
-        p {
-          font-size: 1rem;
-        }
       }
     }
   }
-}
-
-.course-title-section {
-  font-size: $normal-text-font-size;
 }
 
 .course-section-title-container {
--- a/base-theme/assets/css/main.scss
+++ b/base-theme/assets/css/main.scss
@@ -45,7 +45,7 @@ a.coming-soon {
 
 body {
   font-family: "Helvetica";
-  font-size: $font-normal;
+  font-size: $font-root;
 }
 
 .underline {
@@ -328,7 +328,7 @@ table {
 }
 
 article.content {
-  font-size: 1rem;
+  font-size: $font-normal;
 
   h2 {
     color: $highlight-red;
--- a/base-theme/assets/css/main.scss
+++ b/base-theme/assets/css/main.scss
@@ -73,7 +73,7 @@ p {
   font-size: $normal-text-font-size;
 }
 
-ul li {
+li {
   font-size: $normal-text-font-size;
 }
 
--- a/base-theme/assets/css/variables.scss
+++ b/base-theme/assets/css/variables.scss
@@ -17,6 +17,7 @@ $medium-light-gray: #e9e9e9;
 $medium-gray: #9c9d9e;
 $dark-gray: #3d3d3d;
 $partial-collapse-height: 6em;
+$normal-text-font-size: 18px;
 
 // full-color design
 $highlight-red: #a31f34;
--- a/base-theme/assets/css/variables.scss
+++ b/base-theme/assets/css/variables.scss
@@ -17,7 +17,6 @@ $medium-light-gray: #e9e9e9;
 $medium-gray: #9c9d9e;
 $dark-gray: #3d3d3d;
 $partial-collapse-height: 6em;
-$normal-text-font-size: 18px;
 
 // full-color design
 $highlight-red: #a31f34;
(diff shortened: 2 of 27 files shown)


base-theme/assets/css/main.scss
```diff
@@ -41,6 +41,18 @@ a {
...
+}
+
+li {
+  font-size: $normal-text-font-size;
+}
+
+table {
+  font-size: $normal-text-font-size;
+}
+
```
MAbdurrehman12, line 55: Note to reviewer:
If there are other any other elements too which have the body font-size (i.e: `16px`), they can also be listed here.
 
Ideal case/solution  is to give this font-size to `body` element so it is reflected everywhere but it has already been given a font-size of `16px`, not sure why, and `p` element has been given a font-size of `18px`, again not sure why we have this difference at the first place so I have followed/extended `p` font-size (i.e: `18px`).

pdpinch:
Will this fix https://ocwnext.odl.mit.edu/courses/8-03sc-physics-iii-vibrations-and-waves-fall-2016/pages/part-i-mechanical-vibrations-and-waves/lecture-1/ ?

base-theme/assets/css/variables.scss
```diff
@@ -17,6 +17,7 @@ $medium-light-gray: #e9e9e9;
 $medium-gray: #9c9d9e;
 $dark-gray: #3d3d3d;
 $partial-collapse-height: 6em;
+$normal-text-font-size: 18px;
```
gumaerc, line 20: Can we change this to be `1rem;`?

gumaerc (changes requested):
@MAbdurrehman12 As you've probably noticed, the CSS in this project is a bit disorganized.  Some of the rules in here are from the initial designs which are pretty different from where we've ended up.  There have been some occurrences where we've been working off designs in Invision that have weird font sizes set, and a number of custom font sizes have been set directly in `px` units.

I'd like @pdpinch to confirm this, but I think the general desire here is to have a uniform base font size throughout the site.  The best way to do this in general IMO is to set the `font-size` attribute on the `body` element (it's currently set to the usual default of `16px`) and then use `rem` units everywhere, expressed as a percentage of the base font size.  This way, if you want to change your base font size it's as easy as changing it in one place, then everywhere that uses `rem` will scale accordingly.

I don't think the goal here is for the font to be standardized on `18px` for lists and tables, rather that it's to have one base `font-size` that everything is scaled from.  I think the main source of this has been people working off of Invision designs and taking the font size in the inspector literally, applying the pixel value directly to elements.

gumaerc:
I'm also interested in @alicewriteswrongs 's thoughts on this

MAbdurrehman12:
@gumaerc Yes, you're right! Ideal case is to apply standard font-size to `body` so our font-sizes are uniform in various elements and we will not have to do this which I have done in this current PR. But as stated [here](https://github.com/mitodl/ocw-hugo-themes/pull/474#discussion_r813813249), I was not sure why we have a separate font-size for `p` element as it is being used quite often. Even apart from this, there are many places where font-sizes have been separately defined in `px`, somewhere 18px, somewhere 16px etc. 

Anyhow to simplify this, my vote is for your suggestion, that we decide a general/normal font-size for the normal text of whole site and apply it to `body` and remove the font-sizes for `p`, `li`, `table` elements for now.

MAbdurrehman12:
> Will this fix https://ocwnext.odl.mit.edu/courses/8-03sc-physics-iii-vibrations-and-waves-fall-2016/pages/part-i-mechanical-vibrations-and-waves/lecture-1/ ?

Yes

base-theme/assets/css/variables.scss
MAbdurrehman12: Sure. 
1rem=16px, right? 
Though 16px is the default size, not sure but I doubt it might look a bit smaller in our course content.
@pdpinch What do you suggest?

alicewriteswrongs:
@gumaerc I think what you suggested makes sense to me - it would be great to clean up some of the messy CSS over here!

base-theme/assets/css/variables.scss
gumaerc: `1rem` equals whatever the `font-size` property is on the `body` element.  So, if that's `16px`, then it's `16px`

MAbdurrehman12: Okay, if the font-size is equal to that of `body` element then do we need to separately give the same font-size to these elements?
I think, what we can do is to decide a `font-size` for the `body` element, (either 16px or 18px) and remove these explicit font-sizes
image
So that all of these elements and other elements can be of the same font-size (i.e: font-size of body).
For `h1`, `h2`, `h3` and few other elements, we can replace their font-sizes with `rem`. 

What do you think?

pdpinch: yes please. simpler is better.

pdpinch:
> I'd like @pdpinch to confirm this, but I think the general desire here is to have a uniform base font size throughout the site. 

Yes, that is the goal. The folks doing QA testing have been reacting to places where the sizing is inconsistent.

base-theme/assets/css/variables.scss
MAbdurrehman12: Okay, what should be the font-size of `body` then, 16px or 18px? Currently it's `16px`.
Note: This font-size will also be applied to `p`, `li`, `table` and other elements unless explicitly defined.

pdpinch:
@MAbdurrehman12 have you tried using the git blame feature to see why 18px was used in some places and 16px in other places?

MAbdurrehman12:
> @MAbdurrehman12 have you tried using the git blame feature to see why 18px was used in some places and 16px in other places?

Sorry, I got engaged in some other stuff, I'll look into it now

MAbdurrehman12:
> have you tried using the git blame feature to see why 18px was used in some places and 16px in other places?

@pdpinch
So `body` element was given a font-size of `16px` in the first [PR](https://github.com/mitodl/ocw-hugo-themes/pull/1) which looks fine. 

`p` element was given a font-size of `18px` in this [PR](https://github.com/mitodl/ocw-hugo-themes/pull/309) which I shouldn't have given it in such an open way but I think I followed the design at that time and now this is causing the inconsistent font-sizes. `p` elements have `18px` and all the other elements, which don't have their own explicit font-size, inherits the font-size of `body` element which is `16px`.

Solution: I think we should remove the explicit font-size of `p` element so all the elements have consistent font-sizes (i.e: `16px` but this needs to be confirmed) OR give it `1 rem` as @gumaerc suggested [here](https://github.com/mitodl/ocw-hugo-themes/pull/474#pullrequestreview-892821038) and [here](https://github.com/mitodl/ocw-hugo-themes/pull/474#discussion_r814148599), but if in future, this `1 rem` is changed (increased/decreased) then we will again have inconsistency.

pdpinch:
> I think we should remove the explicit font-size of p element so all the elements have consistent font-sizes (i.e: 16px but this needs to be confirmed)

I agree. Can you include screenshots of the educator page on the PR so I can see what the impact is?

> give it `1 rem` as @gumaerc suggested [here](https://github.com/mitodl/ocw-hugo-themes/pull/474#pullrequestreview-892821038) and [here](https://github.com/mitodl/ocw-hugo-themes/pull/474#discussion_r814148599), but if in future, this `1 rem` is changed (increased/decreased) then we will again have inconsistency.

I don't have an opinion about using `px` or `rem` units. If that's not necessary for this issue, I think it's fine to leave as is, and possibly address later.

MAbdurrehman12:
> Can you include screenshots of the educator page on the PR so I can see what the impact is?

I have added screenshots from the educator page in the PR description. Do you also need a full web page capture?

gumaerc:
@pdpinch @MAbdurrehman12 It seems my understanding of this was also not entire accurate, but the reason I'm suggesting we use `rem` mostly has to do with responsive scaling.  It's not just a concern of responsiveness though, it's also an accessibility issue.  More on that [here](https://uxdesign.cc/why-designers-should-move-from-px-to-rem-and-how-to-do-that-in-figma-c0ea23e07a15).  As far as why we have explicit font sizes set throughout the CSS, I believe that has to do with taking the font sizes set in InVision designs literally.  Some of it comes from very early on in the project when we were implementing the initial designs.

Anyway, from my research, we shouldn't even be setting an explicit root font size and should be using `rem` wherever we define font size, unless we want to set font size relative to the given *element* a font size is set on (not the root element), in which case we would use `em`.  If you explicitly set font sizes in pixels (an absolute unit) then you can prevent users who set a font size scale in their browser from seeing the scaled font size.  From what I've read this is important for vision impaired users, and I wouldn't want to have this be a problem for users of OCW.

MAbdurrehman12:
@gumaerc I agree with that and I think we should proceed with it. So I guess we can do something like this: (please correct me if I'm wrong)
- Remove explicit font-size of `body` element as it will be 16px by default.
- Remove all the other explicit font-sizes which are set to `16px` as, by default, they will inherit root's font-size which will already be `16px` OR replace with `1 rem` (if required)
- Replace all the existing `px` font-sizes with `rem`
- Replace all the existing `px` font-sizes with `em` which are relative to any particular element whose font size is set.

Let me know if we are on the same page or in case I missed or misunderstood anything.
If this is good, then I'll go on and implement these changes since there might be changes in many CSS files so I guess we'll have to spend some time testing and verifying if everything is good.
@pdpinch Can you also please share your thoughts on this and its priority?

pdpinch:
This is a relatively high priority, but it's also important that we do some testing and get it right. We'll have to live with this for some time. 

We should get input from @alicewriteswrongs as well.

alicewriteswrongs:
@MAbdurrehman12 I think what you laid out makes sense. It will be good to comb through our styles and convert everything over to `rem`.

MAbdurrehman12:
> It would be good to keep an eye out when doing so for where we can define a few reusable styles for special elements here  ....

This is a good suggestion, I'll try to keep common styles reusable

base-theme/assets/css/variables.scss
```diff
@@ -3,15 +3,15 @@ $max-content-width: 1447px;
 $desktop-header-height: 116px;
 $desktop-header-max-width: 1664px;
 $extra-small-screen-width: 400px;
-$enable-responsive-font-sizes: true;
-$font-size-base: 0.8rem;
-$headings-font-weight: 600;
-$table-cell-padding: 0.5rem;
-$offcanvas-width: 85vw;
-$transparent: #ffffff00;
+$enable-responsive-font-sizes: true; /* not found */
```
MAbdurrehman12, line 6: @gumaerc I was unable to find any usage of these variables and a few other classes pointed below, are these being/will be used somewhere or we can remove them?

alicewriteswrongs: I think we can safely delete any unused variables - if we need them later, they'll be in the git history 🙂

MAbdurrehman12: @alicewriteswrongs If you get time, can you also review this PR?

MAbdurrehman12:
> but I did notice some differences that are maybe worth taking a look at in case we want to make some changes

Thanks for reviewing the PR and pointing this out @gumaerc. I think this is due to the removal of some variables that I thought are unused but it looks like they are being used somehow, I'm not sure how, maybe they are being used in course content as I couldn't find their usage in `hugo-themes` as stated [here](https://github.com/mitodl/ocw-hugo-themes/pull/474#discussion_r822343897).

Anyhow, I have recovered them except this one: `$font-size-base: 0.8rem;` and now the heading weight looks fine as shown below:

RC:
image

This Branch:
image

> I think we should merge this but still have a design review in RC

I agree, let's merge this and do further testing on RC.

PR #487, opened 2022-02-28
Added course collection filter based on Title

### LINKED ISSUE #486: Home: hard code course collections
In #411 we added course collections to the home page. We later decided that instead of having _all_ the course collections on the home page, we'd rather link to pages that aggregate several, smaller course collections. 

We will probably iterate on this further, but this should be good enough for the MVP launch in March. 

#### Acceptance Criteria:

- [ ] There should be just 7 course collections on the home page: 

 n | title | partial URL 
-- | -- | -- 
1 | Open Learning Library | /pages/open-learning-library/
2 | Introductory Programming | /pages/introductory-programming/
 3 | Energy | /pages/energy/
 4 | Entrepreneurship | /pages/entrepreneurship/
 5 | Environment | /pages/environment/
 6 | Life Sciences | /pages/life-sciences/ 
 7 | Transportation | /pages/transportation/

- [ ] if it's possible to make the links conditional on the presence of the pages, I think that will avoid future problems later.

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

Closes #486 

#### What's this PR do?
Adds relevant filter in course collection partial

Verify the following
- All course collection available in ocw-www content are filtered based on `homePageCourseCollections`  param
- If a collection exists in ocw-www content but the title is not in `homePageCourseCollections` it is not rendered

Screenshot 2022-02-28 at 5 00 15 PM

### DIFF
--- a/www/content/_index.md
+++ b/www/content/_index.md
@@ -1,10 +1,10 @@
 ---
 renderSearchIcon: false
 homePageCourseCollectionsUrls:
-- /pages/introductory-programming 
-- /pages/open-learning-library/
-- /pages/open-learning-library/
-- /pages/open-learning-library/
-- /pages/open-learning-library/
-- /pages/open-learning-library/
+- /pages/open-learning-library
+- /pages/introductory-programming
+- /pages/energy
+- /pages/entrepreneurship
+- /pages/environment
+- /pages/transportation
 ---
--- a/www/content/_index.md
+++ b/www/content/_index.md
@@ -1,11 +1,10 @@
 ---
 renderSearchIcon: false
-homePageCourseCollections:
-- Open Learning Library
-- Intro Programming
-- Energy
-- Entrepreneurship
-- Environment
-- Life Sciences
-- Transportation 
+homePageCourseCollectionsUrls:
+- /pages/introductory-programming 
+- /pages/open-learning-library/
+- /pages/open-learning-library/
+- /pages/open-learning-library/
+- /pages/open-learning-library/
+- /pages/open-learning-library/
 ---
--- a/www/content/_index.md
+++ b/www/content/_index.md
@@ -1,3 +1,11 @@
 ---
 renderSearchIcon: false
+homePageCourseCollections:
+- Open Learning Library
+- Intro Programming
+- Energy
+- Entrepreneurship
+- Environment
+- Life Sciences
+- Transportation 
 ---
--- a/www/layouts/partials/home_course_collections.html
+++ b/www/layouts/partials/home_course_collections.html
@@ -6,12 +6,14 @@
       collections. Below are some topics available for you to explore:
     </p>
     <div class="course-collection-section d-flex flex-row m-0 flex-wrap">
-      {{ range where .Site.Pages "Type" "course_collections" | intersect (where .Site.Pages "Title" "in" .Params.homePageCourseCollections)}}
-      <a href="{{.Permalink}}">
-        <div class="collection-item">
-          <p>{{.Title}}</p>
-        </div>
-      </a>
+      {{ range $url := .Params.homePageCourseCollectionsUrls}}
+        {{ with $.Site.GetPage $url }}
+          <a href="{{ .Permalink }}">
+            <div class="collection-item">
+              <p>{{ .Title }}</p>
+            </div>
+          </a>
+        {{ end }}
       {{ end }}
     </div>
   </div>
--- a/www/layouts/partials/home_course_collections.html
+++ b/www/layouts/partials/home_course_collections.html
@@ -6,7 +6,7 @@
       collections. Below are some topics available for you to explore:
     </p>
     <div class="course-collection-section d-flex flex-row m-0 flex-wrap">
-      {{ range where .Site.Pages "Type" "course_collections" }}
+      {{ range where .Site.Pages "Type" "course_collections" | intersect (where .Site.Pages "Title" "in" .Params.homePageCourseCollections)}}
       <a href="{{.Permalink}}">
         <div class="collection-item">
           <p>{{.Title}}</p>


pdpinch:
@gumaerc can I get your thoughts on this approach to "hard coding" these links on the home page?

mbertrand (commented):
Most of these course collections don't exist yet at https://github.mit.edu/mitocwcontent/ocw-www/tree/main/content/course_collections, so only "Open Learning" shows up for me.  But that seems to satisfy "make the links conditional on the presence of the pages" mentioned in the issue.  Should these collections be created before this is merged?

www/content/_index.md
```diff
@@ -1,3 +1,11 @@
 ---
 renderSearchIcon: false
+homePageCourseCollections:
+- Open Learning Library
+- Intro Programming
+- Energy
+- Entrepreneurship
+- Environment
+- Life Sciences
```
pdpinch, line 9: Remove "Life Science" 

We don't want to include it any more.

www/layouts/partials/home_course_collections.html
```diff
@@ -6,7 +6,7 @@ <h2>Discover Collections</h2>
       collections. Below are some topics available for you to explore:
     </p>
     <div class="course-collection-section d-flex flex-row m-0 flex-wrap">
-      {{ range where .Site.Pages "Type" "course_collections" }}
+      {{ range where .Site.Pages "Type" "course_collections" | intersect (where .Site.Pages "Title" "in" .Params.homePageCourseCollections)}}
```
pdpinch, line 9: These are page now, not course collections. Sorry for the confusing naming, but we want the home to link to pages, and the pages will link to course collections. 

On production, the relevant URLs are:




title | production URL (although partial URLs would be better) 
-- | --
Open Learning Library | https://ocwnext.odl.mit.edu/pages/open-learning-library/
Introductory Programming | https://ocwnext.odl.mit.edu/pages/introductory-programming/
Energy | https://ocwnext.odl.mit.edu/pages/energy/
Entrepreneurship | https://ocwnext.odl.mit.edu/pages/entrepreneurship/
Environment | https://ocwnext.odl.mit.edu/pages/environment/
Transportation | https://ocwnext.odl.mit.edu/pages/transportation/

pdpinch (changes requested):
Sorry, we had some miscommunication. These links go to _pages_ now.

Also, we want to remove one of them, Life Sciences.

gumaerc:
@pdpinch @Wassaf-Shahzad Overall this approach is fine and pretty straightforward, at least in the short term.  My only question here is if we really want to handle the linking with `title`.  On one hand it's nice that the array is human readable, and there's the benefit of (in theory) the array of names would be the same in all environments.

On the other hand though, that same benefit could come back to bite us.  If we rename any of the pages that this will link to, we will need to remember to also re-name the entry in this array.  In the rest of the site, we handle situations like this with a relation widget.  With all this being said we might want to go with a different approach, at least in the long term.

In the `ocw-studio` starter for course sites, we define the Metadata section using the `file` directive with a path of `data/course.json`, which tells `ocw-studio` to add a UI option for editing metadata fields that will produce the JSON file at the defined path.  What we might want to do is, in `ocw-hugo-projects`, add a section called "Home Page" with the `file` directive set to `content/_index.md` and add these metadata properties as fields.  For things like pages, we could use the relation widget to link pages to the collection via UUID, and then the lookup of these pages would be much more robust.

pdpinch:
I like this idea, but I don’t think we have time for it right now. Can we do that as a follow-up issue? 

> In the ocw-studio starter for course sites, we define the Metadata section using the file directive with a path of data/course.json, which tells ocw-studio to add a UI option for editing metadata fields that will produce the JSON file at the defined path. What we might want to do is, in ocw-hugo-projects, add a section called "Home Page" with the file directive set to content/_index.md and add these metadata properties as fields. For things like pages, we could use the relation widget to link pages to the collection via UUID, and then the lookup of these pages would be much more robust.

www/layouts/partials/home_course_collections.html
Wassaf-Shahzad: I updated  the partial to link to pages. Sorry for the delay, I misunderstood the issue

Wassaf-Shahzad:
> I like this idea, but I don’t think we have time for it right now. Can we do that as a follow-up issue?
> […](#)
> In the ocw-studio starter for course sites, we define the Metadata section using the file directive with a path of data/course.json, which tells ocw-studio to add a UI option for editing metadata fields that will produce the JSON file at the defined path. What we might want to do is, in ocw-hugo-projects, add a section called "Home Page" with the file directive set to content/_index.md and add these metadata properties as fields. For things like pages, we could use the relation widget to link pages to the collection via UUID, and then the lookup of these pages would be much more robust.



> @pdpinch @Wassaf-Shahzad Overall this approach is fine and pretty straightforward, at least in the short term. My only question here is if we really want to handle the linking with `title`. On one hand it's nice that the array is human readable, and there's the benefit of (in theory) the array of names would be the same in all environments.
> 
> On the other hand though, that same benefit could come back to bite us. If we rename any of the pages that this will link to, we will need to remember to also re-name the entry in this array. In the rest of the site, we handle situations like this with a relation widget. With all this being said we might want to go with a different approach, at least in the long term.
> 
> In the `ocw-studio` starter for course sites, we define the Metadata section using the `file` directive with a path of `data/course.json`, which tells `ocw-studio` to add a UI option for editing metadata fields that will produce the JSON file at the defined path. What we might want to do is, in `ocw-hugo-projects`, add a section called "Home Page" with the `file` directive set to `content/_index.md` and add these metadata properties as fields. For things like pages, we could use the relation widget to link pages to the collection via UUID, and then the lookup of these pages would be much more robust.

Wassaf-Shahzad:
@pdpinch @gumaerc 
I updated the PR with the relevant changes.However i have a few thoughts
- I dont think my current implementation is the best solution overall Although it will do for the short term. To me the best solution would be to add to extra fields in frontmatter of the course collection page  which would editable from the ocw-studio. `isCourseCollectionPage`, `renderOnHomePage`. Then we would be able to dynamically render them based on this values. New pages could be added and old pages could be archived as per requirements.
- Also currently I am using course collection pages url as the filtering key since there no property which could be used 
to distinguish between course collection pages and other pages (I could do a `collection in title`  kinda check but that would be even worse).

gumaerc (approved):
:+1: This approach looks fine for now.  In the future we should take either the approach @Wassaf-Shahzad suggested by putting a front matter property in the pages expected to show up here, or defining a section for editing the home page in the studio UI and using a multiple choice relation widget there.

PR #509, opened 2022-03-08
remove description from video-gallery partial

ChristopherChudzicki:
- [x] Testing

#### What's this PR do?
We use the markdown body for descriptions, not the description field. This is safe to remove:

- there were 9 video galleries with descriptions on RC, all in test courses... I moved them all markdown body via admin panel.
- there are 0 video galleries with descriptions on prod (of 218 video galleries)

Open a video gallery and check that it still works.

For example, http://localhost:3000/video_galleries/derivatives/ in `res.18-005-spring-2010`. But any video gallery would do.


mbertrand (approved):
👍

Issue #523, opened 2022-03-15
support links in resource short codes


pdpinch:
On legacy OCW, there are >1000 images with links on them. 

For example, from [CMS6.11J F14](https://github.mit.edu/mitocwcontent/cms.611j-fall-2014/blame/main/content/pages/projects/cholera-control.md#L16-L17)

```
[{{< resource 69e1f411-a309-0f84-2b21-fa36db9f1d60 >}}  
](/ans7870/CMS/CMS.611/f14/games/cholera_control_final/index.html)
```

As currently constructed, we have no way to render these. To address this, we would like to add two optional parameters to the resource shortcode:

- [ ] `href` where the value is set to a full or partial URL. When set, the resource (typically an image) would be wrapped in an anchor tag with an `href` attribute, with the value set to the passed value. 
- [ ] `href_uuid` where the value to set to the uuid of another resource (typically a page, or a resource). When set, the resource would be wrapped in an anchor tag, with an `href` attribute, with the value set to a URL for the resource. 
- [ ] having both parameters set is not supported. It's OK to use the `href` value in this case. 

### Related issues

- https://github.com/mitodl/ocw-studio/issues/1130

PR #538, opened 2022-03-21
feat: support links in resource (image) short codes

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/523

#### What's this PR do?
- Since we have to pass `href` or `href_uuid` to resource shortcode so now our resource shortcodes will have `named` parameters unlike before (having only 1 positional parameter i.e: UUID).
- So this PR checks and get named parameters which are/can be:
    - UUID (required)
    - href (optional)
    - href_uuid (optional) 
But since all the previous resource shortcodes have 1 positional parameter (UUID) so its support is also there hence if UUID is not found as a named parameter then it is assumed that it will be a positional parameter and it is fetched as a positional parameter.
- Default values (empty) values are set for named parameters.
- Images resource partial is created which does the following:
    - For all the image resources, a check is placed for parameters: `href` and `href_uuid`, if any one of them is found, then the image resource is wrapped up with an anchor tag, redirecting to the href value
    - `href` where the value is set to a full or partial URL. When set, the resource (typically an image) would be wrapped in an anchor tag with an `href` attribute, with the value set to the passed value. 
    - `href_uuid` where the value to set to the uuid of another resource (typically a page, or a resource). When set, the resource would be wrapped in an anchor tag, with an `href` attribute, with the value set to a URL for the resource. 
    - having both parameters set is not supported. It's OK to use the `href` value in this case. 

- Go to any course which has image resources and run it locally. [example](https://github.mit.edu/ocw-content-rc/1.252j-fall-2016/blob/main/content/resources/south_end-compressor.md)
-Pass a parameter to any image resource: `href` and set its value to any URL: (e.g: `{{< resource  uuid="4b2d6fb6-96a3-3b08-c576-8653fb9e1a07" href="https://google.com" >}}`)
     - Verify that now image is clickable and redirects to this URL.
- Replace this `href` with `href_uuid` and set it's value to UID of any resource: (e.g: `{{< resource  uuid="4b2d6fb6-96a3-3b08-c576-8653fb9e1a07" href_uuid="**uuid of any other resource**" >}}`)
     - Verify that now image is clickable and redirects to this resource (whose UID has been set)
- Add both parameters, `href` and `href_uuid` with their respective values: (e.g: `{{< resource  uuid="4b2d6fb6-96a3-3b08-c576-8653fb9e1a07" href="https://google.com" href_uuid="**uuid of any other resource**" >}}`)
     - Verify that image is clickable and redirects to `href` value
- Remove both parameters:
     - Verify that now image is not clickable
- Verify old resources which have 1 positional parameter (e.g: `{{< resource  4b2d6fb6-96a3-3b08-c576-8653fb9e1a07 >}}`) and verify that it is also working smoothly.
- Any other use-case which I have missed.

### DIFF
--- a/course/layouts/shortcodes/resource.html
+++ b/course/layouts/shortcodes/resource.html
@@ -1,9 +1,28 @@
 {{- $uuid := index .Params 0 -}}
+{{- $imageHref := "" -}}
 {{- range $.Site.Pages -}}
   {{- if eq (replace .Params.uid "-" "") (replace $uuid "-" "") -}}
     {{- if eq .Params.resourcetype "Image" -}}
-      {{- $metadata := .Params.image_metadata | default dict -}}
-      <img src="{{ partial "resource_url" .Params.file }}" {{ if index $metadata "image-alt" -}}alt="{{ index $metadata "image-alt" }}"{{- end }} />
+
+      {{- if .Params.href_uuid -}}
+        {{- range where $.Site.Pages "Params.uid" .Params.href_uuid -}}
+          {{ $imageHref = .Permalink }}
+        {{- end -}}
+      {{- end -}}
+
+      {{- if .Params.href -}}
+        {{$imageHref = .Params.href }}
+      {{- end -}}
+
+      {{- if $imageHref -}}
+        <a href="{{ $imageHref }}" target="_blank">
+      {{- end -}}
+        {{- $metadata := .Params.image_metadata | default dict -}}
+        <img src="{{ partial "resource_url" .Params.file }}" {{ if index $metadata "image-alt" -}}alt="{{ index $metadata "image-alt" }}"{{- end }} />
+      {{- if $imageHref -}}
+        </a>
+      {{- end -}}
+
     {{- else if eq .Params.resourcetype "Video" -}}
       {{ partial "video_embed.html" . }}
     {{- else -}}
--- a/course/layouts/shortcodes/resource.html
+++ b/course/layouts/shortcodes/resource.html
@@ -1,9 +1,16 @@
-{{- $uuid := index .Params 0 -}}
+{{- $uuid := default "" (.Get "uuid") -}}
+{{- $href := default "" (.Get "href") -}}
+{{- $href_uuid := default "" (.Get "href_uuid") -}}
+
+{{- if not $uuid -}}
+  {{- $uuid = index .Params 0 -}}
+{{- end -}}
+
 
 {{- range $.Site.Pages -}}
   {{- if eq (replace .Params.uid "-" "") (replace $uuid "-" "") -}}
     {{- if eq .Params.resourcetype "Image" -}}
-      {{ partial "image_resource.html" . }}
+      {{ partial "image_resource.html" (dict "context" . "href" $href "href_uuid" $href_uuid) }}
     {{- else if eq .Params.resourcetype "Video" -}}
       {{ partial "video_embed.html" . }}
     {{- else -}}
--- a/course/layouts/shortcodes/resource.html
+++ b/course/layouts/shortcodes/resource.html
@@ -1,28 +1,9 @@
 {{- $uuid := index .Params 0 -}}
-{{- $imageHref := "" -}}
+
 {{- range $.Site.Pages -}}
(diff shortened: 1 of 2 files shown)


gumaerc (changes requested):
I think there may have been some miscommunication about how this is supposed to be implemented.  It seems here that you have implemented the ability to set `href` or `href_uuid` properties in the front matter of a resource, but I think what's desired here is to be able to embed a resource with an arbitrary link wrapping it in the content like so:

`{{< resource uuid="cc7d2383-1fd1-3c1d-286d-15ac64cdba9b" href="https://www.google.com" >}}`

Setting the `href` or `href_uuid` properties on the resource itself means that only one link can ever be associated with that resource.  If you wanted multiple embedded resources to link to different places on different pages, you couldn't accomplish it with the functionality implemented like it is here. @pdpinch do I have this right?

pdpinch:
Carey’s right. I was looking to have a parameter on the resource shortcode, not the resource front matter. The use case is a thumbnail image of a person and when you click on it it take you to their bio page.

MAbdurrehman12:
My bad!
Thanks for pointing this out @gumaerc

I'm updating this PR.

MAbdurrehman12:
@gumaerc @pdpinch In order to achieve this, I think, we will have to switch to named parameters for resource shortcodes instead of positional parameters. 
But since currently all the resource shortcodes have positional parameter (e.g: `{{< resource 4b2d6fb6-96a3-3b08-c576-8653fb9e1a07 >}}`) so it's support is also there atm. 
Hence even if we pass a positional parameter, named parameters including/excluding `href` or `href_uuid`, it should work

course/layouts/shortcodes/resource.html
```diff
@@ -1,9 +1,16 @@
-{{- $uuid := index .Params 0 -}}
+{{- $uuid := default "" (.Get "uuid") -}}
+{{- $href := default "" (.Get "href") -}}
+{{- $href_uuid := default "" (.Get "href_uuid") -}}
+
```
MAbdurrehman12, line 4: I think it will be better if we get this href and href_uuid in image_resource partial instead of getting them here and passing them as a parameter to image_resource partial as `resource.html` as no usage of them, but I was having an issue while doing that.
@gumaerc Do you have any thoughts on this?

gumaerc: @MAbdurrehman12 one of the issues is that `ocw-studio` isn't set up to embed an "`image_resource`" shortcode on its own.  The shortcode would need to be created here in `ocw-hugo-themes`, and then `ocw-studio` would need to be set up to utilize it.  You do have a good point though in that I can't think of another type of embeddable resource that would be wrapped in a link.  What do you think @pdpinch ?

gumaerc: FYI, the downside of accepting these parameters on the `resource` shortcode has to do with the fact that we currently only have one positional parameter on it, the UUID.  If we decide we want to start passing in other things like `href` then we will need to also have `uuid` be a named agument in all existing `resource` shortcodes, so we will have to go through and replace that in the existing data.

MAbdurrehman12: but do let me know if you find out any other issue as we have quite a lot of resources so its better to play safe

gumaerc (approved):
:+1: LGTM

The conditional surrounding getting the positional parameter if the named parameter isn't present seems sensible to me. I tested every situation this supports manually and also ran through building all courses and didn't encounter any errors, so I'd say we're good to merge this.

gumaerc:
@MAbdurrehman12 before merging this can you update the description to clarify on the way this works now, so if people are referring to this PR in the future they'll be less confused?

MAbdurrehman12:
> I tested every situation this supports manually and also ran through building all courses

Thanks for this thorough testing @gumaerc 

> before merging this can you update the description to clarify on the way this works now

Yes, sure

MAbdurrehman12:
@gumaerc I have updated the description of this PR. 
Let me know if there's any issue or if I have missed anything.

Thanks for your support :)

PR #540, opened 2022-03-21
add course collection display

alicewriteswrongs:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

no ticket over here

#### What's this PR do?

This does basically two things:

- rename our existing templates and code to reflect renaming 'course collections' to 'course lists'
- add rendering of the new, list-of-lists which is what we're now calling a 'course collection'

You'll need to get ahold of some course list and course collection content. The easiest way right now I think is to check out https://github.com/mitodl/ocw-studio/pull/1146 and then get the content out of that. If you have a github organization set up on MIT github you can then clone your `ocw-www` repo locally and set `WWW_CONTENT_PATH` to be the path to that repo. Then you could be able to check out rendered course lists (at `/course-lists/$SLUG`) and collections (at `/collections/$SLUG`).

Screen Shot 2022-03-21 at 2 33 14 PM


ChristopherChudzicki (approved):
Looks good and is working for me locally.

I know this is a priority, so do we want to do Ferdi's UI tweaks as separate work? If so, we should make an issue for it.

Or if you want to do them here, let me know and I can re-review.

Issue #548, opened 2022-03-23
remove remaining @ts-nocheck comments


alicewriteswrongs:
when we converted over to building w/ typescript typechecking of all files was initially disabled using `// @ts-nocheck`. A lot of files have since been converted over fully to typescript, but there are some remaining to be done.

We should:

- remove them from all react-related files
- remove them from all non-react JS files (for light interactivity on course templates, for instance)

PR #549, opened 2022-03-23
remove @ts-nocheck on two test files

alicewriteswrongs:

part of #548

#### What's this PR do?

Just remove `@ts-nocheck` and fix all the type issues on two files.

Just tests, so they should pass and whatnot.


Wassaf-Shahzad (approved):
LGTM!

Issue #536, opened 2022-03-18
download link on embedded video fails


pdpinch:
in PR https://github.com/mitodl/ocw-hugo-themes/pull/445 we fixed the download button on the single video page. It looks like we need to do the same thing on the embedded video theme as well. 

See, for example, https://ocwnext.odl.mit.edu/courses/5-95j-teaching-college-level-science-and-engineering-fall-2015/pages/class-session-videos/

PR #556, opened 2022-03-24
fix: embedded video download

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/536

#### What's this PR do?
- Fixes the download of embedded video

- Checkout to this branch
- Build any courses locally which has embedded video(s)
- Try downloading the video(s) and verify that videos are downloaded/played successfully.

### DIFF
--- a/course/layouts/partials/video_embed.html
+++ b/course/layouts/partials/video_embed.html
@@ -17,7 +17,7 @@
     </div>
    	{{ if  $downloadLink }}
 	  <div class="video-download-container">
-	    <a class="download-file video-download-button" href="{{ partial "resource_url" $downloadLink }}"><span class="material-icons">file_download</span> Download Video</a>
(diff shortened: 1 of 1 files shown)


Wassaf-Shahzad (approved):
LGTM! 
however the same code exists in video.html as well. I think we should aggregate them into a separate partial to avoid  code duplication

MAbdurrehman12:
> however the same code exists in video.html as well. I think we should aggregate them into a separate partial to avoid code duplication

Yup, maybe a better approach can be to make a generic partial for download button which can be reused for all the download buttons (transcript, video, or any other download button), but I think this will be a technical debt and we can get @pdpinch's thoughts.

For now, I think, we can merge this PR to fix the bug and come to the improvement later on.

Issue #520, opened 2022-03-14
specify video start time in the URL 


pdpinch:
As a user, I'd like to be able to jump ahead in a long video to the most relevant segment. 

As a course author, I'd like to be able to specify the start time of a long video. 

#### Designs and Mockups

In legacy OCW, there any many video links that use javascript to specify the start time for the video. For example, the links the rightmost column in the table at the bottom of [this page in 8.03SC](https://ocw.mit.edu/courses/physics/8-03sc-physics-iii-vibrations-and-waves-fall-2016/part-i-mechanical-vibrations-and-waves/lecture-1/)

It would be best to pass the start time as a query parameter in the URL, as is done on YouTube, or in [this PR in OVS](https://github.com/mitodl/odl-video-service/pull/546). 

#### Acceptance Criteria:

- [ ] If the URL includes a query string with the name `start` and an integer value, advance the YouTube player to that start time

#### Related issues

- Are the video start times available in the parsed json? Is it in the legacy HTML?
- When importing courses with video start times in ocw-to-hugo, include the start time in the short code
- for embedded videos, add a `start` parameter to the video embed shortcode
- design a UI for setting start times in ocw-studio

dseaton:
More information gathering [here](https://docs.google.com/document/d/1hAbTZovVSZXTCiw7_sopUui9hnacZ7no9aKNffLS7lk/edit#).

PR #560, opened 2022-03-25
add start and end to videos

abeglova:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/ocw-hugo-themes/issues/520

#### What's this PR do?
This pr passes the start and end times to the video js player

Run ocw-to-hugo for `18-01sc-single-variable-calculus-fall-2010` with AWS_BUCKET_NAME=open-learning-course-data-rc in your .env file

In ocw-hugo-themes set COURSE_CONTENT_PATH=<path to your ocw-to-hugo output> and OCW_TEST_COURSE=18-01sc-single-variable-calculus-fall-2010

Run `npm run start:course` go to http://localhost:3000/resources/clip-3-limit-of-secants/. Verify that the video starts at 8:46

### DIFF
--- a/course/layouts/partials/video.html
+++ b/course/layouts/partials/video.html
@@ -5,17 +5,21 @@
 {{ $relatedResourses := .Params.related_resources_text | default "" }}
 {{ $optionalTabTitle := .Params.optional_tab_title | default "" }}
 {{ $optionalTabContent := .Params.optional_text | default ""}}
+{{ $startTime := .Params.start_time | default "" }}
+{{ $endTime := .Params.end_time | default "" }}
+
 {{ if (eq $downloadLink nil) }}
 {{ $downloadLink = .Params.video_files.archive_url }}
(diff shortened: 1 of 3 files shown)


abeglova:
The video-js/  youtube ui makes it a bit hard to keep playing from where the clip left off when the user reaches the end of the clip. I don't think there is an easy way to fix that. This is definitely good enough for launch

ChristopherChudzicki (approved):
> The video-js/ youtube ui makes it a bit hard to keep playing from where the clip left off when the user reaches the end of the clip. I don't think there is an easy way to fix that. This is definitely good enough for launch

@abeglova  yeah, i see what you mean... Should we make an issue for this to handle post launch? _maybe you did already_

abeglova:
I think this current change is just for legacy courses anyway, so I don't think we need a followup pr

PR #565, opened 2022-03-29
fixed styling issues for appzi feedback button

### LINKED ISSUE #564: Appzi Feedback tab overlaps content on mobile...
... and sometimes even on the desktop

### Screenshot or Screencast

examples:

1. 
Screen Shot 2022-03-25 at 11 29 24 (1)

2. 
image

### Expected Behavior

- [ ] on mobile, increase the right margin to ensure that the body text doesn't covered by the feedback tab
- [ ] on desktop, move the feedback tap down the page so it doesn't overlap with the course info drawer button (unless the user has scrolled the page) 
- [ ] the feedback tab should appear in the tab order so it can be keyboard accessibile.

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

Closes #564 

#### What's this PR do?
Fixes styling issues for Appzi feedback button

Verify the following
- `Hide course info` button does not overlap with `feedback` button
-  `feedback` button does not overlap with content
- `feedback` button possition on ocw-www and ocw-course  is consistant

Screenshot 2022-03-29 at 1 25 45 PM
Screenshot 2022-03-29 at 1 26 11 PM
Screenshot 2022-03-29 at 1 28 24 PM

### DIFF
--- a/course/layouts/partials/course_content.html
+++ b/course/layouts/partials/course_content.html
@@ -28,7 +28,7 @@
     </div>
   </header>
   {{ partial "mobile_nav_toggle.html" . }}
-  <article class="content pt-3 mt-1">
+  <article class="content pt-3 mt-1 mr-5">
     <main id="course-content-section">{{- .Content -}}</main>
   </article>
 </div>
--- a/course/layouts/partials/course_content.html
+++ b/course/layouts/partials/course_content.html
@@ -28,7 +28,7 @@
     </div>
   </header>
   {{ partial "mobile_nav_toggle.html" . }}
-  <article class="content pt-3 mt-1 mr-5">
+  <article class="content pt-3 mt-1 mr-5 mr-md-0">
     <main id="course-content-section">{{- .Content -}}</main>
   </article>
 </div>
--- a/base-theme/assets/css/main.scss
+++ b/base-theme/assets/css/main.scss
@@ -391,3 +391,8 @@ article.content {
 .aspect-ratio-16-9 {
   aspect-ratio: 16/9 !important;
 }
+
+// selector for appzi button
+div[data-appzi-dom] {
+  top: 30% !important;
+}


course/layouts/partials/course_content.html
```diff
@@ -28,7 +28,7 @@ <h1 class="text-uppercase font-weight-bold pb-1 mb-3">{{ .Title }}</h1>
     </div>
   </header>
   {{ partial "mobile_nav_toggle.html" . }}
-  <article class="content pt-3 mt-1">
+  <article class="content pt-3 mt-1 mr-5">
```
MAbdurrehman12, line 31: I think we should give `right margin` on mobile screens only as mentioned in the issue ticket. 
Maybe we can give equal margins on both sides (left and right) (if it looks good) so there is no inconsistency in left/right margins on mobile screens

MAbdurrehman12 (changes requested):
Looks fine to me, just a small thing mentioned.
And you can test is as well but I'm unable to access feedback tab via my keyboard tab button.

pdpinch:
Screenshots look good to me. 

I don't have an opinion on @MAbdurrehman12's comment -- I can see an argument for consistency across all screen widths, but I also see an argument for keeping the change just to mobile.  

In any case, can we merge this today?

course/layouts/partials/course_content.html
pdpinch: Since this change is in a partial, does it mean that courses will need to be republished to pick up the change?

Wassaf-Shahzad: No, since this is a change in ocw-hugo-themes styling

Wassaf-Shahzad:
> Screenshots look good to me.
> 
> I don't have an opinion on @MAbdurrehman12's comment -- I can see an argument for consistency across all screen widths, but I also see an argument for keeping the change just to mobile.
> 
> In any case, can we merge this today?

@pdpinch @MAbdurrehman12 
One thing is that we cant add the feedback button to tab-order since it actually isn't a button but a span and is inserted dynamically from appzi. We might me able to add it through appzi's dashboard but even that is long shot.
We can use the `nav-link` css property but that isn't  recommended.

And yes this can be merged today.

PR #610, opened 2022-04-05
Added check to fix hugo build failure

Wassaf-Shahzad:
#### Pre-Flight checklist

- [c] Testing
  - [x] Code is tested

No Ticket. Hot fix

#### What's this PR do?
added a safe guard in course_card.html from nil values in course.level

- Build should pass
- Manual testing can be done by passing nil to course_card.html in new_course_card.html partial

### DIFF
--- a/www/layouts/partials/course_cards_level.html
+++ b/www/layouts/partials/course_cards_level.html
@@ -2,12 +2,15 @@
   {{- $levelSearchUrl := partial "get_search_url.html" (dict "key" "level" "value" .) -}}
   <a href="{{ $levelSearchUrl }}">{{ . }}</a>
 {{- else -}}
-  {{- $numLevels := len . -}}
-  {{- range $index, $level := . -}}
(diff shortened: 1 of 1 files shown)


MAbdurrehman12 (approved):
Looks good 👍

mbertrand (approved):
It works, was this happening due to a specific course?  There were some recent ocw-www failures on RC but they were caused by my PR changes to the `/api/websites/` response which I will be reverting.

Wassaf-Shahzad:
> It works, was this happening due to a specific course? There were some recent ocw-www failures on RC but they were caused by my PR changes to the `/api/websites/` response which I will be reverting.

@mbertrand Yes probably a course which does not have a course level attribute (I have no idea idea about the course since i couldn't reproduce it locally )

MAbdurrehman12:
@Wassaf-Shahzad @mbertrand Is this good to merge?
So I can rebase a couple of other PRs

PR #655, opened 2022-04-19
fix: change in jquery ready handler

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/ocw-hugo-themes/issues/651

#### What's this PR do?
- Updates jQuery ready handler from `$(document).ready(() => ` to `$(function() {` as it seems like that `ready` is deprecated now
- Note: I was unable to reproduce the error locally so I'm not very sure whether this will fix the issue or not but since `ready` is deprecated and usually the cause of `something not defined` error is due to it's usage before it's initialisation/declaration. And `$(function() {` is also the recommended way as stated [here](https://api.jquery.com/ready/)

- Try to reproduce the issue locally
- Checkout to this branch
- Check whether the issue is reproduced again or not

### DIFF
--- a/course/assets/course.js
+++ b/course/assets/course.js
@@ -17,7 +17,7 @@ import {
   showSolution
 } from "./js/quiz_multiple_choice"
 
-$(document).ready(() => {
+$(function() {
   initPdfViewers()
   initDesktopCourseInfoToggle()
   initCourseInfoExpander(document)


pdpinch:
@gumaerc do we have any way to keep track of where we are using jquery in ocw?

gumaerc (changes requested):
Good catch here.  Can we also make this change in `base-theme/assets/index.ts` as well as `www/assets/www.tsx`?  There are calls there to `$(document).ready` as well.

MAbdurrehman12:
> Can we also make this change in base-theme/assets/index.ts as well as www/assets/www.tsx

Sure, thanks for pointing them out.
btw, I was thinking that `$(document).ready(() =>` is deprecated, not removed, so do you think this is actually the cause of error mentioned above? 
Maybe, its worth giving a try

gumaerc:
@MAbdurrehman12 I just looked at the original issue in Sentry and it is pretty odd.  I'm not able to reproduce it on the live site on the page indicated, or locally.  According to the stack trace, the error was triggered upon toggling the course info.  I tried this on multiple pages and was not able to reproduce the error.  Either way, this seems like a prudent change that we should make with the deprecation of `$(document).ready` regardless.  I'm just not sure it's related to the issue you linked in the description.  My hunch is that maybe VideoJS (which is bundled into our `course.js` webpack bundle) had some code gated behind `$(document).ready` and maybe they committed a change that fixed it?  I can't find any evidence of this in their Github though.

MAbdurrehman12:
@gumaerc Thanks for looking into this.

> the error was triggered upon toggling the course info. I tried this on multiple pages and was not able to reproduce the error.

Yes, I also tried to reproduce it but couldn't.

> this seems like a prudent change that we should make with the deprecation of $(document).ready regardless

I agree, let's make this change and proceed.

gumaerc (approved):
:+1: LGTM

PR #662, opened 2022-04-25
feat: error handling in search API

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/646

#### What's this PR do?
- Adds a util to check whether any API was successful or failed by checking it's `status`
- Checks whether the search API was successful or not.
- If search API failed, then an error message is shown to the user with contact us link.

- Execute the search API.
- Test and verify that everything is working as expected.
- Now we want the API to fail, this can be done by any method  (give an invalid URL, let the code break from any other method etc)
- Verify that now an error msg is shown to the user with contact us link which takes the user to the contact page.
- Any use-case which needs to be tested.

image

image

### DIFF
--- a/www/assets/js/components/SearchPage.tsx
+++ b/www/assets/js/components/SearchPage.tsx
@@ -321,7 +321,8 @@ export default function SearchPage() {
                       <a href={CONTACT_URL}>
                         <b>contact us</b>
                       </a>{" "}
-                      any time.
+                      with the details of what you were trying to do, and what
+                      happened.
                     </span>
                   </div>
                 ) : completedInitialLoad ? (
--- a/www/assets/js/components/SearchPage.tsx
+++ b/www/assets/js/components/SearchPage.tsx
@@ -61,13 +61,13 @@ export default function SearchPage() {
         size: SEARCH_PAGE_SIZE,
         sort: sort
       })
+      setRequestInFlight(false)
 
       if (newResults["apiFailed"]) {
         setSearchApiFailed(true)
         return
       }
 
-      setRequestInFlight(false)
       const { suggest } = newResults
       if (!emptyOrNil(suggest) && !emptyOrNil(text)) {
         setSuggestions(
--- a/www/assets/js/components/SearchPage.tsx
+++ b/www/assets/js/components/SearchPage.tsx
@@ -54,18 +54,19 @@ export default function SearchPage() {
       }
 
       setRequestInFlight(true)
-      const { results, apiSuccessful } = await search({
+      const newResults = await search({
         text,
         from,
         activeFacets,
         size: SEARCH_PAGE_SIZE,
         sort: sort
       })
-      if (!apiSuccessful) {
+
+      if (newResults["apiFailed"]) {
         setSearchApiFailed(true)
         return
       }
-      const newResults = results
+
       setRequestInFlight(false)
       const { suggest } = newResults
       if (!emptyOrNil(suggest) && !emptyOrNil(text)) {
--- a/www/assets/js/components/SearchPage.tsx
+++ b/www/assets/js/components/SearchPage.tsx
@@ -15,7 +15,7 @@ import Loading, { Spinner } from "./Loading"
 
 import { search } from "../lib/api"
 import { searchResultToLearningResource } from "../lib/search"
-import { COURSENUM_SORT_FIELD } from "../lib/constants"
+import { COURSENUM_SORT_FIELD, CONTACT_URL } from "../lib/constants"
 import { emptyOrNil, isDoubleQuoted } from "../lib/util"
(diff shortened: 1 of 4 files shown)


MAbdurrehman12:
@pdpinch Can you check the error message and suggest what should be the text of it? btw the contact us link also doesn't seem to be much prominent, maybe we can change its color to indicate that its a link or replace it with a button under text. What do you suggest?

I think I missed this:
> I don't know why we got a 400. I'd like to know. I think we should log the query when that happens.

@abeglova Do we have any predefined function/API call/mechanism to log the query?

www/assets/js/lib/api.ts
```diff
@@ -12,6 +13,12 @@ export const search = async (params: SearchQueryParams) => {
     })
   })
 
-  const results = await response.json()
+  let results: any = {}
+  if (isApiSuccessful(response.status)) {
```
gumaerc, line 17: I'm not 100% certain about this but I think that due to where this is being checked, certain scenarios don't result in an error being thrown.  I tested 3 scenarios here:

1.  CORS error - This was an easy one to test because all you have to do is point `SEARCH_API_URL` at RC or production and try and hit it in a normal browser session.  This resulted in an infinite loading animation and a CORS error in the console.
2. Domain unreachable - I intentionally redirected `discussions-rc.odl.mit.edu` in my `/etc/hosts` file to `localhost`, which would cause the request to fail.  This resulted in an infinite loading animation and an error in the console.
3. Intentionally set a blank body in the code - This was the only way I was able to actually see the error message.  I went into `api.ts` and intentionally commented out the body line, replacing it with a blank string.  This ensured the API would return a 500 which allowed me to see the error.

So, I think we should investigate how to make sure cases 1 & 2 are handled.

gumaerc (changes requested):
I found a couple scenarios in which the error is not caught:

www/assets/js/lib/api.ts
MAbdurrehman12: Thank you for doing this testing @gumaerc 
Regarding the CORS issue, I thought that it might be something out of scope as its occurrence and fix are via the API which we are hitting, but we can try to handle it here.
So I'll look into these 1 and 2 use-cases.

MAbdurrehman12:
@gumaerc Can you please have another look? Scenario 1 and 2, mentioned above, are handled now.
If you have some more use-cases in mind, I think, it would be good to test them too.
also, do you have any thoughts on this [comment](https://github.com/mitodl/ocw-hugo-themes/pull/662#issuecomment-1109039986)?

www/assets/js/components/SearchPage.tsx
```diff
@@ -306,7 +313,18 @@ export default function SearchPage() {
...
-                {completedInitialLoad ? (
+                {searchApiFailed ? (
+                  <div className="no-results-found">
+                    <span>
+                      Oops! Something went wrong. Please accept our apologies
+                      and feel free to{" "}
+                      <a href={CONTACT_URL}>
+                        <b>contact us</b>
+                      </a>{" "}
+                      any time.
```
pdpinch, line 324:
```suggestion
                      with the details of what you were trying to do, and what happened.
```

gumaerc (approved):
:+1: LGTM

As far as error reporting, we use Sentry for that and the JS lib should be included in our bundle.  If you want to intentionally log errors to sentry at runtime, you can use `Sentry.captureException` or `Sentry.captureMessage`: https://docs.sentry.io/platforms/javascript/usage/

MAbdurrehman12:
@gumaerc Thank you for reviewing the PR. 
I have updated the PR to log errors in sentry. Can you please have another quick look? 
If all is okay then I'll merge it!

gumaerc:
@MAbdurrehman12 the only issue I see here is a bit out of scope, and I think we should create a separate issue / PR for it.  If you look at `sentry.js`, we've got the DSN hard coded.  This means that people spinning up stuff locally will log errors into Sentry from their local instance.  This should be disabled somehow for local development.  I would say this PR is okay to merge though.

MAbdurrehman12:
> @MAbdurrehman12 the only issue I see here is a bit out of scope, and I think we should create a separate issue / PR for it. If you look at `sentry.js`, we've got the DSN hard coded. This means that people spinning up stuff locally will log errors into Sentry from their local instance. This should be disabled somehow for local development. I would say this PR is okay to merge though.

Yeah, I was thinking the same. Maybe we can move it to `env` or have a flag that controls whether to log errors into sentry or not.
Anyhow, we can handle it in a separate issue, I'll merge this PR.

Thanks

PR #664, opened 2022-04-27
feat: error handling for localstorage

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/645

#### What's this PR do?
- Add utils for setting and getting data from local storage with checks and error handling so in case an exception is thrown, that exception is caught silently and a message is shown in the console.
- Some more info:
    - I think this type of error can occur when the user has blocked cookies and we try to access cookies either via local storage or any other way. 
    - We are accessing local storage in `notifications.js` and this PR fixes it by using the util in which we are catching all exceptions.
    - But I think cookies are also being accessed in third-party libraries like `pdfjs` and `youtube.js` and this error can occur there too if a user has blocked cookies. So for this, I think we'll have to figure out a good approach to handle this.

- Go to chrome settings -> Cookies and other site data and block all cookies.
- Go to ocw-www, add notification in content so a notification banner is shown, close the banner.
- Open console and you will be seeing this type of error: `Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`  
- Now checkout this branch and repeat the above steps and verify that now no error is shown. Only a normal `console log` will be shown regarding the error.

### DIFF
--- a/base-theme/assets/index.ts
+++ b/base-theme/assets/index.ts
@@ -10,6 +10,7 @@ import "shifty"
 import "hammerjs"
 import "imagesloaded"
 import "screenfull"
+import "./js/utils"
 
 import { initSentry } from "./js/sentry"
 
--- a/base-theme/assets/js/utils.ts
+++ b/base-theme/assets/js/utils.ts
@@ -1,0 +1,41 @@
+export const setLocalStorageItem = (key: string, value: any) => {
+  return setOrGetLocalStorageItemHelper("set", key, value)
+}
+
+export const getLocalStorageItem = (key: string): any => {
+  return setOrGetLocalStorageItemHelper("get", key)
+}
+
+/**
+ * depending upon action, either set or get a value in localstorage with validation/checks to avoid error/exception
+ *
+ * @param {string} action either "set" or "get"
+ * @param {string} key The key against which the data is/will be stored
+ * @param {any} value actual data/value to store, if any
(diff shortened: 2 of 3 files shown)


MAbdurrehman12:
@pdpinch 
I think this type of error can occur when the user has blocked cookies and we try to access cookies either via local storage or any other way. (I have reproduced the issue by blocking the cookies in my chrome settings.)

We are accessing local storage in `notifications.js` and this PR fixes it.

But I think cookies are also being accessed in third-party libraries like `pdfjs`,  `youtube.js`  etc and this error can easily occur from there too if a user has blocked cookies. So, I think we'll have to figure out a good approach to handle this third-party use-case too.

For our own functions, which uses localstorage, we can use the utils created in this PR.
 
@gumaerc Do you have any thoughts/suggestions?

gumaerc (approved):
:+1: LGTM

MAbdurrehman12:
Correction: 

> I think this type of error can occur when the user has blocked cookies and `we try to access cookies either via local storage or any other way.`

The highlighted statement is wrong as local/session storage does not use cookies. Cookies and local/session storage are different things. So when a user blocks cookies in browser settings, it is the browser that also disables local/session storage. 
This [comment](https://stackoverflow.com/a/31164410/19125174) has a more detailed explanation.

Issue #657, opened 2022-04-20
Sanitize search parameters


abeglova:
OCW search gets parameters from the url and uses them to construct the search query. We should sanitize these parameters so that for example,  https://ocw.mit.edu/search/?l=Hi%20Anastasia does not add "Hi Anastasia" as a level

Also this error:
https://github.com/mitodl/open-discussions/issues/3555

Is being caused by hubspot parameters being appended to the sort parameter

PR #668, opened 2022-05-04
sanitize facets

abeglova:
#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested

closes https://github.com/mitodl/ocw-hugo-themes/issues/657

#### What's this PR do?
This pr sanitizes facets so that arbitrary facets are not pulled in from the url to search

Run `npm run start:www`
Click around search selecting various facets. Verify that everything looks normal.

Try to enter an invalid facet value in the url, for example http://localhost:3000/search/?d=NotADepartment

Verify that the invalid facet is ignored


mbertrand (approved):
Have some concerns about the hard-coded facet values but I couldn't think of a better approach assuming these values should rarely change.    Functionality works great.  👍

Issue #658, opened 2022-04-20
video: update player design on single-video page


pdpinch:
#### Designs and Mockups

https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/461529499/comments?scrollOffset=576

Note that you will need to scroll to the right to see the new player design. 

#### Acceptance Criteria:

- [ ] move player controls below video player and make them permanently visible, instead of displaying on hover
- [ ] attach the transcript widget to the video player
- [ ] #731
- [ ] link the youtube button to the youtube page for this video
- [ ] make a screenshot of a single video page the includes related resources ([example](https://ocw.mit.edu/courses/16-885j-aircraft-systems-engineering-fall-2005/resources/lecture-3/))
- [ ] make a screenshot of a page with an embedded video player ([example](https://ocw.mit.edu/courses/16-885j-aircraft-systems-engineering-fall-2005/pages/related-resources/)) 

#### related issues 

- #139 
- #140 
- update player design for the embedded player

pdpinch:
We have a design now for the embedded video player. It's very similar to the single-video page player, so perhaps we can reuse the same code. See https://projects.invisionapp.com/share/CG122QSAFDRS/#/screens/461529499/comments and scroll right to "View Video page"

PR #678, opened 2022-05-10
video player design update

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/658

#### What's this PR do?
- Updates the styling according to the new [design](https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/461529499/comments?scrollOffset=576).
- Moves player controls below the video player and make them permanently visible, instead of displaying on hover
- Attaches the transcript widget to the video player
- Link the youtube button to the youtube page for the current video
- **Note:** A separate [issue](https://github.com/mitodl/ocw-hugo-themes/issues/731) has been created for moving download buttons to in video control bar. 
 
- Verify the points mentioned above.
- Test any use-case that needs to be tested.

image

image

Embedded video player:
image

When the page loads and the video has not started yet: 

image

image

### DIFF
--- a/course/assets/css/videojs_player.scss
+++ b/course/assets/css/videojs_player.scss
@@ -1,3 +1,8 @@
+// color variables
+$control-item-color: $dark-gray;
+$control-item-bg-color: $light-gray;
+$slider-color: $pink;
+
 .vjs-ocw.video-js {
   .ytp-chrome-top-buttons {
     display: none;
@@ -12,8 +17,14 @@
 
   .vjs-control-bar {
     height: 4.4em;
-    background-color: #2b333f;
+    background-color: $control-item-bg-color;
+    color: $control-item-color;
     transition: visibility 2s, opacity 0.1s;
+    bottom: auto;
+  }
+
+  .vjs-menu-content{
+    background-color: $control-item-bg-color !important;
   }
 
   .vjs-volume-bar {
@@ -28,6 +39,11 @@
     min-width: 50px;
   }
 
+  .vjs-volume-level{
+    background-color: $control-item-color;
+  }
+
+
   .vjs-progress-control {
     position: absolute;
     bottom: 37px;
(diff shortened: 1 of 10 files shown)


course/assets/css/videojs_player.scss
```diff
@@ -12,8 +17,14 @@
 
   .vjs-control-bar {
     height: 4.4em;
-    background-color: #2b333f;
+    background-color: $control-item-bg-color;
+    color: $control-item-color;
     transition: visibility 2s, opacity 0.1s;
+    bottom: auto;
+  }
```
ChristopherChudzicki, line 24: @MAbdurrehman12 I was reading about videojs today hoping to have some insight re  (which I don't have much yet). But regarding
> However, I'm looking into it, as to how we can show the controls even when the video has not played yet.

it seems that

```suggestion
    display: flex;
  }
```
works. Basically, just overriding the `display:none` specified by `.video-js .video-js `. Doesn't feel great messing with their CSS like this, but I guess it's what they recommend https://videojs.com/guides/skins/#creating-a-skin

MAbdurrehman12: Thank you @ChristopherChudzicki for looking into this. 

> Doesn't feel great messing with their CSS like this

I agree with this so I also tried to find any configuration for it but couldn't find any, hence I think, for now, we can override the default CSS (which we have already done at a few places).

However, the customization of the controls is still a blocker. 1 solution is to replace `video.js controls` with `youtube.js controls` and if we do that, then I think all these design updates of controls, including the one you suggested, will not be used, but I'll update this PR for now.

Thank you for this suggestion!

abeglova (approved):
LGTM

Issue #683, opened 2022-05-11
lighthouse broken


gumaerc:
### Steps to Reproduce

Create any PR in `ocw-hugo-themes`

### Expected Behavior

The lighthouse build succeeds

### Actual Behavior

The lighthouse build fails

### Stacktrace

```
Run foo-software/lighthouse-check-action@master
lighthouse-check: Auditing mobile (1/2): https://ocw-hugo-themes-pr-682--ocw-next.netlify.app/
lighthouse-check: Auditing mobile (2/2): https://ocw-hugo-themes-pr-6[8](https://github.com/mitodl/ocw-hugo-themes/runs/6394098269?check_suite_focus=true#step:5:8)2--ocw-next.netlify.app/search/


+--------------------------------------------------------------------------+
|                             Lighthouse Audit                             |
+--------------------------------------------------------------------------+

URL: https://ocw-hugo-themes-pr-682--ocw-next.netlify.app/
Device: mobile
Local Report: /home/runner/work/ocw-hugo-themes/ocw-hugo-themes/tmp/artifacts/report-16522[9](https://github.com/mitodl/ocw-hugo-themes/runs/6394098269?check_suite_focus=true#step:5:9)5050695.html


lighthouse-check:
 Error: Table must define at least one column.
    at validateTableData (/home/runner/work/_actions/foo-software/lighthouse-check-action/master/node_modules/table/dist/validateTableData.js:30:[11](https://github.com/mitodl/ocw-hugo-themes/runs/6394098269?check_suite_focus=true#step:5:11))
    at table (/home/runner/work/_actions/foo-software/lighthouse-check-action/master/node_modules/table/dist/table.js:96:34)
    at /home/runner/work/_actions/foo-software/lighthouse-check-action/master/node_modules/@foo-software/lighthouse-check/dist/logResults.js:56:34
    at Array.forEach (<anonymous>)
    at _default (/home/runner/work/_actions/foo-software/lighthouse-check-action/master/node_modules/@foo-software/lighthouse-check/dist/logResults.js:39:11)
    at /home/runner/work/_actions/foo-software/lighthouse-check-action/master/node_modules/@foo-software/lighthouse-check/dist/lighthouseCheck.js:[22](https://github.com/mitodl/ocw-hugo-themes/runs/6394098269?check_suite_focus=true#step:5:22)3:33
Error: Table must define at least one column.
```

PR #685, opened 2022-05-12
fix lighthouse

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/683

#### What's this PR do?
This PR simply updates the version of Node used when running Lighthouse.  An older version was being used which caused the check not to work correctly

The lighthouse check on this PR should pass, and you should be able to visit the Netlify deploy and see the site, although some images and search will not work.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -100,7 +100,7 @@ jobs:
       - name: Set up node.js
         uses: actions/setup-node@v2-beta
         with:
-          node-version: 12.13.1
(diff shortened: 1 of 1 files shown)


MAbdurrehman12 (approved):
Works good 👍 
I also tested this with one of my old PR, by rebasing it with the main. 
Previously, lighthouse was failing, but now, with this node version, it works good!

Issue #699, opened 2022-05-19
theme builds urls using name instead of url_path


gumaerc:
### Steps to Reproduce

 - Set a custom `url_path` on a site in `ocw-studio` and publish that course
 - Clone `ocw-www` locally and set your `OCW_STUDIO_BASE_URL` to the same environment you created the test course on
 - `npm run start:www`

### Expected Behavior

 - Your course shows up in the new courses carousel with a correct URL

### Actual Behavior

 - It has an incorrect URL

PR #700, opened 2022-05-19
update theme to use url_path

gumaerc:
#### Pre-Flight checklist

- [ ] Testing

Closes https://github.com/mitodl/ocw-hugo-themes/issues/699

#### What's this PR do?
In https://github.com/mitodl/ocw-studio/pull/1316 and subsequently https://github.com/mitodl/ocw-studio/pull/1371, we added the ability for content authors to customize the URL of their site.  This PR changes the theme so that `url_path` is used where it's available, assuming the `url_path` is already prefixed with `root-url-path`.  It's also assumed that `id` in some relation data is filled with the value of `url_path`.

 - Clone `ocw-www` from `ocw-content-rc`
 - In your `.env`, set the following:
```
OCW_STUDIO_BASE_URL=https://ocw-studio-rc.odl.mit.edu/
RESOURCE_BASE_URL=https://ocwnext-rc.odl.mit.edu
STATIC_API_BASE_URL=https://ocwnext-rc.odl.mit.edu/
WWW_CONTENT_PATH=/path/to/ocw-content-rc/ocw-www/
```
 - Run `npm run start:www`
 - Scroll down to the new courses section, click any of the courses and replace the domain with `ocwnext-rc.odl.mit.edu` and the course should load

### DIFF
--- a/www/assets/js/components/SearchResult.tsx
+++ b/www/assets/js/components/SearchResult.tsx
@@ -120,7 +120,7 @@ export function LearningResourceDisplay(props: SRProps) {
   const maxTags = 3
   const runSlug = object.run_slug
   const url = runSlug
-    ? `${runSlug.includes("courses/") ? "/" : "/courses"}${runSlug}`
+    ? `${runSlug.includes("courses/") ? "/" : "/courses/"}${runSlug}`
     : ""
 
   if (isResource(object)) {
--- a/www/assets/js/components/SearchResult.tsx
+++ b/www/assets/js/components/SearchResult.tsx
@@ -118,6 +118,10 @@ function isResource(object: LearningResource): boolean {
 export function LearningResourceDisplay(props: SRProps) {
   const { object, id } = props
   const maxTags = 3
+  const runSlug = object.run_slug
+  const url = runSlug
+    ? `${runSlug.includes("courses/") ? "/" : "/courses"}${runSlug}`
+    : ""
 
   if (isResource(object)) {
     return (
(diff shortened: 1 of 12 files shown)


mbertrand (changes requested):
There are other files where I think `"/courses/"  $courseItem.name` or `"/courses/" .id` will need to be replaced:

featured_course_cards.html
layouts/collection/single.html
fields/layouts/home.html
layouts/subfields/single.html
layouts/course-lists/single.html
layouts/resource-collections/single.html

www/assets/js/lib/search.ts
```diff
@@ -537,7 +537,7 @@ export const courseJSONToLearningResource = (
   run_title: null,
   run_slug: null,
   content_type: null,
-  url: `/courses/${name}/`,
+  url: `/${name}/`,
```
mbertrand, line 540: There are a few other places in this file where `courses/` is prepended.  I wonder if as a temporary measure we should have some function that adds `courses` if not there, until the open-discussions change has been deployed and the index fully updated, to avoid any downtime?

```javascript
url: `${name.startsWith("courses") ? "/" : "/courses/"}${name}/`
```

mbertrand (changes requested):

mbertrand (changes requested):

www/assets/js/components/SearchResult.tsx
```diff
@@ -118,6 +118,10 @@ function isResource(object: LearningResource): boolean {
 export function LearningResourceDisplay(props: SRProps) {
   const { object, id } = props
   const maxTags = 3
+  const runSlug = object.run_slug
+  const url = runSlug
+    ? `${runSlug.includes("courses/") ? "/" : "/courses"}${runSlug}`
```
mbertrand, line 123: `runSlug` isn't going to include a preceding slash either way, so I think this should be:

```javascript
${runSlug.includes("courses/") ? "/" : "/courses/"}
```

mbertrand (changes requested):

www/assets/js/components/SearchResult.tsx
gumaerc: Thanks, this should be all set now.

mbertrand (approved):
👍

PR #712, opened 2022-05-27
fix slash prefix issue with course sitemap disallowed url check

gumaerc:
#### Pre-Flight checklist

- [ ] Testing

Hotfix for https://github.com/mitodl/ocw-hugo-themes/pull/710

#### What's this PR do?
In https://github.com/mitodl/ocw-hugo-themes/pull/710, we fixed an issue with the course sitemap construction not taking the `--baseUrl` setting into account.  Something that I failed to notice was that our `--baseUrl` setting in RC and production doesn't have a trailing slash.  This causes the check to fail, because there will be a preceeding slash on `$path` after stripping the `baseUrl`.  This PR does just that, strips the preceeding slash.

Follow the same testing instructions in https://github.com/mitodl/ocw-hugo-themes/pull/710.  Beyond that, further validation will be needed in RC


abeglova (approved):
LGTM

Issue #708, opened 2022-05-26
ocw-www resource pages don't publish 


pdpinch:
### Steps to Reproduce

1. Add a resource to ocw-www, like https://ocw-studio-rc.odl.mit.edu/sites/ocw-www/type/resource/edit/65120dad-ad2a-41d8-98ac-54d9339b4468/
2. Add a link to it on a page, like https://ocw-studio-rc.odl.mit.edu/sites/ocw-www/type/page/edit/fabb4fde-a5ad-4013-962c-34e3273d183f/
3. publish to stage

### Expected Behavior

- the link should work
- there should be a resource page at https://ocw-draft-qa.global.ssl.fastly.net/resources/full-study-gender-roles-and-occupations-v2/

### Actual Behavior

- link is broken
- no resource pages at all

MAbdurrehman12:
@pdpinch We don't have any `resources` layout in `themes/base-theme` or `themes/www`, though we have one in `themes/courses`. Due to this, we can load/serve `course` resources but cannot load `ocw-www` resources. 

So I think moving `resources layout`, along with several other `partials` which are being used in this layout, to `base-theme` will do the job!  But I wonder why it wasn't done in the first place? Maybe it wasn't required at that time or there was any other reason, not sure. If @gumaerc has any thoughts on this?

Another thing is that the `resources` URL is being redirected to `courses`.
e.g: `ocw.mit.edu/resources/resource-slug` redirects to `ocw.mit.edu/courses/resource-slug`

So the resource which you're trying to access here `fastly/resources/full-study-gender-roles-and-occupations-v2/` is being redirected to `fastly/courses/full-study-gender-roles-and-occupations-v2/` which will certainly throw `404`.

I'm not sure whether this has been done deliberately due to some reason or this is a bug at DevOps side, so even if we move the `resources` layout to `base-theme`, this redirection will still be an issue for `ocw-www` resources.

pdpinch:
You are correct about the missing layout. Carey and I agree with the plan to move it from courses to the base theme.

As for the redirect, it only applies when the URL is a 404. Once we have the files in place, it won’t interfere.

MAbdurrehman12:
Thanks for the update.
Did Carey start working on this or I can create its PR?

PR #713, opened 2022-05-31
resources layout moved to base theme from courses

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

- Closes https://github.com/mitodl/ocw-hugo-themes/issues/708

#### What's this PR do?
- Creates a new partial in base-theme named `resource_layout.html` and moves all the code of `courses/layouts/resources/single.html` to this partial.
- Uses this partial in `courses/layouts/resources/single.html` and creates a new `resources layout` in www (i.e: `www/layouts/resources/`) which again uses this partial. 
- Moves some partials, which are being used in both themes, courses, and www from `courses` theme to `base` theme.
- Some little CSS tweaks

- Add some resources to `ocw-www`, like: https://ocw-studio-rc.odl.mit.edu/sites/ocw-www/type/resource/edit/65120dad-ad2a-41d8-98ac-54d9339b4468/
- Publish it.
- Take a pull on your local content. i.e: `WWW_CONTENT_PATH` 
- build `ocw-www` locally. i.e: `npm run start:www`
- Verify that all the resources added in step 1 are being loaded/server at `http://localhost:3000/resources/resource-slug`.
- Along with this, please verify that all the resources of different courses are being loaded/served like before. There is no change in course resources.
- Any other use-case which needs to be checked/tested.

#### Screenshots 
image

image


www/layouts/resources/single.html
```diff
@@ -0,0 +1,11 @@
+{{ define "main" }}
+{{ block "header" . }}
+  {{ partialCached "header" . }}
+{{ end }}
+<div class="container mt-5">
+  <div class="rounded mb-2">
+    {{ partial "content_header.html" . }}
+  </div>
+  {{ partial "resource_layout.html" . }}
```
gumaerc, line 9: Take it or leave it, but I think a better name for this partial might be `resource_body.html`.

base-theme/layouts/partials/resource_layout.html
```diff
@@ -0,0 +1,30 @@
...
+      {{- end -}}
+      {{- if .Params.file -}}
+      <div class="row">
+          <div class="col-12 d-flex flex-direction-row align-items-center">
+              <a class="download-file" href="{{ partial "resource_url" .Params.file }}"><span class="material-icons">file_download</span> Download File</a>
+          </div>
+      </div>
+      {{- end -}}
+  {{- else -}}
+      {{ partial "video.html" . }}
```
gumaerc, line 23: So I tested video resources with this and while the partials for the video resource page have been moved over, there seems to be an issue with the video metadata.  The [starter configuration](https://github.com/mitodl/ocw-hugo-projects/blob/main/ocw-www/ocw-studio.yaml) for `ocw-www` doesn't have all the necessary video metadata that the `ocw-course` starter has, so that will need to be added.  There may also be some changes necessary in `ocw-studio` relating to the processing of uploaded videos, but I'm not certain.  This is outside the scope of this PR, but this will need to be addressed if we want videos to work in `ocw-www`.

gumaerc (changes requested):
The PDF viewer issues should be sorted out in this PR, but the video issues are out of scope like I noted below.  Overall, this looks good though and works well for displaying resources added to `ocw-www`.

MAbdurrehman12:
Thanks for the review @gumaerc.
I have made the changes, can you please have another look? 
let me know if you find anything else. Thanks!

gumaerc (changes requested):
The changes look good.  I just noticed one thing with the PDF viewer.  Now that it's being loaded outside of a course page, there is a bit of a styling issue with the height.  The PDF viewer's height in `ocw-www` is too short:

image

In the course theme, it extends to show at least one page:
image

This is because there is a wrapper around the container for course pages that has the CSS rule `min-height: 100vh`, guaranteeing that the page will be at least as tall as the browser window.  I tried this out on the `iframe` element containing the PDF viewer, and it fixed the problem for me.

base-theme/layouts/partials/resource_layout.html
pdpinch: We can leave videos for later. What we've been waiting for, for a while, is PDFs.

pdpinch:
Sorry if I'm being paranoid, but do we have test coverage that will catch any side effects on _course_ resource pages?

gumaerc (approved):
@MAbdurrehman12 The changes look good, and this PR is good to go :+1: 

@pdpinch I manually tested loading every type of resource page in a course locally and everything looks good.  We still don't have automated testing of Hugo output, which is why I created this discussion to talk about how we might accomplish that: https://github.com/mitodl/hq/discussions/45

Issue #608, opened 2022-04-05
stand-alone 404 page


pdpinch:
The 404 page was built in studio, but it's served directly from the CDN. The current version has dependencies on versioned CSS and JavaScript files in the theme, which we don't keep indefinitely. 

The 404 page needs to be rewritten to use (minimal) in-line JavaScript and CSS

#### Designs and Mockups

See https://ocw.mit.edu/pages/page-not-found/

#### Acceptance Criteria:

- [ ] If it requires an javascript, add it to the header
- [ ] move all necessary styles out of external css files
- [ ] follow hugo best practices for where to put the file in the theme
- [ ] We only need one of these, so it can go in the ocw-www theme

pdpinch:
@Wassaf-Shahzad where can I tell @mitodl/devops to get the HTML for this page? They will need to add it to the Fastly configuration.

Wassaf-Shahzad:
> @Wassaf-Shahzad where can I tell @mitodl/devops to get the HTML for this page? They will need to add it to the Fastly configuration.

Form the code or by url ?
if by url on rc its available at https://ocwnext-rc.odl.mit.edu/404.html

pdpinch:
When I load the page at https://ocwnext-rc.odl.mit.edu/404.html it still has external js and css dependencies: 

www.a3c26.css
main.a3c26.css
www.a3c26.js
main.a3c26.js

Can those be removed? 

My concern is that those versions of the css and js files are not guaranteed to be permanently available.

Wassaf-Shahzad:
> When I load the page at https://ocwnext-rc.odl.mit.edu/404.html it still has external js and css dependencies:
> 
> [www.a3c26.css](http://www.a3c26.css) main.a3c26.css [www.a3c26.js](http://www.a3c26.js) main.a3c26.js
> 
> Can those be removed?
> 
> My concern is that those versions of the css and js files are not guaranteed to be permanently available.

@pdpinch I can remove our dependency on the css files but not the JS files since they contain code for the header and footer.

I think simplest solution would be to redesign the 404 page with put any use of hugo partials like header and footer.
Since adding these would mean to code for both of this in 404.html rather than only including partials

pdpinch:
> redesign the 404 page without any use of hugo partials like header and footer

That makes sense to me. I don't see why the 404 page would need any JavaScript at all.

pdpinch:
For example, I'm seeing the notification banner on the 404 page currently. We don't need that.

[...]

PR #716, opened 2022-05-31
updated 404 page

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

Part of #608

#### What's this PR do?
- Moves all the code for the external css to the 404 page itself

- Verify 404 page does not depend on any external css.

Screenshot 2022-05-31 at 4 30 42 PM

### DIFF
--- a/www/layouts/404.html
+++ b/www/layouts/404.html
@@ -1,5 +1,15 @@
-{{ define "main" }}
 <style>
+  body {
+    font-family: "Helvetica";
+    font-size: 16px;
+    margin: 0;
+    text-align: left;
+    font-weight: 400;
+    line-height: 1.5;
+    color: #212529;
+    background-color: #ffff;
+  }
+
   .page-single {
     max-width: 100%;
     margin: 0 auto;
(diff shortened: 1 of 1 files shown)


pdpinch:
Is it possible to include the site header and footer? (Sorry if I asked this before)

pdpinch:
> I would probably need to add CDN links to bootstrap and JQuery

I get bootstrap, but why jquery?

In any case, not a blocker.

Wassaf-Shahzad:
> > I would probably need to add CDN links to bootstrap and JQuery
> 
> I get bootstrap, but why jquery?
> 
> In any case, not a blocker.
We dont need jquery if we are using bootstrap 5 but some components in bootstrap 4 require it https://www.w3schools.com/bootstrap4/bootstrap_get_started.asp#:~:text=Bootstrap%204%20use%20jQuery%20and,you%20don't%20need%20them.

I am certain that we dont use any jquery for the footer but I am not sure about the header

gumaerc:
I think we still have a few lingering dependencies on JQuery and Popper that we would need to deprecate / work around, but we should definitely upgrade to Bootstrap 5 at some point.  As evidenced by Paul's recent trouble spinning up `ocw-studio` on an M1 Mac, we don't go through our projects and upgrade dependencies often enough, at least not until it gets to the point where it causes a technical problem.

This being said, with the goal just being to get a static 404 page working, I think it's okay to create something like a `404_header.html` and `404_footer.html` that preserves the design in the most minimal and inlined way possible.  Looking at the templates, I think the reason we need Bootstrap here are automatically handling layout switching and bringing in the hamburger menu for mobile.  If we could imbed some CSS media queries to accomplish a minimal version of the same thing here, I think that would be fine.  Maybe on the 404 page header, we don't need a hamburger menu but instead could just have a link to the home page on the header image that will bring them there where the whole site is available?

pdpinch:
We're spending way too much time on this. Let's merge as is and I'll decided later if I really want a header and a footer.

Issue #705, opened 2022-05-26
Course lists aren't sorted according to studio order


pdpinch:
(I'm not sure if this is a bug in the way studio writes out course lists, or in the way the theme interprets the list) 

### Steps to Reproduce

1. Take a look at the order of courses in the list at https://ocw-studio.odl.mit.edu/sites/ocw-www/type/course-lists/edit/2b404bc3-dd20-4e8c-b4c2-bc70f9d4a838/
2. Note that the first course is "Fundamentals of Biology"
3. Take a look at the markdown in github: https://github.mit.edu/mitocwcontent/ocw-www/blame/main/content/course-lists/scholar-courses.md
4. Note that the first course is "Fundamentals of Biology"
5. Take a look at the rendered site: https://ocw.mit.edu/course-lists/scholar-courses/

### Expected Behavior

- The order on https://ocw.mit.edu/course-lists/scholar-courses/ should be the same as the order in studio

### Actual Behavior

- On https://ocw.mit.edu/course-lists/scholar-courses/ the list starts with "Principles of Microeconomics"

MAbdurrehman12:
This issue seems to occur from [here](https://github.com/mitodl/ocw-hugo-themes/blob/ab90cfe81964a6ca7fc972b76e8fc87a4d1f6654/www/layouts/course-lists/single.html#L18) as I think on merging dicts in Hugo, order is not maintained, and then the order of this dis-ordered dict is again disturbed while iterating through it via `Object.entries` [here](https://github.com/mitodl/ocw-hugo-themes/blob/ab90cfe81964a6ca7fc972b76e8fc87a4d1f6654/www/assets/js/hooks/hugo_data.ts#L90). 

I spent some time figuring out a suitable solution/tweak but couldn't come up with anything. I think 1 solution can be to change the approach here from `nested objects` to array of objects, so we can replace `merge` with `append` or something else, as I think with arrays we can maintain the original order.

There might be some other solution that I can't think of at the moment so I wanted to have someone's thoughts on this before implementing it. 
Maybe @gumaerc or @mbertrand?

PR #721, opened 2022-06-01
fix: course list order maintained

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/705

#### What's this PR do?
- Changes the approach of storing course list data from nested objects to array of object so original list order can be maintained.
- Makes appropriate changes in React component and tests

- Checkout this branch.
- Open any course list (e.g: http://localhost:3000/course-lists/open-learning-library/ OR `netlify-link`/course-lists/open-learning-library/)
- Verify that the order of courses in all the course lists is exactly the same as their order in studio.
- Test any other use case that needs to be tested.

#### Screenshots 
screencapture-ocw-hugo-themes-pr-721-ocw-next-netlify-app-course-lists-open-learning-library-2022-06-01-22_07_11

### DIFF
--- a/www/assets/js/lib/util.ts
+++ b/www/assets/js/lib/util.ts
@@ -1,6 +1,8 @@
 import * as Sentry from "@sentry/browser"
 import { either, isEmpty, isNil, match } from "ramda"
 import { STATUS_CODES } from "./constants"
+import { LearningResource, CourseJSONMap } from "../LearningResources"
+import { courseJSONToLearningResource } from "./search"
 
 export const emptyOrNil = either(isEmpty, isNil)
 
@@ -32,4 +34,21 @@ export const sentryCaptureException = (excetpion: any) => {
 
 export const sentryCaptureMessage = (message: string) => {
   Sentry.captureException(message)
+}
+
+/**
+ * Converts course list array into learning resources array while maintaing order
+ */
+export const getLearningResourcesFromCourseList = (
+  courseList: CourseJSONMap[]
+): LearningResource[] => {
+  const numberOfCourses = courseList.length || 0
+  const learningResources: LearningResource[] = []
+  for (let i = 0; i < numberOfCourses; i++) {
+    const key = Object.keys(courseList[i])[0]
+    learningResources.push(
+      courseJSONToLearningResource(key, courseList[i][key])
+    )
+  }
+  return learningResources
 }
--- a/www/assets/js/lib/util.ts
+++ b/www/assets/js/lib/util.ts
@@ -42,13 +42,9 @@ export const sentryCaptureMessage = (message: string) => {
 export const getLearningResourcesFromCourseList = (
   courseList: CourseJSONMap[]
 ): LearningResource[] => {
-  const numberOfCourses = courseList.length || 0
-  const learningResources: LearningResource[] = []
-  for (let i = 0; i < numberOfCourses; i++) {
-    const key = Object.keys(courseList[i])[0]
-    learningResources.push(
-      courseJSONToLearningResource(key, courseList[i][key])
-    )
-  }
+  const learningResources: LearningResource[] = courseList.map(course => {
+    const key = Object.keys(course)[0]
+    return courseJSONToLearningResource(key, course[key])
+  })
   return learningResources
 }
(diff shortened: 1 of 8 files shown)


www/assets/js/lib/util.ts
```diff
@@ -33,3 +35,20 @@ export const sentryCaptureException = (excetpion: any) => {
 export const sentryCaptureMessage = (message: string) => {
   Sentry.captureException(message)
 }
+
+/**
+ * Converts course list array into learning resources array while maintaing order
+ */
+export const getLearningResourcesFromCourseList = (
```
pdpinch, line 42: What does this PR have to do with resources?

gumaerc: This doesn't refer to resources as we define them nextgen OCW, rather this refers to `LearningResource` objects as they're defined in `open-discussions` because that's what's being returned by the Open search API.

MAbdurrehman12: We are converting course JSON to learning resource for every course in a particular course list [here](https://github.com/mitodl/ocw-hugo-themes/blob/96eb72e6ef08cbf8188c7b9189c4c826864368e5/www/assets/js/hooks/hugo_data.ts#L91), which is then rendered by the course list component. 
I have wrapped up that logic of conversion (courselist -> array of learning resources) in a `util` to increase reusability.

www/assets/js/lib/util.ts
```diff
@@ -33,3 +35,20 @@ export const sentryCaptureException = (excetpion: any) => {
...
+  courseList: CourseJSONMap[]
+): LearningResource[] => {
+  const numberOfCourses = courseList.length || 0
+  const learningResources: LearningResource[] = []
+  for (let i = 0; i < numberOfCourses; i++) {
```
gumaerc, line 47: I think we can still use `map` here, no?  Was `map` not maintaining the order? If you need the index from `map`, it will be the second argument passed to the function call. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

gumaerc (changes requested):
The use of `append` over `merge` in the Hugo code makes sense as far as maintaining order, but in the React code I think we should still be using `map` rather than a `for` loop.

www/assets/js/lib/util.ts
MAbdurrehman12: Yeah, I was just being a bit extra-careful :P haha
but `map` is good, thanks for pointing this out. 
I have updated the PR

pdpinch:
1. Is it a problem that in the screenshot all of the thumbnail images are 404s?
2. Out of scope: our accessibility advisor has recommended that the ALT tag for decorative images should be empty, i.e. `alt=""`. I'll open a an issue for this change.

MAbdurrehman12:
> Is it a problem that in the screenshot all of the thumbnail images are 404s?

Nope. It's just on my local setup as I don't have these resources locally. On RC and Prod, it should be fine. 

> Out of scope: our accessibility advisor has recommended that the ALT tag for decorative images should be empty, i.e. alt="". I'll open a an issue for this change.

Sure, btw, it will be a small change: as we only have to set `alt=""` so it can also be done in this same PR.
btw, with alt="" and no image, it will look something like this: 

image

Maybe, one solution can be to place a check: if image is found then show it, if not found then don't even render the image tag (``) so we don't have this broken image section. Not sure if this will be worth the effort.

pdpinch:
> Maybe, one solution can be to place a check: if image is found then show it, if not found then don't even render the image tag () so we don't have this broken image section. Not sure if this will be worth the effort.

thanks for the idea, but I don't think it's worth the effort. A broken image on this page should be fixed, instead of relying on the ALT text.

pdpinch (approved):

gumaerc:
> Nope. It's just on my local setup as I don't have these resources locally. On RC and Prod, it should be fine.

@pdpinch @MAbdurrehman12 For local testing, we have the `RESOURCE_BASE_URL` variable you can set in your `.env`.  This will prefix all direct calls to resources with the URL you supply, so you can point at RC or production resources in your local instance.  In my case, I have `RESOURCE_BASE_URL=https://ocwnext-rc.odl.mit.edu` so my local dev server will write in URLs sourcing images from RC:

image

Issue #703, opened 2022-05-25
move to `resources.GetRemote`


gumaerc:
In https://github.com/mitodl/ocw-hugo-themes/pull/679, we updated the version of Hugo we're using to 0.98.0 by swapping out the `hugo-bin` Node package for `hugo-bin-exitended`, which is more frequently updated.  This was done so that we could use the [`resources.GetRemote`](https://gohugo.io/hugo-pipes/introduction/#get-resource-with-resourcesget-and-resourcesgetremote) function to be able to query the authenticated `published` endpoint from `ocw-studio`.

We should move to using `resources.GetRemote` anywhere we fetch an external file at build time.  The main target for this is our `getJSON` calls, which do not support error handling.  `resources.GetRemote` does support error handling, so we should swap out calls to `getJSON` with it.  Errors during the build process can be caught and logged to Sentry, and then the build can continue.

PR #734, opened 2022-06-08
getJSON replaced with resources.GetRemote

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/703

#### What's this PR do?
- Replaces all remote calls via `getJSON` with `resources.GetRemote` so we have more flexibility in error handling.

- Checkout this branch
- Open all those pages in which this change has been implemented and verify that those pages are rendering/showing content correctly/like before and there is no apparent change in the content.


base-theme/layouts/partials/featured_course_cards.html
```diff
@@ -23,8 +23,12 @@ <h2 class="{{ $klass }}">Featured Courses</h2>
             {{ range $index, $courseItem := $courses }}
               {{- $urlPath := partial "site_root_url.html" $courseItem.id -}}
               {{- $url := delimit (slice (strings.TrimSuffix "/" $staticApiBaseUrl) $urlPath "/data.json") "" -}}
-              {{- $courseData := getJSON $url -}}
-              {{ partial "course_carousel_card.html" (dict "itemsInCarousel" $itemsInCarousel "courseData" $courseData "index" $index "urlPath" $urlPath "numCourses" (len $courses))}}
+              {{ with resources.GetRemote $url }}
+                {{ if not .Err }}
```
MAbdurrehman12, line 27: @gumaerc Do you have any suggestion of which approach to take in order to log an error to Sentry from here in HTML? 
Do we have anything in Hugo that can help us with it?
I think we have `errorf` and `warnf` in Hugo but both will cause the build to fail

www/layouts/partials/new_course_cards.html
```diff
@@ -4,37 +4,43 @@
...
-    {{ $itemsInCarousel := index $carouselInfo "size" }}
-    {{ $breakpointVisibilityClass := index $carouselInfo "class" }}
-    {{ $isMobile := (eq $itemsInCarousel 1) }}
-    {{ $carouselId := (printf "new-course-carousel-%v" $breakpoint) }}
-    <div class="{{ $breakpointVisibilityClass }}">
-      <div id="{{ $carouselId }}" class="carousel slide" data-interval="false" data-touch="true" data-ride="carousel">
-        <div class="carousel-header d-flex flex-row justify-content-between font-weight-bold">
-          <h2>New Courses</h2>
-          {{ partial "carousel_controls.html" $carouselId }}
+{{  with resources.GetRemote (print (strings.TrimSuffix "/" $studioBaseUrl) "/api/websites/?" $typeQuery)  }}
```
gumaerc, line 7: So I did some experimentation with this here in `new_course_cards.html` and here is what I found:

Firstly, Sentry needs to be made available to be called on the frontend.  The way I accomplished this was to modify `sentry.js` to return the `Sentry` instance after init, then assign it to the `window` object on page load: 
`sentry.js`:
```
import * as Sentry from "@sentry/browser"

export const initSentry = () => {
  Sentry.init({
    dsn:
      "https://eee58f41dda54d2b814296e12dced4b7@o48788.ingest.sentry.io/5304953"
  })
  return Sentry
}
```
`index.ts`:
```
import * as Sentry from "@sentry/browser"
...
export interface OCWWindow extends Window {
  $: JQueryStatic
  jQuery: JQueryStatic
  Popper: typeof Popper
  Sentry: typeof Sentry
}
...
$(function() {
  // hacky coming-soon popover
  document.querySelectorAll(".coming-soon").forEach(el => {
    tippy(el, {
      content: "Coming soon!",
      trigger: "click",
      placement: "top"
    })
  })

  window.Sentry = initSentry()
})
```
Secondly, we need a partial to handle logging the errors, so I created `base-theme/layouts/partials/sentry_capture_message.html`:

```
<script type="text/javascript">
  window.addEventListener("load", (event) => {
    window.Sentry.captureMessage("{{ . }}")
    console.error("{{ . }}")
  })
</script>
```

Lastly, in each place where `resources.GetRemote` is used, we need to call our new partial when we want to log the exception.  One thing that I noticed about this function is that `.Err` is only populated when the server actually returns an actual 500 error.  If the resource simply isn't found, `nil` is returned.  In this case, you need to use the `{{ else }}` statement on the `{{ with }}` block itself.  I had thought we might be able to create a partial that can have a closing tag to have some reusability around error handling specific to `resources.GetRemote`, but unfortunately it looks like that's only possible with shortcodes.  So, wherever we use `resources.GetRemote`, we'll have to do something like this (this example was done here in `new_course_cards.html`):

```
...
              {{ $resultsSlice := first 10 $results }}
              {{ range $index, $courseItem := $resultsSlice  }}
                {{- if $courseItem.url_path -}}
                  {{- $url := delimit (slice (strings.TrimSuffix "/" $staticApiBaseUrl) "/" $courseItem.url_path "/data.json") "" -}}
                  {{ with resources.GetRemote $url }}1
                    {{ with .Err }}
                      {{ $errorMessage := printf "New courses carousel on %v failed to fetch %v with error: %v" site.BaseURL $url .Err }}
                      {{ partial "sentry_capture_message.html" $errorMessage }}
                    {{ else }}
                      {{- $courseData := . | unmarshal -}}
                      {{ partial "course_carousel_card.html" (dict "itemsInCarousel" $itemsInCarousel "courseData" $courseData "index" $index "urlPath" $courseItem.url_path "numCourses" (len $resultsSlice))}}
                    {{ end }}
                  {{ else }}
                    {{ $errorMessage := printf "New courses carousel on %v failed to fetch %v (404)" site.BaseURL $url }}
                    {{ partial "sentry_capture_message.html" $errorMessage }}
                  {{ end }}
                {{- end -}}
              {{ end }}
...
```

gumaerc (changes requested):
This is a good start, but some more work needs to be done to properly log errors:

www/layouts/partials/new_course_cards.html
MAbdurrehman12: This is a good suggestion! 
Thanks, @gumaerc, I'll update the PR

www/layouts/partials/new_course_cards.html
```diff
@@ -4,37 +4,46 @@
...
+                  {{- $url := delimit (slice (strings.TrimSuffix "/" $staticApiBaseUrl) "/" $courseItem.url_path "/data.json") "" -}}
+                  {{ with resources.GetRemote $url }}
+                    {{ with .Err }}
+                      {{ $errorMessage := printf "New courses carousel on %v failed to fetch %v with error: %v" site.BaseURL $url .Err }}
+                      {{ partial "sentry_capture_message.html" $errorMessage }}
+                    {{ else }}
+                      {{- $courseData := . | unmarshal -}}
+                      {{ partial "course_carousel_card.html" (dict "itemsInCarousel" $itemsInCarousel "courseData" $courseData "index" $index "urlPath" $courseItem.url_path "numCourses" (len $resultsSlice))}}
+                    {{ end }}
+                  {{ end }}
```
gumaerc, line 37: Unfortunately, I think we need to also have an `{{ else }}` statement on each `[[ with resources.GetRemote $url }}`, because of what I detailed above.  When Hugo gets a 500 error it will populate that `.Err` object, but when it gets a 404 it doesn't seem to hit that block and instead returns `nil` which would fall to the `else` on the `with` statement instead.  This is what I saw in my testing, anyway.

gumaerc (changes requested):
I think we're just missing one more piece:

www/layouts/partials/new_course_cards.html
MAbdurrehman12: Yes, sorry for missing it out.
I'll update the PR

fields/layouts/home.html
```diff
@@ -34,12 +34,25 @@ <h1>
...
-          {{- $subfield_data = $subfield_data | append (dict .id $data) -}}
-      {{- end -}}
+      {{ with $subfield.Params.courses }}
+        {{- range (first 5 $subfield.Params.courses) -}}
+            {{- $id := .id -}}
+            {{- $url := delimit (slice (strings.TrimSuffix "/" $staticApiBaseUrl) "/" $id "/data.json") "" -}}
+            {{ with resources.GetRemote $url }}
+              {{ with .Err }}
+                {{- $errorMessage := printf "Failed to fetch sub-fields on %v via %v with error: %v" site.BaseURL $url .Err -}}
+                {{- partial "sentry_capture_message.html" $errorMessage -}}
```
MAbdurrehman12, line 44: @gumaerc Not sure if this and all other partials, used inside `JS`/`script tag` will work as expected as loading this partial here will render another `script` tag inside an already loaded script.
I tried to test this but couldn't. 
I also tried to use custom/raw JS here but it also had some issues.

Let me know if you have any thoughts on this

gumaerc: @MAbdurrehman12 Good catch here, I didn't notice this scenario.  I think the simplest solution would be to just create another partial, something like `sentry_capture_message_internal.html` with just the raw JS and call it explicitly.  Then, in `sentry_capture_message.html`, we can call that partial as well inside of the script tags.

gumaerc (changes requested):
Just one thing to address here regarding the issue that you found, then this also needs a rebase.  I had to run `yarn install` after that because it includes changes that update dependencies.

MAbdurrehman12:
>  I think the simplest solution would be to just create another partial, something like sentry_capture_message_internal.html with just the raw JS

Yup, or how about passing a flag as a parameter to the same partial and then load JS with/without `script` tag according to that flag, this way, we can reuse the same partial instead of creating another one.

MAbdurrehman12:
> > I think the simplest solution would be to just create another partial, something like sentry_capture_message_internal.html with just the raw JS
> 
> Yup, or how about passing a flag as a parameter to the same partial and then load JS with/without `script` tag according to that flag, this way, we can reuse the same partial instead of creating another one.

I think, your opinion might be better and will require fewer changes. 
So yeah, we can create another partial `sentry_capture_message_internal.html` or maybe `sentry_capture_message_js.html` and use it in JS instead

gumaerc:
@MAbdurrehman12 either way is fine with me

MAbdurrehman12:
@gumaerc I was having some trouble in rendering JS with the approach we decided above. It seems that when we load a partial to render JS inside a script tag, it is rendered as a string and the JS is never executed/compiled. 
So I have moved the Hugo code outside the script tag and now it works well! `<script>` tag is used for JS now.
[Latest commit](https://github.com/mitodl/ocw-hugo-themes/pull/734/commits/189d707d385901328fd0d22f711d5565890b3855) can be seen for these changes.

I'm not sure why Hugo was included/written inside the `<script>` tag in the first place but I think it should not matter. What do you think?

Can you please have another look? and if you find any issues, please let me know

Thanks

gumaerc (approved):
:+1: This looks good to go!

I can't believe I didn't think of just moving the script tag on pages that set `window.courseListsData`, that's a good call.

PR #747, opened 2022-06-10
fix resource filter

abeglova:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

none

#### What's this PR do?
Some [Resource Type facets](https://live-qa.ocw.mit.edu/search/?type=resourcefile) are not clickable. Assignments, Recitations, Online Textbooks, Tutorials don’t work. This fixes the issue

Run search locally. Verify that all resource type filters are clickable.


Ferdi:
@abeglova can you elaborate a bit what the problem is , why it's happening and how we solve it ?

abeglova:
This is happening because we sanitize parameters we send to elastic search so we can't add arbitrary assignment types through the url params. The list of allowed parameters is missing a few. This PR fixes the issue

ChristopherChudzicki (approved):
👍

Issue #751, opened 2022-06-14
lists of course "resources"


pdpinch:
from https://github.com/mitodl/hq/discussions/92

#### Designs and Mockups

[Design](https://invis.io/2A12NFC0HTQJ#/466519313_Resources_C)

#### Acceptance Criteria:

- [ ] Technically we need to provide a [hugo list template](https://gohugo.io/templates/lists/) at course/layouts/resources/list.html (Thanks @gumaerc )
- [ ] [Group](https://gohugo.io/templates/lists/#by-parameter) resources by "Learning resource type"
- [ ] Within a group order them alphabetically
- [ ] Show only a [limited number](https://gohugo.io/templates/lists/#filtering-and-limiting-lists) of resources for each type (e.g. 10)
- [ ] If more than the limit, provide a "See all" link that links to a page that lists all the resources of that type 
- [ ] Separate link (button) do download the resource.

#### Out of Scope

- [ ] download all 
- [ ] if there is a `/resources/` page created, add it to the sitemap (i.e. revert part of #698)

PR #753, opened 2022-06-15
feat: lists of course resources grouped by learning resource types

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/751

#### What's this PR do?
- [x] List course resources at `base-url/resources/`
- [x] `learning_resource_types` taxonomy used, introduced in `ocw-hugo-projects` in this PR https://github.com/mitodl/ocw-hugo-projects/pull/197
- [x] Resources being shown are grouped by `Learning resource type`
- [x] Within a group, ordered alphabetically
- [x] Limited number of resources shown. for now limit is 10.
- [x] If more than the limit, "See all" link is shown that links to a page that lists all the resources of that type. 
- [x] Separate link (button) do download the resource.
- [x] Clicking on resource takes us to the resource page.

- If this PR https://github.com/mitodl/ocw-hugo-projects/pull/197 is merged then take a pull of `ocw-hugo-projects` locally. If not merged then checkout to its branch.
- Checkout to the branch of this PR.
- Build any course locally.
- Go to `http://localhost:3000/resources/`
- Verify that all the resources are being rendered properly.
- Verify that resources are shown according to `learning resource type`
- Verify that resources in each group are sorted alphabetically.
- Verify that upon clicking of each resource, we are redirected to its page.
- Verify that all the resources are downloadable.
- Verify that the correct thumbnail is shown.
- Verify that only/upto 10 resources are being shown at max.
- Verify that for more resources than 10, a `See All` link is being shown.
- Verify that we are able to click that link and view all the resources of that particular type.
- Any other use-case that needs to be tesed. 

#### Screenshots

URL: `http://*base-url*/resources/`
image

image

URL: `http://*base-url*/learning_resource_types/lecture-videos/`
image

URL: `http://*base-url*/learning_resource_types/`
image


course/layouts/resources/list.html
```diff
@@ -0,0 +1,61 @@
+{{ define "main" }}
+{{- $numberOfResourcesLimit := 10 -}}
+{{ partial "course_content.html" . }}
+
+<div class="mb-5">
+  {{ range  .Pages.GroupByParam "resourcetype"  }}
```
MAbdurrehman12, line 6: **Note:** Currently, to get the UI and other functionalities sorted, resources have been groupedBy `resourceType` due to [this](https://github.com/mitodl/hq/discussions/92#discussioncomment-2954821). Once this issue is sorted, then we will group resources by `learning_resource_types`

MAbdurrehman12:
**Note for the reviewer:** 
I have the following things on my mind which might need to be handled, but I think we can wait for your feedback: 

1. Show a message for any course with no resources. For now a blank page is shown.
2. Upon clicking `See All`, the URL is `/learning_resource_types/type/`, not  `/resources/type/`. Not sure if we do need to cater to this.
3. `/learning_resource_types/` shows/lists learning resource types, which I'm not sure is needed or not. 
e,g: 
image

but if its needed then we need to correct the thumbnail and hide download button.
4. Currently, we are using `learning resource type` taxonomy in resources layout. Not sure if its really a good idea as we are restricting resources list layout to `learning resource type` taxonomy, not sure if it will be needed differently in future.

MAbdurrehman12:
I think I missed the hover state for resource title, will also add it shortly

gumaerc:
@MAbdurrehman12 can you rebase this?

course/layouts/learning_resource_types/list.html
```diff
@@ -0,0 +1,10 @@
+{{ define "main" }}
```
gumaerc, line 1: FYI, this layout is actually not necessary if the desired functionality is to have the list page at `/resources` and in fact this generates an extra set of pages at `/learning_resource_types` that looks like this:

image

I neglected to realize and mention this before you merged https://github.com/mitodl/ocw-hugo-projects/pull/197, but we might want to disable XML output for the `learning_resource_types` taxonomy as well.  The setting you added in that PR will generate "pages" at `site.Pages` for the `learning_resource_types` taxonomy and iterating them manually in the list template at `course/layouts/resources/list.html` will still work.  I tested this locally by removing this layout and regenerating my site.  We also need to remove the default `categories` and `tags` taxonomies from the Hugo config, so I can include removing the XML output for `learning_resource_types` in that PR, but for now we don't need a list template here as it will generate extra HTML.  We have to be careful about unintentionally generating pages now because all of them are inserted into the sitemap and subsequently indexed by Google.

gumaerc (changes requested):
This looks good and the list page generated at `/resources` works as expected.  There's just one unintended side effect caused by a layout you added here:

Ferdi:
> Upon clicking See All, the URL is /learning_resource_types/type/, not /resources/type/. Not sure if we do need to cater to this.

It would be great to have a slightly better URL there,  but it's fine for now.  We can follow up with another PR if needed

> /learning_resource_types/ shows/lists learning resource types, which I'm not sure is needed or not.
e,g:

I suggest we leave the page up for now.  Can you remove the icons ?

course/layouts/learning_resource_types/list.html
gumaerc: Yea, I'll create an issue in `ocw-hugo-projects` to deal with the extra XML files being generated by the taxonomies.  In the meantime, @MAbdurrehman12 , like @Ferdi mentions below we should actually leave the `course/layouts/learning_resource_types/list.html` template because I misunderstood how you were using it and obviously it's central to how this whole thing functions.  Without it, the list pages for each learning resource type won't be generated.  However, we should clean up the title on that page so that it doesn't have underscores.  Instead of calling `course_content.html` you might want to create a new partial, something similar to `content_header.html` (maybe `taxonomy_header.html`?) that takes titles automatically inferred from types like that and strips out the underscores and title cases everything so it looks nice.

As far as the resulting base URL of the taxonomy lists (`/learning_resource_types`) I think we can use the `permalinks` section of the Hugo config for this as seen here: https://github.com/gohugoio/hugo/issues/1208#issuecomment-859222720.  I was able to get the taxonomy lists to be generated at `/resources/exams` etc. by adding this to the Hugo config:

```
permalinks:
  learning_resource_types: "/resources/:slug/"
```

The `/learning_resource_types/index.html` page is still generated for some reason, but that's not a big deal as long as we can fix the title and remove the icons, as Ferdi mentioned.

base-theme/layouts/partials/taxonomy_header.html
```diff
@@ -0,0 +1,8 @@
+<div class="rounded mb-4">
+  <header>
+    <div class="course-section-title-container">
+      <h1 class="text-uppercase font-weight-bold pb-1 mb-3">{{ replace .Title "_" " " }}</h1>
```
MAbdurrehman12, line 4: @gumaerc We can also replace `_` with `" "` in `.Title` in `content_header.html` file but I was unsure whether we have deliberately added underscores (`_`) in any of the page Titles or not. Hence created this new partial.

But if we are sure that we don't want underscores (`_`) in page title then my suggestion would be to delete this partial and do `{{ replace .Title "_" " " }}` in `content_header.html` for better consistency. 

Perhaps if @pdpinch has any thoughts on having underscores (`_`) in page title?

course/layouts/learning_resource_types/list.html
MAbdurrehman12: Thanks for the review guys.

@gumaerc Now we have to create a PR in `hugo-projects` to 
- disable XML for `learning_resource_type` taxonomy
- disable  Hugo default taxonomies (`categories` and `tags`)
- add the config to get the taxonomy lists generated at `/resources/*type*`.

Is this correct?
If yes, then I can go on create the PR.

In the meanwhile, I think you can review this PR again and if all is good then we can merge it to get it to RC, if there's any issue, let me know.

Thanks

gumaerc:
@MAbdurrehman12 Go ahead and filter the pages on `content_type: resource` in the `learning_resource_types` list layout.   I also somehow didn't see your comment above detailing basically all the issues I mentioned, so sorry about that.  I still think we should also add the `permalinks` section in `ocw-hugo-projects` to have the list pages generated at `/resources` like I detailed above.

base-theme/layouts/partials/taxonomy_header.html
gumaerc: I would go with the separate partial for now, although I do agree that intentional underscores in a page title does seem unlikely.  Instead of simply using `replace` though, you might try using `humanize`: https://gohugo.io/functions/humanize/

Ferdi: Can we have a proper title "Learning resource types"   https://gohugo.io/content-management/taxonomies/#add-custom-metadata-to-a-taxonomy-or-term ?

course/layouts/learning_resource_types/list.html
gumaerc: @MAbdurrehman12 :

The XML output from Hugo comes from the "RSS" output format built into Hugo.  This is already disabled for the `home`, `page` and `section` kinds.  If you look at the `output` section in either of the `config.yaml` files, you'll see that "RSS" is not included as an output format.  The default configuration for outputs includes "HTML" and "RSS," and when you redefine this if you don't explicitly specify those they aren't included.  So, I think we'd just need to modify the `outputs` section like so:
 
```
outputs:
  home:
    - coursedata
    - contentmap
    - HTML
  page:
    - coursedata
    - HTML
  section:
    - coursedata
    - HTML
  taxonomy:
    - HTML
  term:
    - HTML
```

I talked with @Ferdi about RSS support and it is desirable, but I think we want to be more deliberate about what kinds of RSS feeds we provide, i.e. a feed of new / updated courses or something.  The automatic RSS feeds generated by Hugo would just list every piece of content available, which isn't super useful.

In the same vein, it seems that the default taxonomies are actually already disabled.  See here for more info: https://gohugo.io/content-management/taxonomies/#example-removing-default-taxonomies.  By having this in the config:

```
taxonomies:
  learning_resource_type: learning_resource_types
```

... we are already telling Hugo that these are the only taxonomies it should be creating.  I verified this locally by looking at the `public` output folder when running a course.  The `tags` and `categories` folders are not created.  They still exist in the live site bucket, and this is because the folders are not being cleared out of old content with each build.  For `ocw-www`, we might want to add:

```
disableKinds:
- taxonomy
- term
```

...until we have taxonomies that we want generated by `ocw-www`.  As far as the `permalinks` section I mentioned above, we should go ahead and do that because it simply instruct Hugo that for the `learning_resource_types` taxonomy, it should generate pages at `/resources`.

base-theme/layouts/partials/taxonomy_header.html
gumaerc: @Ferdi has a good point.  We can add "stock content" to any theme in the same way you would add content to a Hugo site, under the `content` folder.  So, we could in theory add a file at `course/content/learning_resource_types/_index.md` and set the `title` property in the front matter there.

MAbdurrehman12:
> @MAbdurrehman12 A few things I found:
> 
> * I was able to get the root `learning_resource_types` page to render in the right spot by setting `url: /resources/index.html` in the front matter of `course/content/learning_resource_types/_index.md`.  See Hugo URL management documentation here: https://gohugo.io/content-management/urls/
> * Rather than defining a `list.html` in `layouts/learning_resource_types`, maybe we should define 2 layouts: `taxonomy.html` and `terms.html`.  This way we can have separate templates for the root (the taxonomy) and the terms below it.  See Hugo lookup order for taxonomies here: https://gohugo.io/templates/lookup-order/#examples-layout-lookup-for-taxonomy-pages

Okay, I'll try it out!

MAbdurrehman12:
@gumaerc
> I was able to get the root learning_resource_types page to render in the right spot by setting url: /resources/index.html in the front matter of course/content/learning_resource_types/_index.md

This doesn't seem to be working for me. I've been trying to do so but not sure where it's going wrong.
However, if we add 
```
permalinks:
  learning_resource_types: "/resources/:slug/"
```
in the config, it works nicely!

Other than this, I have added the taxonomy templates (this was a good suggestion, Thank you.) and refactored the code a bit to minimize duplications and increase reusability.
The PR is up for another review. 
and also let me know if you're able to get `url: /resources/index.html` working.

Thanks

MAbdurrehman12:
@pdpinch Have we got any `image thumbnail` from the designer? 
We have several other thumbnails:
e.g: `video thumbnail`
image

 but I can't find one for image so currently, I'm using `file thumbnail` for images too

gumaerc (changes requested):
Okay, so after a lot more testing I think I have discovered the best way to handle this.  First of all, I discovered a potential issue with Hugo that still exists where the `taxonomy.html` and `terms.html` layouts seem to be "reversed."  More info here: https://discourse.gohugo.io/t/taxonomy-list-page-takes-a-wrong-template/20189.  Hugo seems to be selecting `terms.html` for taxonomies, and `taxonomy.html` for terms.  Here is what I did to get this to work:

First of all, we need to **remove** the template at `course/layouts/resources/list.html`.  This layout will take precedence over any pages that we are telling to render here with the `url` directive as long as it is present, so we need to remove it.

`course/content/learning_resource_types/_index.md`:

```
---
title: Resources
url: /resources/
kind: taxonomy
---
```

It seems Hugo actually doesn't want the `index.html` at the end when specifying `url`, and adding `kind: taxonomy` forces Hugo to use the `taxonomy` template for this page (by default it selects `section` as the kind) which will actually end up being `terms.html` due to the bug I posted above.  We title it "Resources" because it will end up at `/resources/index.html`.

`course/layouts/learning_resource_types/taxonomy.html`:

```
{{/* This is actually the "terms.html" template, more info here: https://discourse.gohugo.io/t/taxonomy-list-page-takes-a-wrong-template/20189 */}}
{{ define "main" }}
{{ partial "course_content.html" . }}

<div class="mb-5">
  {{- partial "resource_list.html" (dict "resources" .Pages "hideThumbnail" false "hideDownloadIcon" false) -}}
</div>
{{ end }}
```

`course/layouts/learning_resource_types/terms.html`:

```
{{/* This is actually the "taxonomy.html" template, more info here: https://discourse.gohugo.io/t/taxonomy-list-page-takes-a-wrong-template/20189 */}}
{{ define "main" }}
{{ partial "course_content.html" . }}

{{- $numberOfResourcesLimit := 10 -}}
{{ $taxonomy := "learning_resource_types" }}

<div class="mb-5">
  {{ with ($.Site.GetPage (printf "/%s" $taxonomy)) }}
    {{ with .Pages }}
      {{ range . }}

        {{ $resources := where .Pages "Params.content_type" "resource" }}
        {{ $numberOfResources := len $resources  }}
        {{ $limitedResourcesShown := false }}

        {{ if gt $numberOfResources $numberOfResourcesLimit }}
          {{ $numberOfResources = $numberOfResourcesLimit }}
          {{ $limitedResourcesShown = true }}
        {{ end }}

        <div>
          <h4 class="my-4">{{ .Title }}</h4>  
        </div>

        {{- partial "resource_list.html" (dict "resources" $resources "limitResources" $numberOfResources) -}}
          
        {{ if eq $limitedResourcesShown true }}
          <div>
            <div class="float-right">
              {{- partial "see_all.html" (dict "permalink" .Permalink) -}}
            </div>
          </div>
        {{ end }}
        <br>

      {{ end }}
    {{ else }}
      {{- partial "no_resources_found.html" -}}
    {{ end }}
  {{ end }}
</div>
{{ end }}

```

`course/layouts/learning_resource_types/taxonomy.html`:

```
{{/* This is actually the "terms.html" template, more info here: https://discourse.gohugo.io/t/taxonomy-list-page-takes-a-wrong-template/20189 */}}
{{ define "main" }}
{{ partial "course_content.html" . }}

<div class="mb-5">
  {{- partial "resource_list.html" (dict "resources" .Pages "hideThumbnail" false "hideDownloadIcon" false) -}}
</div>
{{ end }}
```

This solution prevents any pages from being rendered directly at `/learning_resource_types`, so the `hideThumbnail` and `hideDownloadIcon` arguments may not even be necessary anymore, although it might be good to leave them in for the future in case we decide we need this functionality.

MAbdurrehman12:
> Okay, so after a lot more testing I think I have discovered the best way to handle this.

@gumaerc Great catch and great findings!
Thanks for your thorough testing. I have updated the PR. 
Also fixed an issue related to the download link for videos.

Can you please have another look?

Thanks

MAbdurrehman12:
Now the only thing in my mind is this [comment](https://github.com/mitodl/ocw-hugo-themes/pull/753/#issuecomment-1165488014)

pdpinch:
> I can't find one for image so currently, I'm using this file thumbnail for images too

Let's go with the file thumbnail for now. I suspect we may not want an image section at all.

gumaerc (approved):
:+1: Good to go!

Issue #752, opened 2022-06-14
overlapping timecode in interactive transcript


pdpinch:
### Steps to Reproduce

See https://draft.ocw.mit.edu/courses/18-100a-real-analysis-fall-2020/resources/18100a-lecture-1-multicammp4/

### Expected Behavior

- Time codes shouldn't overlap
- Maybe empty timecodes shoudn't be display at all?

### Screenshot or Screencast

![image (21)](https://user-images.githubusercontent.com/430126/173686598-1a49c4d4-4a76-4dad-a93e-f810afdd4cec.png)

### Excerpt of [webvtt caption file](https://draft.ocw.mit.edu/courses/18-100a-real-analysis-fall-2020/1c3M0aPBTrAxgJPGQ0lbmWtFAHTXsd0P9_transcript.webvtt) 

```
00:00:32.380 --> 00:00:33.630 align:middle line:90%
at some point in their career.

00:00:33.630 --> 00:00:36.280 align:middle line:90%


00:00:36.280 --> 00:00:38.510 align:middle line:84%
So what is the purpose
of this course?

00:00:38.510 --> 00:00:51.770 align:middle line:84%
So this is for 18
100A, Real Analysis.

00:00:51.770 --> 00:00:54.670 align:middle line:84%
So the purpose of this
course is twofold.

00:00:54.670 --> 00:01:03.790 align:middle line:90%


00:01:03.790 --> 00:01:07.520 align:middle line:84%
Really, I think the first
primary purpose of this course

00:01:07.520 --> 00:01:16.685 align:middle line:84%
is to gain experience
with proofs.
```

PR #759, opened 2022-06-20
fix: overlapping timecode

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/752

#### What's this PR do?
- Increases height for those transcript texts which are empty so that the timecode does not overlap.
- **Note:** I'm not sure whether this is a good solution as ideally, we should hide the `transcript line` when the text is empty as stated in the issue ticket but we can not access this parent element via CSS. I also tried accessing the sibling element (i.e: `transcript-timestamp`) and setting `display: none` to it but it also didn't work. 
We can achieve this via JS, but I don't think a custom JS solution only for this thing will be suitable as we will have to iterate thousands of transcript elements, check which are empty, and hide them. 

- Checkout this branch
- Open any transcript page which has missing transcript texts ([e.g](https://draft.ocw.mit.edu/courses/18-100a-real-analysis-fall-2020/resources/18100a-lecture-1-multicammp4/))
- Verify that the timestamps are not overlapping.

#### Screenshots
Screenshot 2022-06-20 at 5 23 05 PM

Screenshot 2022-06-20 at 5 22 53 PM

### DIFF
--- a/course/assets/css/video-transcript.scss
+++ b/course/assets/css/video-transcript.scss
@@ -28,6 +28,10 @@
   margin-left: 95px;
 }
 
+.transcript-text:empty {
+  height: 20px;
+}
+
 .transcript-line:hover {
   background-color: $medium-gray;
 }


MAbdurrehman12:
This PR increases the height for those transcript texts which are empty so that the timecode/timestamp does not overlap.

I'm not sure whether this is a good solution as ideally, we should hide the `transcript line` when the text is empty as stated in the issue ticket but we can not access this parent element via CSS. I also tried accessing the sibling element (i.e: `transcript-timestamp`) and setting `display: none` to it but it also didn't work.
 
We can achieve this via JS, but I don't think a custom JS solution only for this small fix will be suitable as we will have to iterate over thousands of transcript elements, check which are empty, and hide them.

btw I think, the transcript text should not be empty in the first place.

@pdpinch Can you confirm whether this solution can work for us or not? (You can see the attached screenshots above)
If not, then I think we can figure out some way to hide the element completely. @gumaerc Do you have any thoughts on this?

gumaerc (approved):
:+1: LGTM

As @pdpinch said, those spaces are intentional to break between paragraphs in the transcript so they should be included.  In theory you could use something like the CSS `:has` selector here to hide the lines:

```
.transcript-line:has(.transcript-line:empty) {
  display: none;
}
```

...but the `:has` selector [doesn't have wide browser support yet](https://caniuse.com/css-has).  In OCW we generally want to avoid anything that doesn't have wide historical support.  These kinds of selectors have been in the works for a long time though and I've run into a number of situations where they would be great so hopefully they are fully adopted soon.

MAbdurrehman12:
> In theory you could use something like the CSS `:has` selector 

Yeah, this `:has` selector looked pretty nice but I also did proceed with it due to limited browser support.

PR #761, opened 2022-06-22
search API error handling improved

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/736

#### What's this PR do?
- Sends additional data for better debugging when search API fails.
    - Tags added: 
        1. API response status
        2. API response status text
        3. Current URL with query params
    - Level set to "Error"
    - Extras: Search Query Params
- **Note:** I spent some time but was unable to debug the actual cause of why the issue is occurring in the first place. Hence the error message has been improved by which I think we will be in a better position to debug the issue.

- Go to the search page. (If having CORs issue then [run chrome without cors](https://alfilatov.com/posts/run-chrome-without-cors/))
- Test and verify that everything is working as expected.
- Now we want the API to fail, this can be done by any method (give an invalid URL, let the code break from any other method etc)
- Verify that now an error msg is shown to the user and an event is logged in sentry with the following details/info:
    - Event does not have `<unknown>` title.
    - Similar events are grouped/merged into one. 
    - Current URL with query params tag (for quick testing)
    - API response status tag.
    - API response status text tag
    - Level: Error tag
    - Search Query Params
- Any use-case which needs to be tested.

#### Screenshots

image

image

image

### DIFF
--- a/www/assets/js/lib/util.ts
+++ b/www/assets/js/lib/util.ts
@@ -39,7 +39,7 @@ export const sentryCaptureMessage = (
   message: string,
   additionalData: any = {}
 ) => {
-  Sentry.captureException(message, additionalData)
+  Sentry.captureMessage(message, additionalData)
 }
 
 /**
--- a/www/assets/js/lib/util.ts
+++ b/www/assets/js/lib/util.ts
@@ -28,12 +28,18 @@ export const isApiSuccessful = (status: number) => {
   )
 }
 
-export const sentryCaptureException = (excetpion: any) => {
-  Sentry.captureException(excetpion)
+export const sentryCaptureException = (
+  excetpion: any,
+  additionalData: any = {}
+) => {
+  Sentry.captureException(excetpion, additionalData)
 }
 
-export const sentryCaptureMessage = (message: string) => {
-  Sentry.captureException(message)
+export const sentryCaptureMessage = (
+  message: string,
+  additionalData: any = {}
+) => {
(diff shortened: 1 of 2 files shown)


www/assets/js/lib/util.ts
```diff
@@ -28,12 +28,18 @@ export const isApiSuccessful = (status: number) => {
...
+  Sentry.captureException(excetpion, additionalData)
 }
 
-export const sentryCaptureMessage = (message: string) => {
-  Sentry.captureException(message)
+export const sentryCaptureMessage = (
+  message: string,
+  additionalData: any = {}
+) => {
+  Sentry.captureException(message, additionalData)
```
Wassaf-Shahzad, line 42: Shouldn't this be capture message ?

Wassaf-Shahzad (changes requested):

www/assets/js/lib/util.ts
MAbdurrehman12: oops

Wassaf-Shahzad (approved):
LGTM!

MAbdurrehman12:
@pdpinch Do you want to have a look into this PR, in case you have any thoughts/suggestions or should we merge it?

pdpinch (approved):
LGTM, AFAICT 
👍

PR #768, opened 2022-06-24
added deploy configurations for testing

### LINKED ISSUE #764: deploy a test course with each PR
As a developer, I'd like to have a course build with each PR, so I can catch any test failures or regressions in the course theme. 

As part of current [deploy workflow](https://github.com/mitodl/ocw-hugo-themes/blob/main/.github/workflows/deploy.yml), we build a test site for ocw-www on netlify. We'd like to add another one that includes a course build. 

#### Acceptance Criteria:

- [ ] every ocw-hugo-themes PR builds a specific course
- [ ] the reviewer can link to course site 

#### Out of Scope

- build more than one course

Wassaf-Shahzad:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Closes #764 

#### What's this PR do?
- Adds configuration to deploy course preview

- Verify on a new PR with course preview

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -46,11 +46,17 @@ jobs:
       - name: Clone ocw-www for content
         run: (cd ..; git clone git@github.mit.edu:ocw-content-rc/ocw-www.git ocw-www)
 
+      - name: Clone ocw-course for content
+        run: (cd ..; git clone git@github.mit.edu:ocw-content-rc/15.s21-january-iap-2014.git ocw-course)
+
       - name: Clone ocw-hugo-projects for config
         run: (cd ..; git clone https://github.com/mitodl/ocw-hugo-projects)
 
       - name: Run webpack build
         run: npm run build:webpack -- --output-path=../ocw-www/public/static/
+
+      - name: Copy webpack build to course
+        run: cp -R ../ocw-www/public/static/ ../ocw-course/public/static/
 
       - name: Run hugo build
         run: (cd ../ocw-www; hugo --config ../ocw-hugo-projects/ocw-www/config.yaml --themesDir ../ocw-hugo-themes)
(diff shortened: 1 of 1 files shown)


Wassaf-Shahzad:
@gumaerc There probably is a better way to to do this I think but this approach was my first go to. Would love some of your reviews.

.github/workflows/deploy.yml
```diff
@@ -74,6 +89,21 @@ jobs:
         env:
```
gumaerc, line 89: Above, in the `www` deploy to Netlify, can we change the `alias` to `ocw-hugo-themes-www-pr-${{ github.event.number }}`, to be more clear about what the deploy is?

.github/workflows/deploy.yml
```diff
@@ -74,6 +89,21 @@ jobs:
...
+      - name: Deploy Course Preview to Netlify
+        uses: nwtgck/actions-netlify@v1.1
+        if: ${{ github.event_name == 'pull_request' }}
+        with:
+          publish-dir: '../ocw-course/public'
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          deploy-message: "Deploy from GitHub Actions"
+          enable-pull-request-comment: true
+          enable-commit-comment: false
+          overwrites-pull-request-comment: true
```
gumaerc, line 102: I'm not sure if there's a way to get multiple "Netlify deploy" links to show up in the checks section.  The `nwtgck/actions-netlify` library seems to only support the one.  However, if we are going to do multiple Netlify deploys per PR we should try setting `overwrites-pull-request-comment` to `false` so a comment is written in for each deploy.  Currently, you might notice that since the course build / deployment runs second, it overwrites the comment about the `www` build.  Ithink setting this to `false` should give a unique comment for each deploy on the PR.

gumaerc (changes requested):
This is actually pretty good and what I had in mind.  I have a few notes here on naming for the existing `www` related stuff, which I think we should clarify now that we're deploying both the `www` and `course` themes.  I updated the values in the project secrets for `RESOURCE_BASE_URL` and `STATIC_API_BASE_URL` to match the new scheme we're using, as it was using an outdated URL and not working.  There are still some questions here to answer about how we should best use the CI build for automated testing, but that's outside of the scope of this PR obviously.

Issue #770, opened 2022-06-24
Intermittent "No compatible source was found for this media"


pdpinch:
### Steps to Reproduce

1. Visit a video page, like https://ocw.mit.edu/courses/8-01sc-classical-mechanics-fall-2016/resources/9/
2. Clear the cache and reload the page. 
3. It may be necessary to navigate to other videos in this course. 

### Expected Behavior

- no errors & video thumbnail loads
- video plays 

### Actual Behavior

- "No compatible source was found for this media" error 
- No thumbnail
- Video doesn't play 

### Stacktrace

<details>
<summary><b>click to reveal stacktrace</b></summary>

```
course.44902.js:2 VIDEOJS: ERROR: The "youtube" tech is undefined. Skipped browser support check for that tech.
(anonymous) @ course.44902.js:2
r.error @ course.44902.js:2
(anonymous) @ course.44902.js:2
n.selectSource @ course.44902.js:2
n.src_ @ course.44902.js:2
(anonymous) @ course.44902.js:2
e @ course.44902.js:2
e @ course.44902.js:2
(anonymous) @ course.44902.js:2
(anonymous) @ course.44902.js:2
setTimeout (async)
t.setTimeout @ course.44902.js:2
Er @ course.44902.js:2
n.src @ course.44902.js:2
t @ course.44902.js:2
t.addChild @ course.44902.js:2
(anonymous) @ course.44902.js:2
t.initChildren @ course.44902.js:2
t @ course.44902.js:2
xs @ course.44902.js:2
qn @ course.44902.js:2
setTimeout (async)
Wn @ course.44902.js:2
(anonymous) @ course.44902.js:2
n @ course.44902.js:2
(anonymous) @ course.44902.js:2
n @ course.44902.js:2
(anonymous) @ course.44902.js:2
(anonymous) @ course.44902.js:2
n @ course.44902.js:2
(anonymous) @ course.44902.js:2
(anonymous) @ course.44902.js:2
[...]

PR #728, opened 2022-06-06
fix: videojs-youtube lib included inside document.ready function

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-hugo-themes/issues/701 (Not sure if this PR will close it or not as this PR is more of a hit-and-trial. Please see some points below)

#### What's this PR do?
- **Note:** I was unable to reproduce this issue locally, but most likely, this issue is occurring from a third party, i.e: `videojs-youtube` and if this is the case, then I think loading this lib inside `document.ready` function can do the job. Still, I'm not very sure that this will fix the issue so this PR can be considered a hit-and-trial PR. Not sure if this is a good approach/practice.
- Replaces `import` with `require` for `videojs-youtube`
- Moves `require("videojs-youtube")` inside `document.ready` function.
- Declare module `videojs-youtube`

#### How should this be manually tested?
- Try to reproduce the issue (if you succeed in that, then I owe you a treat).
- Checkout this branch
- Verify that the changes done in this PR resolves the issue.
- If you're unable to reproduce it locally, then: 
    - Verify that this issue is occurring from a third party, i.e: `videojs-youtube` and moving its import in ready function can fix the issue. 
    - Verify that everything is working fine and the changes done in this PR do not have any side-effect.

PR #776, opened 2022-06-28
fix: import order of videojs-youtube sorted

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/770

#### What's this PR do?
- Moves `require("videojs-youtube")` outside the load function 

- Open any course which has videos and open any video page.
- Check if you have this error `No compatible source was found for this media.`
    - If not, then you can reproduce this error by delaying the loading of `videojs-youtube` (maybe call the require in a [settimeout](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout) after 1 second or so)
    - Now the issue is reproduced, I think (but not very sure) it is occurring intermittently due to the requiring of package here in this load function along with the calling of all those functions which uses this as a dependency.
- Checkout this branch
- Verify that the error does not occur again.
- Verify that all the functionalities of video work as before (downloading video, transcript) etc etc
- Repeat this after clearing cache or using incognito mode.
- Repeat this in multiple browsers.
- Repeat the above for multiple videos and courses.

- This issue seems to be a side-effect of https://github.com/mitodl/ocw-hugo-themes/pull/728

### DIFF
--- a/course/assets/course.js
+++ b/course/assets/course.js
@@ -18,9 +18,9 @@ import {
   checkAnswer,
   showSolution
 } from "./js/quiz_multiple_choice"
+import "videojs-youtube"
 
 $(function() {
-  require("videojs-youtube")
   initPdfViewers()
   initDesktopCourseInfoToggle()
   initCourseInfoExpander(document)


Wassaf-Shahzad:
LGTM!

PR #777, opened 2022-06-28
course resources revisions

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

No Ticket :) but this stuff was discussed in this [thread](https://github.com/mitodl/hq/discussions/92#discussioncomment-3028326)

#### What's this PR do?
- Switches taxonomy templates (after renaming `terms.html` to `term.html`). For context: [View this](https://discourse.gohugo.io/t/taxonomy-html-and-terms-html-reversed/39234). All thanks to @gumaerc!
- Adjusts resource title so it looks fine on smaller screens
- Hides/Does not render the learning resource type which doesn't have any resource
- Checks if all learning resource types have no resources then print a single message "No Resources found."
- Some more little tweaks.

- Checkout this branch.
- Verify all the above-mentioned points.
- Verify that all things work as expected. (Some testing instructions are mentioned in the description of this [PR](https://github.com/mitodl/ocw-hugo-themes/pull/753))
- Any use-case that needs to be tested.

#### Screenshots
image

image


course/layouts/learning_resource_types/taxonomy.html
```diff
@@ -1,9 +1,51 @@
-{{/* This is actually the "terms.html" template, more info here: https://discourse.gohugo.io/t/taxonomy-list-page-takes-a-wrong-template/20189 */}}
 {{ define "main" }}
 {{ partial "course_content.html" . }}
 
+{{- $numberOfResourcesLimit := 10 -}}
+{{- $showNoResourcesFoundMessage := true -}}
```
gumaerc, line 5: We don't need this extra logic here with this `$showNoResourcesFoundMessage` flag.  All you need to do here is use an `{{ else }}` statement on the `{{ with ($.Site.GetPage (printf "/%s" $taxonomy)) }}`.  If the `with` statement evaluates as false or nil, it will fall back to the else statement.

gumaerc: Okay, I see now that you're making that assessment based on if there are any pages with `content_type` equal to `resource` under the taxonomy and that makes sense.

gumaerc:
@MAbdurrehman12 @pdpinch @Ferdi 

A major issue I'm noticing here with our strategy is that if `learning_resource_types` is not set at all, the resource will not be included in this list.  A large amount of courses in `mitocwcontent` do not have this set on their resources.  Since we are using this taxonomy to generate these pages, if the taxonomy is not set then Hugo will not include the resource when compiling it.  Do we want to maybe include an "Other" category on the taxonomy template itself, querying specifically for pages that have `learning_resource_types` empty or null?

gumaerc:
@MAbdurrehman12 Also, this isn't something that needs to be addressed in this PR but in general there seems to be some confusion surrounding the hyphen delimiters used in the syntax here i.e. `{{- if etc... -}}`.  The purpose of the hyphen delimiters is to trim white space surrounding the tag.  Mainly this should be used when inserting template syntax directly into something like an href property where you want to be certain there are no white space or line breaks i.e. `<a href="{{- partial "some_partial.html" . -}}">`.  I don't think there's really any harm in using it outside of that use case, it just looks a bit confusing.  For instance:

```
{{- $numberOfResourcesLimit := 10 -}}
{{- $showNoResourcesFoundMessage := true -}}
{{ $taxonomy := "learning_resource_types" }}
```

Someone looking at this might ask themselves why the hyphen syntax is on 2 of these variable assignments but not on the 3rd one.  There are some examples [here](https://gohugo.io/templates/introduction/#whitespace) but the gist of it is that if you do something like:

```
<div>
  {{ .Title }}
</div>
```

...then you get:
```
<div>
  The Title
</div>
```

...but conversely if you do:
```
<div>
  {{- .Title -}}
</div>
```

...then on the output you will get:
```
<div>The Title</div>
```

Again, this isn't really a big concern as really the only issue is some potential confusion as to why some stuff is being trimmed and some isn't as well as some messy HTML output.  I think we should probably look into using the `--minify` option on our HTML output anyway to reduce file size and load times.

gumaerc (approved):
:+1: 

I'm approving this PR, but I think we need to think about what I stated above and potentially make another follow up to address listing resources that don't have a `learning_resource_types` value set.

MAbdurrehman12:
> seems to be some confusion surrounding the hyphen delimiters used in the syntax here i.e. {{- if etc... -}}

@gumaerc Yes, I think we should maintain consistency here. I'll go on and update the PR and exclude these unnecessary hyphen delimiters to improve consistency and avoid any confusion in the future. 

Your suggestion is to use `--minify` instead of `hyphen delimiters`? (I was not sure that it improves the load time, if thats the case then we should definitely go with it)

Issue #762, opened 2022-06-22
PDFs have no preview when embedded on ocw-www


dseaton:
### Steps to Reproduce
On ocw-www, embedded pdfs in the "[why give](https://draft.ocw.mit.edu/pages/why-give/)" section do not have pdf previews after clicking embedded resource links. The following three links lead to resource preview pages, but none of the pdfs have a preview:

* [read a compilation of some of the inspirational feedback](https://draft.ocw.mit.edu/resources/20_years_of_ocw_impactv2/)
* [OCW Book of Impact](https://draft.ocw.mit.edu/resources/20_years_of_ocw_impactv2/)
* [read our latest impact report](https://draft.ocw.mit.edu/resources/2020-19_ocw_impact_report/)

### Expected Behavior
Each pdf should open the pdf in a preview mode.

### Actual Behavior
See screenshot. Each pdf only shows the title and a download button.

### Screenshot or Screencast
![image](https://user-images.githubusercontent.com/1703442/175042256-140ba2f0-082e-4350-bc75-8f601e96e68a.png)

PR #780, opened 2022-06-28
replace pdf.js with pdfobject

gumaerc:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/762
Closes https://github.com/mitodl/ocw-hugo-themes/issues/788

#### What's this PR do?
 - Removes [PDF.js](https://mozilla.github.io/pdf.js/) from the project
 - Adds [PDFObject](https://pdfobject.com/) to the project
 - Moves all code regarding the PDF viewer to the base theme so it can be used anywhere

 - Do a fresh install of the Node dependencies:
   - `rm -rf node_modules`
   - `yarn install`
 - Clone any course from `mitocwcontent` that contains a PDF.  For example, `18.01sc-fall-2010`.  Also, clone `ocw-www`
 - Put the following in your `.env`:
```
COURSE_CONTENT_PATH=/path/to/mitocwcontent/
WWW_CONTENT_PATH=/path/to/mitocwcontent/ocw-www/
OCW_TEST_COURSE=<your test course>
```
 - Spin up the test course with `npm run start:course`
 - Browse to http://localhost:3000/ and find a PDF in your test course (you could also search the `resources` folder in your test course markdown for resources with `resourcetype: Document`)
 - Make sure that the PDF displays properly in every browser you have access to, or doesn't display at all if the browser is not supported.  For example, most mobile browsers will not be supported, and you will only get a download link.
 - Stop your dev server and browse to your `mitocwcontent` folder.  Find a markdown resource file representing a PDF in your course and copy it into the `content/resources` directory in `ocw-www`
 - Spin up `ocw-www` with `npm run start:www`
 - Browse to http://localhost:3000/resources/<your resource file here>
 - Test the PDF in the same manner as above

Way back in https://github.com/mitodl/hugo-course-publisher/pull/105, we added the PDF.js library to the project with the goal of having a universal way to embed PDFs on pages that worked regardless of the platform.  There were a number of challenges surrounding this.  Mozilla, who maintains PDF.js, offers the open source `pdfjs` Node package on npm, but it does not offer a way to simply load the "example" viewer in your project.  Instead, what they would like people to do is use their library to build their own PDF viewer.

Not being in the business of making PDF viewers, we worked around this by adding a build step to our `package.json` file that would build PDF.js from source and include the example viewer in our project.  Later on, we reduced complexity by removing this build step and simply including a statically built version of the viewer with the course theme.  There were still a number of issues with this:

 - The fully built PDF.js viewer contains a lot of files that need to be copied up to S3 with every theme assets build
 - Cross-domain PDFs could still not be embedded unless the calling domain was allowed via CORS
 - Embedding PDFs in other themes such as `www` would involve moving it to the base theme 
 - PDF.js is a whole other application dependency that needs to be maintained

Enter PDFObject, a simple library consisting of a single Javascript file whos entire job is simply to asses your browser's capabilities and use one of the standard methods for embedding a PDF like `iframe`, `object` or `embed` using a set of rules about browser quirks  that they maintain.  We didn't use PDFObject initially because it was originally desired to have the PDF viewer work even on iOS devices, but those priorities have been reconsidered.  Since PDFObject is a simple JS class with one job, it can easily be used anywhere on the project and requires little to no maintenance.

image
image


pdpinch:
Functionality is great, but desktop safari and iOS safari are having some layout issues: 

Desktop: 
image

iOS:
File

gumaerc:
@Ferdi @pdpinch I updated the PR with some code to hide the PDF embed on Safari specifically.  Currently, this includes desktop Safari.  If you look at Peter's screenshot he posted above, even on desktop Safari the Apple Webkit embedded PDF viewer iterates the pages one by one and has no paging / zoom controls or anything.  I might be able to specifically allow desktop Safari with some custom CSS to handle overflow manually and still hide the viewer on iOS, but all of this is based on user agent sniffing, which can become unreliable over time if the user agent string changes.  The team that works on PDFJS does the heavy lifting of making sure the PDF viewer is consistent across all browsers, which is why it was a nice solution.  There are a lot of downsides to using it as we've discussed, but it does make the things We've been discussing so far in this PR related to Apple devices a non-issue.  In summary, moving away from PDFJS will remove support for embedded PDFs on:

 - all browsers on iPhone (we don't really care about this one since it's not super useful)
 - all browsers on iPad (this one would be more useful because the iPad has a bigger screen, but a native app is still probably better)
 - desktop Safari on macOS

Since all browsers on iOS are forced to use Apple Webkit, Chrome and Firefox on iPhone and iPad will not work.  We have a couple places we can go from here, assuming we still want to remove PDFJS:

 - Merge this PR as-is and remove support for all iOS devices and desktop Safari
 - Write some custom CSS and enable support for only desktop Safari with the understanding that the PDF viewer in Safari will need some custom CSS and will still only iterate the pages of the PDF with no zoom / paging controls

pdpinch:
I'm OK with providing a different experience for Safari across the board, just make it clear that this is a decision we're making and why -- so if someone reports a "bug" a year from know we'll know how we got here.

pt2302:
I don't think that this line is required here: https://github.com/mitodl/ocw-hugo-themes/blob/5bbbed8cb5698f03d69ba1d9fafd3529c0959b35/course/assets/course.ts#L24

gumaerc:
> I don't think that this line is required here:
> 
> https://github.com/mitodl/ocw-hugo-themes/blob/5bbbed8cb5698f03d69ba1d9fafd3529c0959b35/course/assets/course.ts#L24

Good catch!  This likely was accidentally added from vscode auto completion.  I removed it.

pt2302 (approved):
Tested on Chrome, Edge, Firefox, and Safari browsers, as well as Safari iOS. Looks good to me!

Issue #632, opened 2022-04-11
H3 sizing inconsistent


pdpinch:
### Steps to Reproduce

- On https://ocwnext.odl.mit.edu/courses/6-004-computation-structures-spring-2017/pages/instructor-insights/ the H3 is computed to 24px
- On https://ocwnext.odl.mit.edu/courses/24-912-black-matters-introduction-to-black-studies-spring-2017/pages/instructor-insights/ the H3 is computed to 16px 

### Expected Behavior

H3 should be the same size everwhere (not just instructor insights pages) 

### Actual Behavior

There seems to be an style applied on the 2nd example: 
```
#course-content-section h3 {
  font-size: 1rem !important;
}
```

### Related Issues

- #633

Wassaf-Shahzad:
@pdpinch I did some RnD on this and the reason those h3s have a custom font-size is because those set of h3 are generated by hugo and we can't replace them with h4s unless we change it in ocw-studio and I think in ocw-to-hugo as well for legacy courses, Since these are generated from .md files. So we had to override the font-size in styles. I can remove the custom style but It will make those subheading a bit bigger if you like.

### Screenshots
With default h3 styles
<img width="1348" alt="Screenshot 2022-04-12 at 10 20 12 PM" src="https://user-images.githubusercontent.com/87968618/163018739-267b57e3-c822-4367-a655-d518c2f6a6e2.png">

With current overridden h3 styles
<img width="1348" alt="Screenshot 2022-04-12 at 10 20 18 PM" src="https://user-images.githubusercontent.com/87968618/163018904-3d8e7540-809f-4d65-a0cf-eb46fbc9ff32.png">

pdpinch:
We are planning a site-wide CSS update. I'm going to put this on hold until that is ready, and we consider this at the same time.

pdpinch:
Can we fix this in the courses-v2 theme?

MAbdurrehman12:
@Wassaf-Shahzad Since this issue will be fixed by https://github.com/mitodl/ocw-hugo-themes/pull/794 so I'm assigning it to you

PR #784, opened 2022-07-01
feat: compact design

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-hugo-themes/issues/765

#### What's this PR do?
- Creates a new theme `course-v2` and implements new designs as provided in the issue ticket.

#### How should this be manually tested?
- Checkout this branch
- In your local ocw course config (having env: `COURSE_HUGO_CONFIG_PATH`), add/rename `course` to `course-v2` to use this new theme instead of the previous one.
- Build any course locally.
- Verify that the UI is according to the provided designs for the course home page and course interior pages.
- Repeat this for:
    - as many courses as you can :) I'm also currently testing.
    - multiple browsers (chrome, firefox, safari etc)
    - multiple screen sizes (desktop, laptop, tablet, mobile etc)
- **Important:** Verify that the previous/old design of `course` and `www` still works as expected and there are no side-effects/impacts. 

#### Screenshots
- Please see this [deployed course](https://ocw-hugo-themes-course-pr-784--ocw-next.netlify.app/).

PR #794, opened 2022-07-15
updated headers

### LINKED ISSUE #792: Standardize header sizes
#### Designs and Mockups

See https://projects.invisionapp.com/d/main#/console/21537330/468374061/preview

> This does require a dev to change what is a h1, h2, h3, etc, but it’s needed for consistency. For example the H1 on blue is not currently an H1 and the SYLLABUS is marked an h1, but it should be an h2… https://ocwnext-rc.odl.mit.edu/courses/21g-102-chinese-ii-regular-spring-2015/pages/syllabus/

#### Acceptance Criteria:

- [ ] This should be applied to the coursev2 theme, in a follow up PR to #784 
- [ ] test on a course with instructor insights

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

Closes #792 

#### What's this PR do?
- Updates heading tag of title to H1
- Updates heading tag of course-title to h2 

- Verify heading for course-banner and course-title is h1 and h2

I have updates the headings for the course banner and title however course content headings such as these 
Screenshot 2022-07-15 at 6 09 27 PM

are set by users and the markdown for it also is for a h1

Screenshot 2022-07-15 at 5 56 33 PM

### DIFF
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -56,6 +56,10 @@ html {
   }
 }
 
+.font-black {
+  color: black !important;
+}
+
 #course-main-content article.content {
   a:link {
     color: $course-content-link-unclicked;
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -144,7 +144,7 @@ h1 {
 
 h2 {
   font-size: 1.715rem !important; // 24px
-  color: $black;
+  color: $black !important;
 
   @include media-breakpoint-down(xs) {
     font-size: 1.429rem !important; // 16px
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -61,9 +61,6 @@ html {
 }
 
 #course-main-content article.content {
-  h3 {
-    color: $highlight-red;
-  }
 
   a:link {
     color: $course-content-link-unclicked;
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -61,7 +61,6 @@ html {
 }
 
 #course-main-content article.content {
-
   a:link {
     color: $course-content-link-unclicked;
   }
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -138,20 +138,20 @@ pre {
 // not using font variables here as we need different font sizes on course pages
 // these font sizes are yet to be confirmed
 h1 {
-  font-size: 2.28rem !important; // 32px
+  font-size: 2.286rem !important; // 32px
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 1.7rem !important; // 24px
+    font-size: 1.715rem !important; // 24px
   }
 }
 
 h2 {
-  font-size: 1.7rem !important; // 24px
+  font-size: 1.715rem !important; // 24px
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 1.4rem !important; // 16px
+    font-size: 1.429rem !important; // 16px
   }
 
   &.branded {
@@ -185,16 +185,16 @@ h4 {
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 0.86rem !important; // 12px
+    font-size: 1rem !important; // 12px
   }
 }
 
 h5 {
-  font-size: 1rem !important; // 14px
+  font-size: 0.858rem !important; // 14px
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 0.86rem !important; // 12px
+    font-size: 0.858rem !important; // 12px
   }
 }
 
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -61,6 +61,11 @@ html {
 }
 
 #course-main-content article.content {
+
+  h3 {
+    color: $highlight-red;
+  }
+
   a:link {
     color: $course-content-link-unclicked;
   }
@@ -134,20 +139,20 @@ pre {
 // not using font variables here as we need different font sizes on course pages
 // these font sizes are yet to be confirmed
 h1 {
-  font-size: 1.43rem !important; // 20px
+  font-size: 2.28rem !important; // 32px
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 1.23rem !important; // 18px
+    font-size: 1.7rem !important; // 24px
   }
 }
 
 h2 {
-  font-size: 1.14rem !important; // 16px
+  font-size: 1.7rem !important; // 24px
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 1.14rem !important; // 16px
+    font-size: 1.4rem !important; // 16px
   }
 
   &.branded {
@@ -162,7 +167,7 @@ h2 {
 }
 
 h3 {
-  font-size: 1rem !important; // 14px
+  font-size: 1.143rem !important; // 16px
   color: $black;
 }
 
@@ -172,7 +177,7 @@ a + h3 {
 
 @include media-breakpoint-down(xs) {
   h3 {
-    font-size: 1rem !important; // 14px
+    font-size: 1.143rem !important; // 16px
   }
 }
 
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -61,7 +61,6 @@ html {
 }
 
 #course-main-content article.content {
-
   h3 {
     color: $highlight-red;
   }
--- a/course-v2/layouts/pages/_markup/render-heading.html
+++ b/course-v2/layouts/pages/_markup/render-heading.html
@@ -1,0 +1,5 @@
+{{ if eq .Level 2}}
+    <h{{ add .Level 1}} id="{{ .Anchor | safeURL }}">{{ .Text | safeHTML }}</h{{ add .Level 1 }}>
+{{ else }}
+    <h{{ .Level }} id="{{ .Anchor | safeURL }}">{{ .Text | safeHTML }}</h{{ .Level }}>
+{{ end }}
--- a/course-v2/layouts/pages/_markup/render-heading.html
+++ b/course-v2/layouts/pages/_markup/render-heading.html
@@ -1,4 +1,4 @@
-{{ if eq .Level 2}}
+{{ if and (eq .Level 2) (.Anchor) }}
     <h{{ add .Level 1}} id="{{ .Anchor | safeURL }}">{{ .Text | safeHTML }}</h{{ add .Level 1 }}>
 {{ else }}
     <h{{ .Level }} id="{{ .Anchor | safeURL }}">{{ .Text | safeHTML }}</h{{ .Level }}>
--- a/course/assets/css/instructor-insights.scss
+++ b/course/assets/css/instructor-insights.scss
@@ -1,5 +1,13 @@
 @import "../../../base-theme/assets/css/variables.scss";
 #course-content-section {
+  h2 {
+    margin-bottom: 1.25rem !important;
+  }
+
+  h3 {
+    font-size: $font-normal !important;
+  }
+
   p {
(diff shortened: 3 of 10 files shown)


pdpinch:
Oh, I thought Abdurrehman was going to work on this. In any case, this change should be part of his course-v2 theme, so we can look at it with everything else. Can you move it over there, maybe after he merges #784 ?

pdpinch:
> I have updated the headings for the course banner and title however course content headings such as these
> are set by users and the markdown for it also is for a h1

Oh, that's going to be challenging. Is there any way to transform that in the theme, with hugo?

Wassaf-Shahzad:
> > I have updated the headings for the course banner and title however course content headings such as these
> > are set by users and the markdown for it also is for a h1
> 
> Oh, that's going to be challenging. Is there any way to transform that in the theme, with hugo?

Yes, we can hugo render hooks to update the level for headings in a specific theme. Render hooks for heading were introduced in v0.71

pdpinch:
Why is this blocked?

Wassaf-Shahzad:
> Why is this blocked?

This was blocked by compact design changes

gumaerc (changes requested):
I think that the scope here is supposed to be limited to the `course-v2` theme and we shouldn't be making any changes to the current `course` theme here.  @pdpinch Can you verify?

The reason the issue asks to check Instructor Insights pages is because way back in https://github.com/mitodl/ocw-course-hugo-theme/pull/20 we added a separate bundle for instructor insights pages so that they could have their own isolated CSS bundle that's only applied on Instructor Insights pages.  The invision designs for these pages back then had different font sizes / colors than other pages.  Take a look at `instructor-insights.scss` for these styles, and verify that the styles here match the typography guide.  Thankfully, it looks like whoever came up with that typography guide also gave font sizes in `rem` assuming a base font size of 16px.

gumaerc:
Also @Wassaf-Shahzad this could use a rebase now that https://github.com/mitodl/ocw-hugo-themes/pull/807 is merged

course-v2/layouts/pages/_markup/render-heading.html
```diff
@@ -0,0 +1,5 @@
+{{ if eq .Level 2}}
+    <h{{ add .Level 1}} id="{{ .Anchor | safeURL }}">{{ .Text | safeHTML }}</h{{ add .Level 1 }}>
+{{ else }}
+    <h{{ .Level }} id="{{ .Anchor | safeURL }}">{{ .Text | safeHTML }}</h{{ .Level }}>
+{{ end }}
```
MAbdurrehman12, line 5: This looks fine but I have a concern/question that will it cause confusion for course authors? Like the course author has set `h2` from studio but it's being rendered as `h3` in Hugo?

Another follow-up question: Is this the correct place to cater to this problem? what if we handle it in studio?
Perhaps if @pdpinch has any thoughts on this?

MAbdurrehman12 (changes requested):
Overall it looks fine, since correct headings are being rendered at correct places.
Just a few pointers mentioned below.

Along with this, an important thing here will be to define/re-define font-sizes for different headings (h1, h2, h3 etc) for this `course-v2` theme. Because currently h3, h4, h5 are just 14px and the values/font-sizes of h1, h2, h3 are also incorrect.

I defined these font-sizes for headings in `course-v2.scss` but that needs to be updated as mentioned [here](https://github.com/mitodl/ocw-hugo-themes/blob/270769d6a3c4ce56dea33cc1b4efdaf552c16e70/course-v2/assets/css/course-v2.scss#L131) (btw, once we completely move to `course-v2` theme, then we can move this change from `course-v2.scss` to some more appropriate place)

So I think we should do that too in this PR.
font-sizes for headings are provided [here](https://projects.invisionapp.com/d/main#/console/21537330/468374061/preview), but the issue which I'm seeing is: `16px` has been considered as the `root font size`, where as, for `course-v2` theme, `root font size` is `14px`.

So as mentioned [here](https://projects.invisionapp.com/d/main#/console/21537330/468374061/preview), for heading 1: `2rem` will not be equal to `32px` for `course-v2` theme. `2rem` will be equal to `28px` because `1rem` = `root font size` i.e: `14px`.

Maybe we can get these font-sizes updated? 
CC: @pdpinch

course-v2/assets/css/course-v2.scss
```diff
@@ -130,20 +138,20 @@ pre {
...
 }
 
 h2 {
-  font-size: 1.14rem !important; // 16px
+  font-size: 1.7rem !important; // 24px
   color: $black;
 
   @include media-breakpoint-down(xs) {
-    font-size: 1.14rem !important; // 16px
+    font-size: 1.4rem !important; // 16px
```
MAbdurrehman12, line 154:
```suggestion
    font-size: 1.4rem !important; // 20px
```

MAbdurrehman12 (changes requested):
Looks good!
However, I think we should:
1. Also define font-sizes for `h4` and `h5` as defined in the [design](https://projects.invisionapp.com/d/main#/console/21537330/468374061/preview).
2. Go upto 3 decimal places for rem values because currently, with 2 decimal places, px values are not equal to what we are trying to set.
for example: 
In case of `h1`: `2.28rem` = `31.92px` (2.28x14) but our target value is `32px`. 
and if we do `2.286rem`, it will be equal to `32px`
Though it won't make any noticeable difference but maybe we can do it just for the sake of accuracy.
What do you think?
Sorry, if I'm being a bit paranoid :P

MAbdurrehman12 (approved):
LGTM 👍

gumaerc (changes requested):
The changes look properly scoped to `course-v2` now but I think we need to look at one last thing which would be instructor insights pages. They have their own style sheet located at `course-v2/assets/css/instructor-insights.scss`. There is some custom header sizing that we applied in the past for these pages:

```
  h2 {
    margin-bottom: 1.25rem !important;
  }

  h3 {
    font-size: $font-normal !important;
  }
```
According to @pdpinch 's comments on the [issue](https://github.com/mitodl/ocw-hugo-themes/issues/792) we may want to take a look at these pages and make sure that the headers are standardized. An example of a course with instructor insights would be `18.s096-fall-2015`. Here is what the instructor insights page looks like on this branch:

image

We should just make sure that the headers look like they're supposed to in this case before we merge this. The rest of the changes look good and are properly scoped to `course-v2`.

MAbdurrehman12:
@gumaerc I think we should remove this styling
```
  h2 {
    margin-bottom: 1.25rem !important;
  }

  h3 {
    font-size: $font-normal !important;
  }
```
from `course-v2/assets/css/instructor-insights.scss` and we'll be good to go! Right?

ps:  I think this PR will also fix another issue https://github.com/mitodl/ocw-hugo-themes/issues/632 so we can also close #632 once this PR is merged.

Wassaf-Shahzad:
@MAbdurrehman12 @gumaerc I updated instructors-insights.css However I had to remove update it for both course v1 and v2 due to how we have configured webpack.

pdpinch:
Can you give me some before and after URLs I can look at, to see this change? I'd like to have an example of a regular interior page, and an Instructor Insight page.

pdpinch:
Also, in courses-v2 we should be following the color styles specified in https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/468374061/preview?scrollOffset=60 -- no more red

Wassaf-Shahzad:
> https://ocw-hugo-themes-course-pr-794--ocw-next.netlify.app

@pdpinch Updated heading colors

course-v2/layouts/pages/_markup/render-heading.html
pdpinch: I think I asked for this, but @MAbdurrehman12 makes a good point. What happens if we don't adjust the heading levels? If possible, give me an example of a course in production.

course/assets/css/instructor-insights.scss
```diff
@@ -1,13 +1,5 @@
 @import "../../../base-theme/assets/css/variables.scss";
 #course-content-section {
-  h2 {
-    margin-bottom: 1.25rem !important;
-  }
-
-  h3 {
-    font-size: $font-normal !important;
-  }
-
```
pdpinch, line 10: Can we limit this change to course-v2 ?

pdpinch:
> https://ocw-hugo-themes-course-pr-794--ocw-next.netlify.app/

> @pdpinch Updated heading colors

Sorry, I'm still seeing the red headings in the build at https://ocw-hugo-themes-course-pr-794--ocw-next.netlify.app/ 

Where can I see this in course-v2?

gumaerc:
@pdpinch We'll need to merge https://github.com/mitodl/ocw-hugo-themes/pull/824 before we'll start seeing `course-v2` builds with Github actions / Netlify.

@Wassaf-Shahzad I'm not sure why we need to also remove the changes from the `course` version of `instructor-insights.scss`? I added these lines back in and took a look at the site rendering with `course`:
image

Running again with `course-v2` I get:
image

Can you elaborate on why you believe we need to remove the CSS from both?

MAbdurrehman12:
> Where can I see this in course-v2?

https://ocw-hugo-themes-course-v2-pr-794--ocw-next.netlify.app/

Wassaf-Shahzad:
> @pdpinch We'll need to merge #824 before we'll start seeing `course-v2` builds with Github actions / Netlify.
> 
> @Wassaf-Shahzad I'm not sure why we need to also remove the CSS from the `course` version of `instructor-insights.scss`? I added these lines back in and took a look at the site rendering with `course`: image
> 
> Running again with `course-v2` I get: image
> 
> Can you elaborate on why you believe we need to remove the CSS from both?

@gumaerc i have not idea why it didn’t  work for me before but its working know so i added back the code.

course-v2/layouts/pages/_markup/render-heading.html
gumaerc: @pdpinch Here is the PR preview course in RC with the `course-v2` theme: https://live-qa.ocw.mit.edu/course-v2/courses/15-s21-nuts-and-bolts-of-business-plans-january-iap-2014/

Here is the `course-v2` deploy preview for this PR branch: https://ocw-hugo-themes-course-v2-pr-794--ocw-next.netlify.app/

How font sizes are being set currently:
image

After this PR:
image

The course title header is now wrapped in an h1 tag, but it isn't doing anything to the font size because the link inside of it has a font specified in it via CSS. As far as the content area though, it seems like we're getting h2 and h3 appropriately sized like you'd expect. Why are we demoting h2 to h3 in the above code? If we do end up doing this, there's no need in specifying `{{ add .Level 1}}` because that will always evaluate to 3 in this case, because the if statement is limited to the level being 2.

pdpinch: We've got a couple different requirements, but I can't easily tell if we're meeting them all. 

In order of priority: 

- [ ] every pages should have one and only one H1, for accessibility 
- [ ] additional headers (H2, H3, etc) should follow a logical sequence. In particular, we should avoid skipping any levels. Also for accessibility.
- [ ] Size and color should be correct and consistent, including Instructor Insights pages
- [ ] There should be some logical relationship between the headers users set in Studio (H2, H3, etc) and how they are rendered in the published HTML 
- [ ] We don't want to have to change legacy markdown already in studio. 

Does that help answer your question?

Wassaf-Shahzad: @pdpinch Addressed these checkpoints

dseaton:
If I review this PR based solely on #792, then I think this meets the acceptance criteria. However, there seems to be a lot of other considerations. Could we get @gumaerc to give a final approval in the absence of @pdpinch?

gumaerc (approved):
:+1: 

@dseaton Everything here has been addressed. I only had one concern that we can address in a follow-up if we need to, relating to this thing that @pdpinch said:

> every page should have one and only one H1, for accessibility

While this is true, the H1 element is always the course title. From an accessibility standpoint, wouldn't we want the H1 to be the page title? Maybe I'm incorrect about this, but if this is a point about accessibility I would think that the H1 would be the first element a screen reader dictates to the user, and having that be the same thing every time seems a bit counter-intuitive. Either way, if that's an issue we can address it later. This is now properly scoped and the correct styles are applied where they should be.

dseaton:
Thanks @gumaerc. Let's talk with @pdpinch next week about the H1 issue. @Wassaf-Shahzad you should be good to go now.

Issue #795, opened 2022-07-15
Make the course drawer "sticky"


pdpinch:
As a user, I'd like the course drawer to stay permanently closed (or open), when I change the setting. 

This change should be part of the course-v2 theme

#### Acceptance Criteria:

- [ ] default to open, so that when a user arrives for the first time from a search, they will understand the context of the course they are looking at
- [ ] when the user closes the course info drawer, keep it closed
- [ ] If the user's browser doesn't support local storage or cookies, fail quietly without raising an error to sentry (see https://github.com/mitodl/ocw-hugo-themes/issues/645) 

#### Related issues

- #765

PR #798, opened 2022-07-19
course drawer state maintained

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/795

#### What's this PR do?
- Maintains desktop course drawer state across different course pages.
- **`Approach`**:
  - Whenever the user, either closes or opens the course drawer, its state (`opened` or `closed`) is stored in local storage.
  - Whenever any page is loaded, a function is triggered to check if there is any previously stored drawer state in local storage:
      - If yes, then it either closes or opens the drawer depending upon that state.
      - If not, then it sets the state to `opened` and opens the drawer by default 

- Checkout this branch.
- Build any course locally.
- Go to any page of that course (other than homepage)
- Verify that the course drawer is opened.
- Verify that the course drawer state (either opened or closed) is maintained across different course pages.
- One approach:
    - Close the drawer.
        - Either refresh the page or go to any other page and verify that the drawer is still closed. 
        - Close the tab, open a new tab and verify that the drawer is still closed.
    - Open the drawer and repeat the above steps again and now the drawer should remain open.
- Test this on multiple browsers (chrome, safari, firefox etc)
- Now block the cookies in your browser settings, test the flow again and verify that no error is logged in sentry, only an error message will be shown in console. (functionality will not work with blocked/disabled cookies)
- Test any other use-case that needs to be tested.  

- Currently this branch is created from `abdurrehman/compact-design` branch instead of `main`, once compact design PR is merged then we can change the base of this PR


pdpinch:
> we can also use session storage in order to maintain the state for a very longer period of time but not sure if its required or not

Can you explain this comment a bit more? How long is the state maintained in this PR?

MAbdurrehman12:
> Can you explain this comment a bit more? How long is the state maintained in this PR?

Ahh, Sorry! I was in a rush while writing this description and what I wrote is wrong. The data stored in `local storage` does not expire whereas data stored in session storage is cleared when the browser session ends. So for this PR, using `local storage` makes more sense, which is already being used. 
Please ignore this statement, I'll also remove it from the description.

course-v2/layouts/partials/desktop_course_info.html
```diff
@@ -1,6 +1,8 @@
 <div class="border-left">
-  <div id="desktop-course-info">
-    {{ partial "course_info.html" (dict "context" . "inPanel" true) }}
+  {{/*  adding d-none class here as this section (with id="desktop-course-drawer") is being controlled (shown/hidden) 
+  via desktop_course_drawer.ts hence hidden at the start to avoid a glitch  */}}
```
gumaerc, line 3: I don't think that this is accomplishing what you're trying to do here.  I get some layout thrashing upon refresh.  setting this div to `d-none` does indeed hide the div, but the parent element calling this partial places it inside a bootstrap column with `col-3` set on it with gives that section width regardless of its contents.  We should try and figure out a way to set these classes before the DOM loads to avoid layout thrashing.  One approach could be to inline the script as far up in the page as possible at the start of the `body` tag, rather than including it in the bundle.  If we do this and make the script pure JS (not using any libraries) then it should execute before anything else.

Here's a screen recording that demonstrates this: https://nextcloud.c4103.com/index.php/s/PkBpXfx4j6SmK8H

gumaerc (changes requested):
This functions as described, although I encountered some layout thrashing and I think we should try and see if we can address it:

course-v2/layouts/partials/desktop_course_info.html
MAbdurrehman12: @gumaerc Thanks for taking a look into this.
`d-none` is applied to hide the element when the HTML is being rendered so when JS is loaded (after a couple of milliseconds maybe) it shows/hides the drawer according to the user's preference stored in `local storage`. If we don't apply `d-none` in the first place and the user has hidden drawer then the drawer is first visible and then hidden when the JS is run which looked pretty bad. 
But I agree that the `col-3` element should also be not shown when the drawer is not showing so I guess I'll make this update to hide the complete section including the outer/parent, `col-3` element. Does this sound fine to you? 

For layout thrashing or I'd say the glitch: I also had this in notice but was not sure how much noticeable it is for the end users so I sent a video to @pdpinch in order to take his thoughts and we came to this conclusion that it's not too much noticeable but I think Peter will be in a better position to tell once he experiences it himself.

I have already tried the approach of inline vanilla JS but it also had some issues and was not working perfectly. If you feel that this issue needs to be catered then I think I'll give this inline approach another shot and see how things go. If things don't go well, then we'll have to come up with a different approach.

MAbdurrehman12: @gumaerc I tried inline JS (moved all the JS code) in `baseof.html` and it works fine. Now I'll update the PR shortly and it will be up for review again. 
Thanks

gumaerc (approved):
:+1: LGTM!

gumaerc:
@MAbdurrehman12 I took a look at the revised inline approach and it looks even better.  Your previous commits fixed the issue with the empty column popping in and out, but I was still seeing a bit of flickering on the button itself to open the drawer if it was in a closed state and I refreshed the page.  With the inline'd approach, I don't see any flickering at all.  You can merge this when you're ready.

MAbdurrehman12:
Thanks @gumaerc 
Yeah, this inline JS approach worked well.
So no flickering/glitch now.

I have pushed another commit, improving the error handling stuff when user has either blocked cookies or their browser has no storage support. Do you want to have another look?

gumaerc:
> Thanks @gumaerc Yeah, this inline JS approach worked well. So no flickering/glitch now.
> 
> I have pushed another commit, improving the error handling stuff when user has either blocked cookies or their browser has no storage support. Do you want to have another look?

Just tested this out blocking cookies in an incognito window and it worked as expected.

Issue #802, opened 2022-07-21
Styling tweaks for video player and video player page


pdpinch:
![image](https://user-images.githubusercontent.com/430126/180290797-3cf03dab-ad14-433f-9ec0-858ebcd0c985.png)

from https://ocw.mit.edu/courses/18-02-multivariable-calculus-fall-2007/resources/lecture-3-matrices/

#### Designs and Mockups

https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/461529499/preview?scrollOffset=1982

#### Acceptance Criteria:

- [ ] The video timeline should extend the full width of the video player. Compare a youtube player to ours. 
- [ ] "Related Resource" UI should have same styling as Transcript UI: same controls, same colors

pdpinch:
If it's easier, these changes could be made to course-v2

MAbdurrehman12:
Closed by https://github.com/mitodl/ocw-hugo-themes/pull/803

PR #803, opened 2022-07-22
video player styling for course v2

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/802

#### What's this PR do?
- Updates the style of video progress bar so it covers the complete width of video player.
- Updates toggleable tabs under the video to make them uniform/same. (`Related resources` and `optional` tab's UI made same as of `transcription` tab UI)

- Checkout this branch
- Build any course on new course theme, i.e: course-v2 (more detail in this [PR](https://github.com/mitodl/ocw-hugo-themes/pull/784)) which has videos, transcript, related resources and optional tab.
- Verify that video progress bar covers the complete width of video player.
- Verify that all the 3 toggleable tabs are same in terms of UI and functionality and they work as expected.
- Repeat the above steps on different browsers and different screen sizes. 
- **Important:** Verify that things in the previous course theme are working like before and this PR has no impact on that.

image

image

image


MAbdurrehman12:
@pdpinch I'm thinking of giving some margin from left and bottom to the content of `Related Resources` and `Optional` (Lecture Notes) tabs too, similar to what we have in the above transcript tab. 

image

Do you have any thoughts on this? or on the overall design?

gumaerc (approved):
@MAbdurrehman12 I'm going to approve this because the changes to the video player look mostly good and don't affect the player in the regular `course` theme, but there are some caveats:

 - Firstly, I noticed in Chrome that the transcript has a stock browser scrollbar on the right side that seems to be indented and looks kind of out of place:
 
 
image
 
 - Secondly, and this is probably more related to changes in `abdurrehman/compact-design`, I noticed that the v2 header seems to be applied even if I am not including `course-v2` in my config:
 
image
vs
image

The scroll bar issue is something we might want to look into in this PR before we merge it to the compact design branch, but the header issue should obviously be looked into on that branch directly rather than here.  Either way, if you want to merge this and then address those problems separately I'm fine with that too.

pdpinch:
> @pdpinch I'm thinking of giving some margin from left and bottom to the content of Related Resources and Optional (Lecture Notes) tabs too, similar to what we have in the above transcript tab.

I don't love it, but I we do need some separation, so folks know they can click and also so they have enough space between tappable areas (a mobile requirement). 

I think you should indent of contents of the hide/show the same amount. 

image

MAbdurrehman12:
@gumaerc Thanks for having a look into it.
I have updated the PR with this [feedback](https://github.com/mitodl/ocw-hugo-themes/pull/803/#issuecomment-1194483602) along with some other tweaks.

Regarding your point of scrollbar: I agree, it doesn't look really good but I think its the default one. 
Since I'm using mac so I have this type of default scrollbar:
Screenshot 2022-07-27 at 11 45 02 AM

However, if you think that we should customize this default scrollbar then I'd suggest doing it in other places too (i.e: page scrollbar or any other place where we have a scrolling section or page) for better consistency. And I think this can be done by a separate PR.

What do you think?

Perhaps @pdpinch could also share his thoughts?

gumaerc (approved):
@MAbdurrehman12 The updates look great; the indentation in the related resources section now matches the video time stamps.  While the scrollbar in Linux and Windows does indeed look kind of ugly, my feedback on that was more that it isn't flush with the right side of the container.  It's not really my place to say whether or not this matters, but it feels weird to me for a scroll bar to not be flush with the side of the container it's scrolling.  What do you think @pdpinch?  Either way, we can merge this as is if it matches the design.

MAbdurrehman12:
> my feedback on that was more that it isn't flush with the right side of the container

Ahh right. Thanks for pointing it out. I also did not notice it.
Anyhow, I've updated the PR to push it to the right-most side.
I think we can merge this now!

gumaerc (approved):
:+1: LGTM

Issue #811, opened 2022-08-05
course v2 sidebar and image section height width styling


MAbdurrehman12:
### Steps to Reproduce

- Either:
    - Build any course on new theme locally (How to do it? Check the description of this [PR](https://github.com/mitodl/ocw-hugo-themes/pull/784#issue-1291571736))
    - View this [course deployed on netlify](https://ocw-hugo-themes-course-pr-809--ocw-next.netlify.app/)
- Go to the course homepage
- View the height and width of sidebar and image section and compare it with the [design](https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/467361595/preview)

### Expected Behavior

- Height of all 3 sections (sidebar section, course info section, course image section) on the course homepage should be equal.

### Actual Behavior

- Sidebar section height is less
- Image section height and width are less.

### Screenshot or Screencast

(Ignore the other issues in the screenshot. will have separate tickets for them)

![Nuts-and-Bolts-of-Business-Plans-Sloan-School-of-Management-MIT-OpenCourseWare](https://user-images.githubusercontent.com/93309234/183045799-62f470a6-d3dc-469f-8aac-93f4ea438dae.png)

PR #814, opened 2022-08-05
sidebar and image section height and width styling

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/ocw-hugo-themes/issues/811

#### What's this PR do?
- Makes the height of all 3 sections on CHP the same.
- Makes the height of the sidebar section and the content section on interior pages the same.
- Increases the width of the course image section on the home page.
- Replaces bootstrap classes on sections of CHP and interior page with custom CSS to achieve more precision in section's widths.

- Build any v2 course locally or view [netlify course v2](https://ocw-hugo-themes-course-v2-pr-814--ocw-next.netlify.app)
- Verify the above-mentioned points on multiple browsers and screen dimensions. Also compare it with the [design](https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/467361595/preview)

#### Screenshots
image

### DIFF
--- a/course-v2/assets/css/course-home.scss
+++ b/course-v2/assets/css/course-home.scss
@@ -85,7 +85,7 @@
     }
 
     @include media-breakpoint-down(sm) {
-      width: 100% !important;
+      display: none;
     }
   }
 
--- a/course-v2/assets/css/course-home.scss
+++ b/course-v2/assets/css/course-home.scss
@@ -22,6 +22,7 @@
     border-radius: 0px;
     box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.16);
     border: 0px;
+    height: 100%;
   }
 
   .container-fluid {
--- a/course-v2/assets/css/course-home.scss
+++ b/course-v2/assets/css/course-home.scss
@@ -21,10 +21,7 @@
       padding-left: 5px !important;
       padding-right: 5px !important;
     }
-    
   }
-
-  
 
   .card {
     border-radius: 0px;
@@ -64,32 +61,31 @@
     }
   }
 
-
-#desktop-nav {
-  width: 18% !important;
-}
-
-#course-detail {
-  width: 50% !important;
-
-  @include media-breakpoint-down(md) {
-    width: 65% !important;
+  #desktop-nav {
+    width: 18% !important;
   }
 
-  @include media-breakpoint-down(sm) {
-    width: 100% !important;
-  }
-}
+  #course-detail {
+    width: 50% !important;
 
-#course-image-section {
-  width: 32% !important;
+    @include media-breakpoint-down(md) {
+      width: 65% !important;
+    }
 
-  @include media-breakpoint-down(md) {
-    width: 35% !important;
+    @include media-breakpoint-down(sm) {
+      width: 100% !important;
+    }
   }
 
-  @include media-breakpoint-down(sm) {
-    width: 100% !important;
+  #course-image-section {
+    width: 32% !important;
+
+    @include media-breakpoint-down(md) {
+      width: 35% !important;
+    }
+
+    @include media-breakpoint-down(sm) {
+      width: 100% !important;
+    }
   }
-}
 }
--- a/course-v2/assets/css/course-home.scss
+++ b/course-v2/assets/css/course-home.scss
@@ -8,15 +8,23 @@
   }
 
   .course-home-grid {
-    padding-left: 5px !important;
-    padding-right: 5px !important;
+    padding-left: 10px !important;
+    padding-right: 10px !important;
   }
 
-  .course-cards {
-    @include media-breakpoint-down(xs) {
+  @include media-breakpoint-down(sm) {
+    .course-cards {
       flex-flow: column-reverse;
     }
+
+    .course-home-grid {
+      padding-left: 5px !important;
+      padding-right: 5px !important;
+    }
+    
   }
+
+  
 
   .card {
     border-radius: 0px;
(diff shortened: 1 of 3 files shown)


Wassaf-Shahzad (changes requested):
LGTM!, I just had one concern, The panel on the right most side takes about 25% for the available space. Should it be smaller ?
Screenshot 2022-08-15 at 4 19 03 PM

MAbdurrehman12:
> The panel on the right most side takes about 25% for the available space. Should it be smaller

This PR was created to increase its width, as initially, its width was smaller but then it was pointed out. 
See the [issue ticket](https://github.com/mitodl/ocw-hugo-themes/issues/811) and the [design](https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/467361595/preview).
Though I agree, as compared to the design, its width is a bit larger but this difference is due to the usage of the bootstrap grid system here (we are not using our own `px` values). Since bootstrap cols have their own width and I'd suggest not overriding them for better consistency across the app.
but if I'm being mistaken or the width should be smaller, please let me know.

Maybe if @Ferdi  @pdpinch @gumaerc has any thoughts?

pdpinch:
In the design, the image width is 361px, so this is definitely too wide. Can we reduce it from 4 columns to 3? Or is there some other way to target the image width on a desktop?

MAbdurrehman12:
> Can we reduce it from 4 columns to 3? 

Previously, it was set to 3 columns (have a look [here](https://live-qa.ocw.mit.edu/course-v2/courses/wgs-s10-special-topics-in-women-gender-studies-seminar-latina-womens-voices-spring-2010/)), but then to increase its width, I changed it to 4 columns in this PR.
But if you think that it looks fine with 3 columns, then maybe we can change it back to 3?

> is there some other way to target the image width on a desktop?

1 approach is to use bootstrap (already using).
Another approach is to use CSS to make our own grids and custom width etc, for that, we'll also have to cater the responsiveness by using media queries etc. 
My suggestion would be to stick with bootstrap but if we really want to have the exact width, we can proceed with writing our own CSS.

(btw I'm talking about the width of the complete image section, not just the image only)

gumaerc:
@MAbdurrehman12 Another thing we could try is overriding the SASS variables used to construct the bootstrap styles. Bootstrap columns are constructed based on percentages using flexbox, so while you wouldn't want to give a column a specific width, you can adjust the total amount of columns in a grid.  The default is 12, but you can [override it](https://getbootstrap.com/docs/4.0/layout/grid/#customizing-the-grid):

```
$grid-columns: 12 !default;
$grid-gutter-width: 30px !default;
$grid-breakpoints: (
  xs: 0,
  sm: 480px,
  md: 768px,
  lg: 1024px
);

$container-max-widths: (
  sm: 420px,
  md: 720px,
  lg: 960px
);
```

We have a bootstrap overrides file at `base-theme/assets/css/bootstrap-overrides.scss`. We should be able to override the `$grid-columns` variable to increase the total number of columns, increasing the granularity of each column and giving us finer control.

MAbdurrehman12:
> we could try is overriding the SASS variables used to construct the bootstrap styles

Yeah, I had this in mind but didn't proceed with it to maintain consistency in bootstrap styles: Changing them globally will have a widespread impact, so we can only change it for this section (not sure if it'll be a good approach).

> you can adjust the total amount of columns in a grid

I gave this a shot but it looks like the exact/close to exact width is still not being achieved by adjusting the columns. As reducing a column makes it a bit small and increasing the column makes it a bit large, lol. Adjusting columns also have an impact on the other sections (side navbar, course description section).

I think if the initial width, which can be seen [here](https://live-qa.ocw.mit.edu/course-v2/courses/wgs-s10-special-topics-in-women-gender-studies-seminar-latina-womens-voices-spring-2010/), is not up to the mark, then I'll just go on and override bootstrap col width for CHP only to achieve the exact width as mentioned in the design.

Maybe @pdpinch can share his thoughts on this?

pdpinch:
pixel-perfect is not a concern, but "too wide" is a concern.

MAbdurrehman12:
@gumaerc Your point of having 100 `grid columns` and considering `col` as a percentage makes sense. 
I'm having some trouble applying this change only to CHP (since `$grid-columns` are fetched and defined on the compilation I think) as I don't think it will be a good idea to make this change globally as it will have quite a widespread impact. 
Do you have any possible solution in mind?

After this, we can identify the percentage of the `image` section, along with the `side-navbar` and `course description` section from the design and use them.
Currently, in design, it looks like that image section is taking up to 31-32% of the total grid. 
and if we look into this PR, we have assigned `4 cols`, out of 12, to the image section which is 33% ((4/12)*100).

So I wonder, even after, applying the 100 `grid columns` approach, the result will almost be the same. 
and if 31-33% is too wide, then do we need to get the design updated? or I'm missing something? 

(its not looking too wide in the design since maybe the overall design width is a bit narrow?)

gumaerc:
@MAbdurrehman12 It was just a suggestion of something to try out. Like you pointed out, it seems we're already getting pretty close to the ratio we want so it may not be worth the extra work to re-work `$grid-columns`. If it needs to be more precise than it is, maybe we just want to write some custom CSS to do so. We will have to do responsive breakpoints manually though.

MAbdurrehman12:
I'll just go on, remove bootstrap cols and add custom CSS to be more precise.

Wassaf-Shahzad (approved):
Works great !

MAbdurrehman12:
>  At larger mobile resolutions, the course image is still visible. In the current course theme, we hide the course image at these resolutions

Thanks for pointing this out @gumaerc. 
I've now also hidden the course image section on smaller screens to make it the same as in the current course theme. 
I think, now we can merge this PR!

PR #816, opened 2022-08-05
added safeguard for course description

### LINKED ISSUE #815: Fix build Failure
### Steps to Reproduce
Build failing for ocw-hugo-theme. 

### Stacktrace
[Stack trace](https://cicd.odl.mit.edu/sky/issuer/auth?access_type=offline&client_id=concourse-web&redirect_uri=https%3A%2F%2Fcicd.odl.mit.edu%2Fsky%2Fcallback&response_type=code&scope=openid+profile+email+federated%3Aid+groups&state=eyJyZWRpcmVjdF91cmkiOiIvdGVhbXMvb2N3L3BpcGVsaW5lcy9kcmFmdC9qb2JzL2J1aWxkLW9jdy1zaXRlL2J1aWxkcy80P3ZhcnMuc2l0ZT0lMjIyMXctNzgwLWNvbW11bmljYXRpbmctaW4tdGVjaG5pY2FsLW9yZ2FuaXphdGlvbnMtZmFsbC0yMDAxJTIyIiwiZW50cm9weSI6IjljOWQ0MmUxNTFmZTE0ZDhmYWExMGE1NWU0ZjAxZWU5ZmYxNTk2MTUyOThiMzY2MjE0ZDFkZDM3ZTgzNGM3ZDIifQ%3D%3D)

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing

Closes #815 

#### What's this PR do?
Adds check to home.html and course description partial to verify if description is available.

- Start ocw-hugo-themes course
- Remove course description in course.json
- Verify hugo doesn't crash

### DIFF
--- a/course-v2/layouts/partials/course_description.html
+++ b/course-v2/layouts/partials/course_description.html
@@ -1,7 +1,7 @@
 {{ $courseData := .context.Site.Data.course }}
 {{ $shouldCollapseDescription := false }}
 {{ with $courseData.course_description }}
-  {{ $shouldCollapseDescription := gt (len .) 320 }}
+  {{ $shouldCollapseDescription = gt (len .) 320 }}
 {{ end }}
 
 <div class="course-description">
--- a/course-v2/layouts/partials/course_description.html
+++ b/course-v2/layouts/partials/course_description.html
@@ -1,5 +1,8 @@
 {{ $courseData := .context.Site.Data.course }}
-{{ $shouldCollapseDescription := gt (len $courseData.course_description) 320 }}
+{{ $shouldCollapseDescription := false }}
+{{ with $courseData.course_description }}
+  {{ $shouldCollapseDescription := gt (len .) 320 }}
+{{ end }}
 
 <div class="course-description">
   <div class="course-detail-title">
--- a/course/layouts/home.html
+++ b/course/layouts/home.html
@@ -57,8 +57,11 @@
                           {{ partial "mobile_nav_toggle.html" . }}
                         </div>
                         <!-- the number here is a estimate, a count of 5 lines of characters on a sample course at minimum browser width -->
-                        {{ $shouldCollapseDescription := gt (len $courseData.course_description) 320 }}
-                        <div class="col-lg-8 course-description">
+                        {{ $shouldCollapseDescription := false}}
+                        {{ with $courseData.course_description }}
+                          {{ $shouldCollapseDescription := gt (len .) 320 }}
+                        {{ end }}
+                          <div class="col-lg-8 course-description">
                           <h4 class="font-weight-bold course-description-title">Course Description</h4>
                           <div id="course-description" class="description {{ if $shouldCollapseDescription }}collapse{{ end }}" aria-expanded="false">
                             {{- $courseData.course_description | markdownify -}}
--- a/course/layouts/home.html
+++ b/course/layouts/home.html
@@ -59,7 +59,7 @@
                         <!-- the number here is a estimate, a count of 5 lines of characters on a sample course at minimum browser width -->
                         {{ $shouldCollapseDescription := false}}
                         {{ with $courseData.course_description }}
-                          {{ $shouldCollapseDescription := gt (len .) 320 }}
+                          {{ $shouldCollapseDescription = gt (len .) 320 }}
                         {{ end }}
                           <div class="col-lg-8 course-description">
                           <h4 class="font-weight-bold course-description-title">Course Description</h4>


MAbdurrehman12 (changes requested):
So this build failure occurs when we don't have `course_description` key in `course.json`? 
I'm wondering how it went missing in the first place? Shouldn't that be fixed? Because if it happens for any other key, the build will fail again. 
and adding checks for all the keys might a widespread change.

So I think the key/object should always be there in the `course.json`, regardless of it's value, which can/can not be empty.

I have tried removing the description and other fields in the course metadata via studio and published the course, the fields/keys were there in the `course.json` of the published course but with empty values. So I'm not sure how did this (missing keys) happen. Maybe @gumaerc or @mbertrand might have some more context of this.

ps: the check you added seems fine, however, the following approach looks a bit easy as compared to `with` usage.
```
{{ $courseDescription := $courseData.course_description | default "" }}
```
and now we can use this `$courseDescription` wherever required

gumaerc:
@MAbdurrehman12 @Wassaf-Shahzad So, I think the reason that we are seeing failed builds with this particular error has to do with the fact that when a new site is created in `ocw-studio`, you are allowed to attempt to publish the site without first setting the required metadata. I just attempted creating a fresh site in RC and it created [this](https://github.mit.edu/ocw-content-rc/cpg-test-course-9) repo. Notice how there is nothing in the repo aside from the readme. Attempting to build a draft for this site generates the expected error:

```
Error: Error building site: failed to render pages: render of "home" failed: "/tmp/build/d50ae459/ocw-hugo-themes/course/layouts/home.html:60:61": execute of template failed: template: home.html:60:61: executing "home.html" at <len $courseData.course_description>: error calling len: reflect: call of reflect.Value.Type on zero Value
```

Attempting to go to the metadata section and hit save will trigger client side errors regarding missing required fields:
image

So, I think in `ocw-studio` we should do one or both of these:
 - Prevent publishing until all required metadata is filled out
 - In the code that initially creates the Github repo for a site, make sure the required metadata file (`data/couse.json`) is created with blank default values

Sane defaults in the templates wherever we can add them like this is also a good idea, but ultimately a game of whack-a-mole unless we create some kind of partial for fetching any course level metadata and do the check there. We could create a partial, something like `course_metadata.html` and have the argument be a key to fetch, then have error handling in the partial itself. That way, any metadata fetching in the future done through this hypothetical partial would have errors handled.

Wassaf-Shahzad:
> So this build failure occurs when we don't have `course_description` key in `course.json`? I'm wondering how it went missing in the first place? Shouldn't that be fixed? Because if it happens for any other key, the build will fail again. and adding checks for all the keys might a widespread change.
> 
> So I think the key/object should always be there in the `course.json`, regardless of it's value, which can/can not be empty.
> 
> I have tried removing the description and other fields in the course metadata via studio and published the course, the fields/keys were there in the `course.json` of the published course but with empty values. So I'm not sure how did this (missing keys) happen. Maybe @gumaerc or @mbertrand might have some more context of this.
> 
> ps: the check you added seems fine, however, the following approach looks a bit easy as compared to `with` usage.
> 
> ```
> {{ $courseDescription := $courseData.course_description | default "" }}
> ```
> 
> and now we can use this `$courseDescription` wherever required

Although I agree, that the above approach is more readable. However I dont think we should be assigning a default value just to essentially calculate `$shouldCollapseDescription`.

course-v2/layouts/partials/course_description.html
```diff
@@ -1,5 +1,8 @@
 {{ $courseData := .context.Site.Data.course }}
-{{ $shouldCollapseDescription := gt (len $courseData.course_description) 320 }}
+{{ $shouldCollapseDescription := false }}
+{{ with $courseData.course_description }}
+  {{ $shouldCollapseDescription := gt (len .) 320 }}
```
gumaerc, line 4: If we go with this approach, when you reassign variables in a Go template you need to drop the `:`. So, it should look like this:

```
{{ $shouldCollapseDescription := false }}
{{ with $courseData.course_description }}
  {{ $shouldCollapseDescription = gt (len .) 320 }}
{{ end }}
```

Using `with` is a fine approach when you are not sure if the variable will exist, and this will ensure that `$shouldCollapseDescription` will default to false. The approach @MAbdurrehman12 suggested would work as well:

```
{{ $courseDescription := $courseData.course_description | default "" }}
{{ $shouldCollapseDescription = gt (len $courseDescription) 320 }}
```

In the scope of just addressing the course description I don't have an issue with either of these approaches, but I think that we should look at the big picture and come up with a solution for error handling when it comes to fetching any course level metadata that might be un-set like I described above.

gumaerc (changes requested):
This looks okay, but if you take this approach you need to adjust it a bit:

course-v2/layouts/partials/course_description.html
gumaerc: P.S. The larger issue of checking all of our course level metadata before accessing doesn't have to be completed in this PR btw, I'm fine with just merging this once the syntax is corrected here. In a later PR we should address the larger issue of sanely handling errors for any metadata fetching.

MAbdurrehman12:
> Although I agree, that the above approach is more readable. However I dont think we should be assigning a default value just to essentially calculate $shouldCollapseDescription

Just curious: What is the reason/con for not assigning a default value? (even if it has only 1 usage)

btw it has 2 usages: 1 for calculating `$shouldCollapseDescription` and 1 for rendering/showing.
But anyhow, this `with` will also work fine.

course/layouts/home.html
```diff
@@ -57,8 +57,11 @@
                           {{ partial "mobile_nav_toggle.html" . }}
                         </div>
                         <!-- the number here is a estimate, a count of 5 lines of characters on a sample course at minimum browser width -->
-                        {{ $shouldCollapseDescription := gt (len $courseData.course_description) 320 }}
-                        <div class="col-lg-8 course-description">
+                        {{ $shouldCollapseDescription := false}}
+                        {{ with $courseData.course_description }}
+                          {{ $shouldCollapseDescription := gt (len .) 320 }}
```
gumaerc, line 62: This is also variable reassignment and you need to drop the `:`

gumaerc (changes requested):
Missed a spot where there's still incorrect variable reassignment:

gumaerc (approved):
:+1:

PR #818, opened 2022-08-10
added topics partial to course-v2-home to render topics in a hierarchy 

### LINKED ISSUE #812: course v2 home page topics hierarchical structure
### Steps to Reproduce

- Either:
    - Build any course on new theme locally (How to do it? Check the description of this [PR](https://github.com/mitodl/ocw-hugo-themes/pull/784#issue-1291571736))
    - View this [course deployed on netlify](https://ocw-hugo-themes-course-pr-809--ocw-next.netlify.app/)
- Go to the course homepage
- View topics structure

### Expected Behavior

- Topic's structure should be hierarchical same as in the [design](https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/467361595/preview).

### Actual Behavior

- Topic's structure is not hierarchical as shown in the [design](https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/467361595/preview),

### Screenshot or Screencast

(Ignore the other issues in the screenshot. will have separate tickets for them)

Nuts-and-Bolts-of-Business-Plans-Sloan-School-of-Management-MIT-OpenCourseWare

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

Closes #812

#### What's this PR do?
This pr adds the `topics` partial to course v2 home page. Which causes the topics listed in the home page to be rendered in hierarchical order.

- Build any v2 course locally or view [netlify course v2](https://ocw-hugo-themes-course-v2-pr-818--ocw-next.netlify.app/) 
- Visit home page
- Verify topics are rendered hierarchical order

In Course theme version 2, we moved away from rendering topics on the home page in a single line with `>` delimiter to using the same format we use to render course topics in course info. To achieve that result, we replace topics_online partial with topics partial. 

Screenshot 2022-08-10 at 3 03 39 PM
Screenshot 2022-08-10 at 3 03 48 PM


MAbdurrehman12 (changes requested):
Overall it looks good!
However I just noticed a couple of things: 

1. If the partial `topics_oneline.html` has become completely obsolete and we are no longer using it then should we delete it from `course-v2` theme?
2. Do we need to apply some left margin to the sub-topics as shown in the design? (so the parent topic and its sub-topcis are distinguished)

Design:
image
Current structure: 
image
So if we give `ml-2` class [here](https://github.com/mitodl/ocw-hugo-themes/blob/02f05465a675b428dbd4bb96d4350d16f714f6fd/course-v2/layouts/partials/topic.html#L40) and [here](https://github.com/mitodl/ocw-hugo-themes/blob/02f05465a675b428dbd4bb96d4350d16f714f6fd/course-v2/layouts/partials/topic.html#L46) then we have something like this: 
image
I think we can also place a check: if the sub-topic is toggle-able (has further sub-topics), then don't apply `ml-2` class so we don't have this space b/w toggle arrow and sub-topic like shown below (see the space between history and its arrow):  
image
4. Lastly, the topics having sub-topics and the topics not having sub-topics are not aligned: 
image
Maybe, we can align them to avoid any confusion?

Ferdi:
A bit of left margin would be great

MAbdurrehman12 (commented):
Looks good!

However, this 4th point is still a bit noticeable
> 4. Lastly, the topics having sub-topics and the topics not having sub-topics are not aligned

As currently, we have something like this: 
image
Since topics are not aligned, I feel like this is confusing for the user.
Maybe I'm being a bit too paranoid. 

But something like this would look a bit cleaner I guess: 
image
(I just did this by adding some left margins etc, but I think we can figure out a more suitable and better approach)

MAbdurrehman12 (approved):
Looks good 👍 

Now `Topics` heading and its items are not aligned. Do we also need to apply some left padding to topics heading as well? 
image

So it can be aligned/same as other headings.
e.g: 
image

But if we do this, we will also have to take care of it on smaller screens.

Not sure if this is something to be taken care of. @pdpinch can comment better.

MAbdurrehman12 (approved):
Looks & works nice!

PR #809, opened 2022-08-04
build and deploy course v2 on netlify

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
- None

#### What's this PR do?
- For github workflow:
    - Creates a separate build for course v2
    - Deploy it to netlify so now we have both course versions deployed on netlify.
    - sets `overwrites-pull-request-comment` to `true` so there's only 1 netlify comment instead of comment on every pushed commit(s). 

#### How should this be manually tested?
- Verify that course v2 is also deployed along with course v1
- Verify that the changes made in this PR has no side-effect/impact on anything.

PR #824, opened 2022-08-15
Netlify deployment: www, course, course-v2 comment

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

- None

#### What's this PR do?
- A flag `overwrites-pull-request-comment` was set to `true` in this PR https://github.com/mitodl/ocw-hugo-themes/pull/809/ to avoid/stop the bot from posting new comments on every commit. As ideally, there should be only initial netlify comments which should be overridden on every commit instead of new comments. But it turns out that the bot is posting only 1 comment and overriding/edititing it for `www`, `course`, and `course-v2`. [Example](https://github.com/mitodl/ocw-hugo-themes/pull/820#issuecomment-1211902434). Hence that change is being reverted.
- As suggested [here](https://github.com/mitodl/ocw-hugo-themes/pull/824#discussion_r946099853), we can combine and show all the deployment links in 1 comment instead of having 3 separate comments. This has been achieved by manually creating/updating the comment after `course-v2` build.

- View deployment [comment](https://github.com/mitodl/ocw-hugo-themes/pull/824#issuecomment-1218025657).
- For testing: 
    - Push any commit with a change that doesn't affect anything. (maybe some new line in a file or correct some formatting anywhere).
    - Verify that no new comment is posted by the bot after this commit.
    - Verify that `www`, `course`, `course-v2` are built, deployed and the deployed [comment](https://github.com/mitodl/ocw-hugo-themes/pull/824#issuecomment-1218025657) is edited.
    - Verify no github action msgs are displayed. View this [comment](https://github.com/mitodl/ocw-hugo-themes/pull/824#issuecomment-1218444582) for some more context

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -89,7 +89,7 @@ jobs:
           RESOURCE_BASE_URL: ${{ secrets.RESOURCE_BASE_URL }}
 
       - name: Deploy www Preview to Netlify
-        uses: nwtgck/actions-netlify@v1.1
+        uses: nwtgck/actions-netlify@v1.2.3
         if: ${{ github.event_name == 'pull_request' }}
         with:
           publish-dir: '../ocw-www/public'
@@ -104,7 +104,7 @@ jobs:
           NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
       
       - name: Deploy Course Preview to Netlify
-        uses: nwtgck/actions-netlify@v1.1
+        uses: nwtgck/actions-netlify@v1.2.3
         if: ${{ github.event_name == 'pull_request' }}
         with:
           publish-dir: '../ocw-course/public-v1'
@@ -119,7 +119,7 @@ jobs:
           NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
 
       - name: Deploy v2 Course Preview to Netlify
-        uses: nwtgck/actions-netlify@v1.1
+        uses: nwtgck/actions-netlify@v1.2.3
         if: ${{ github.event_name == 'pull_request' }}
         with:
           publish-dir: '../ocw-course/public-v2'
@@ -134,7 +134,7 @@ jobs:
           NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
 
       - name: Deploy CI build to Netlify
-        uses: nwtgck/actions-netlify@v1.1
+        uses: nwtgck/actions-netlify@v1.2.3
         if: ${{ github.event_name == 'push' }}
         with:
           publish-dir: '../ocw-www/public'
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           deploy-message: "Deploy from GitHub Actions"
           enable-pull-request-comment: true
           enable-commit-comment: false
-          overwrites-pull-request-comment: true
+          overwrites-pull-request-comment: false
```
gumaerc, line 100: So instead of having every deploy post its own comment, we might instead try a different strategy. In the `nwtgck/actions-netlify` readme, this is listed as an option: `deploy-message A custom deploy message to see on Netlify deployment (e.g. ${{ github.event.pull_request.title }})`

If we simply set one of the deploys with a custom message including the URLs for all the expected deploys, then we should be able to use the "overwrite comment" option. If we put the comment on the `www` build, the `course` and `course-v2` deploys might not be ready yet when the comment is posted / updated, but I think that's a fine tradeoff and better than the PR getting flooded with comments. Since the PR number will be the same for each deploy, we should have all the info we need.

gumaerc (changes requested):
I think there's a better strategy to be had here regarding the comments that I detail below, but I also think we should also update `nwtgck/actions-netlify` to 1.2.3 which is the latest version.

gumaerc:
@MAbdurrehman12 I didn't realize that `deploy-message` isn't the Github comment but rather the comment in the Netlify deploy that you see on Netlify.com in the dashboard. I'm glad that you took the suggestion and ran with it though, what you've got going with manually updating the comment looks solid. It looks like it's working, but let me know when it's done and ready to be reviewed.

MAbdurrehman12:
@gumaerc yeah, the only issue which I'm seeing at the moment is that with the latest `nwtgck/actions-netlify` version (1.2.3), there are some github action msgs created with every commit we push
image
and this is something which we are already trying to avoid.
Version 1.1 doesn't do this.

Do you have any idea about it? 
There must be a config for it which I'm not sure of or maybe we can stick to the 1.1 version?

I'm also thinking of closing this PR and opening a new one as this PR has a lot of unnecessary comments (created during testing as I was unable to test this flow locally).

What do you suggest?

gumaerc:
@MAbdurrehman12 You might try setting this:

```
enable-pull-request-comment: false
enable-commit-comment: false
```

If you're updating the comment manually with a separate step, no need to have the Netlify action do it.  I'm fine with staying at 1.1 though, I just figured that would be a "while we're in here" thing but if it's causing complications then no need.

MAbdurrehman12:
> You might try setting this

they are already set to false.

gumaerc:
`enable-commit-comment: false` is not set, but looking at the readme I don't think that would hide those messages. I'm not sure where those messages are coming from, but if holding back on the version upgrade prevents them from showing then we should just do that for now.

MAbdurrehman12:
Okay.
Should I open a new PR then and link this PR in it so this conversation remains a part of it? 
or we can proceed with this one?

gumaerc:
@MAbdurrehman12 This PR is fine

gumaerc (approved):
:+1:

Issue #831, opened 2022-08-24
lighthouse build failing on main branch


gumaerc:
### Steps to Reproduce

Run lighthouse checks on the `main` branch

### Expected Behavior

The action completes without error

### Actual Behavior

The action errors out on the "Find Comment" step, likely because the check isn't being run in the context of a PR: https://github.com/mitodl/ocw-hugo-themes/runs/8003200325?check_suite_focus=true

### Stacktrace

```
Run peter-evans/find-comment@v2
Error: Not Found
```

PR #834, opened 2022-08-29
Added check to other create-or-update action commands 

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Part of #831 

#### What's this PR do?
- Adds command to continue on failure to other create-or-update action commands since the if block does not work on failures.

- Deleting all netlify comments and then running lighthouse job should be good enough


MAbdurrehman12 (approved):
👍

gumaerc (approved):
:+1:

Issue #838, opened 2022-09-07
offline theme should not render search links


gumaerc:
### Steps to Reproduce

 - Render a course site using the `course-offline` theme

### Expected Behavior

Text wrapped in search links, like instructor names, are rendered as plain text

### Actual Behavior

The search links are still rendered

PR #839, opened 2022-09-07
remove search links in offline course theme

gumaerc:
#### Pre-Flight checklist

- [ ] Testing

Closes https://github.com/mitodl/ocw-hugo-themes/issues/838

#### What's this PR do?
This PR expands on the `course-offline` theme implemented in https://github.com/mitodl/ocw-hugo-themes/pull/832. It takes the places where a course site would render search links to `/search` and renders a span instead. The offline sites are meant to be downloaded and viewed offline with the course as the root so the links would not work and thus have been removed.

 - Temporarily overwrite the contents of `course-offline/layouts/partials` with the following so the course site renders properly with the webpack dev server:
```
{{- if or site.IsServer (not (eq site.BaseURL "")) -}}
  {{- if or site.IsServer (in site.BaseURL ":3000") -}}
    {{- if hasPrefix .url "/" -}}
      {{- print "http://localhost:3001" .url -}}
    {{- else -}}
      {{- print "http://localhost:3001/" .url -}}
    {{- end -}}
  {{- else -}}
    {{- partial "site_root_url.html" .url -}}
  {{- end -}}
{{- else -}}
  {{- partial "site_root_url.html" .url -}}
{{- end -}}

```
 - If you've never run a course site locally before, read the readme and make sure you're set up to do so
 - Clone [`ocw-hugo-projects`](https://github.com/mitodl/ocw-hugo-projects) locally and point to it in your `.env` using `COURSE_HUGO_CONFIG_PATH=/path/to/ocw-hugo-projects/ocw-course-v2/config-offline.yaml`
 - Spin up the course with `npm run start:course`
 - Verify the following:
   - http://localhost:3000 - The course info section on the home page is now all text and doesn't contain links
   - Click any of the nav links to go to a different section and look at the course info section, make sure there are no links to `/search`
   - Verify that the magnifying glass icon is not present in the header that normally links to search


MAbdurrehman12 (approved):
Looks good 👍

just wondering that if we have removed search links from offline course since it has to be viewed offline, then what about the other www links? (about, contact etc pages)

gumaerc:
> just wondering that if we have removed search links from offline course since it has to be viewed offline, then what about the other www links? (about, contact etc pages)

@MAbdurrehman12 I did think about this and am meeting with Peter and Ferdi to discuss the remaining work on offline sites today.

PR #844, opened 2022-09-14
Fix table renders for course-v1 when table width is greater then main-content width

### LINKED ISSUE #835: Table renders in "mobile" format even on desktop
### Steps to Reproduce

Visit https://draft.ocw.mit.edu/courses/res-3-006-micro-mentoring-resources-and-materials-science-curriculum-spring-2021/pages/application-evaluation-rubric/ to see the table that will not format correctly. Studio: https://ocw-studio.odl.mit.edu/sites/micro-mentoring-resources-and-materials-science-curriculum/type/page/edit/80e34d94-aa1f-45ba-ad92-625b174ae5db/

### Expected Behavior

Tables should be formatted for desktop display, with rows and columns.

### Actual Behavior

Table is rendered in mobile "stacked" format even at desktop widths. 

### Screenshot or Screencast

Screen Shot 2022-08-30 at 9 20 23 AM

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes #835 

#### What's this PR do?
- Removes console.log in course-info-toggle.js
- Fixes bug for course info buttons text
- Closes course info drawer if table width is greater than content width
- Reruns table width calculations on course info toggle

- Start ocw-course using course "RES.3-006_Sp21"
- Open a page with a table and verify the following
    - On pressing couse-info toggle table width is recalculated. ie table renders in mobile form when drawer is open and renders normally when drawer is closes
    - On page load drawer is closes

(Optional)
Screenshot 2022-09-14 at 5 06 38 PM
Screenshot 2022-09-14 at 5 06 43 PM


course/assets/js/close_toggle_button.js
```diff
@@ -0,0 +1,17 @@
+// @ts-nocheck
```
gumaerc, line 1: Why does this code need to be in its own JS file? Couldn't we just put this function in `course_info_toggle.js` and then in `course.ts` do `import { closeToggleButton } from "./js/course_info_toggle"`?

gumaerc (changes requested):
While the changes in this PR do indeed fix some issues with activating the mobile table styles, I'm a little confused about some things. In [this](https://github.com/mitodl/ocw-hugo-themes/pull/798) PR, we set up `course-v2` to maintain the state of the drawer. I'm not saying that we necessarily want to backport that functionality to v1, but I'm wondering where the decision to always load the page with the drawer closed came from. I don't have an issue with it, as `course` will be replaced soon, just curious as to how this was decided. As far as the code, I only have one suggestion:

course/assets/js/close_toggle_button.js
Wassaf-Shahzad: Removed unnecessary file

Wassaf-Shahzad:
> While the changes in this PR do indeed fix some issues with activating the mobile table styles, I'm a little confused about some things. In [this](https://github.com/mitodl/ocw-hugo-themes/pull/798) PR, we set up `course-v2` to maintain the state of the drawer. I'm not saying that we necessarily want to backport that functionality to v1, but I'm wondering where the decision to always load the page with the drawer closed came from. I don't have an issue with it, as `course` will be replaced soon, just curious as to how this was decided. As far as the code, I only have one suggestion:

@gumaerc In this PR the drawer is never always closed. This PR only closes the drawer if the table width is greater than the main page width.

gumaerc (changes requested):
Thanks for consolidating that extra file. I took another look at this and I see what you mean by the code only triggering when toggling the course info drawer, but there is one other case we might want to handle. I did a quick screen recording to show you what I'm talking about:

https://nextcloud.c4103.com/index.php/s/WK5yNcGRkYjp8Zj

As you can see, the mobile style is successfully applied if I toggle the course info drawer in a way where doing so would push the drawer over the top of the table. However, if you toggle the drawer outside of this area and then collapse the window inward to the point where it would intersect, the mobile style is not applied when the course info drawer crosses this boundary. Can we look into making sure the mobile style is also applied in this case?

Wassaf-Shahzad:
> Thanks for consolidating that extra file. I took another look at this and I see what you mean by the code only triggering when toggling the course info drawer, but there is one other case we might want to handle. I did a quick screen recording to show you what I'm talking about:
> 
> https://nextcloud.c4103.com/index.php/s/WK5yNcGRkYjp8Zj
> 
> As you can see, the mobile style is successfully applied if I toggle the course info drawer in a way where doing so would push the drawer over the top of the table. However, if you toggle the drawer outside of this area and then collapse the window inward to the point where it would intersect, the mobile style is not applied when the course info drawer crosses this boundary. Can we look into making sure the mobile style is also applied in this case?

Sure will do

course/assets/js/responsive_tables.js
```diff
@@ -3,6 +3,9 @@ function initResponsiveTables() {
   const tables = document
     .getElementById("main-content")
     .getElementsByTagName("table")
+
+  const observer = new ResizeObserver(() => calculateOverlap())
```
gumaerc, line 7: `ResizeObserver` is a decent solution to this problem. The only issue with it is that while it has pretty wide support in modern browsers, there are older versions that it is not compatible with. It is also not compatible with any version of Internet Explorer. Because OCW is used in classrooms all over the world, it's important that major functionality be as backwards compatible as possible with old browsers and as compatible as possible with as many different browsers as possible, including IE. A lot of schools have computers that are locked down with older versions of Windows and IE is the only browser they would have access to. The overlap issue doesn't make the site unusable, so I think it's okay to merge this, but for critical functionality we should be as careful as possible to preserve this compatibility. Before you use any functionality like this, it's a good idea to check caniuse: https://caniuse.com/resizeobserver

gumaerc (approved):
:+1: LGTM, but the resizing issue is also a problem in `course-v2`. We can address that in a separate issue. One other thing to possibly think about:

course/assets/js/responsive_tables.js
Wassaf-Shahzad: To preserve the compatibility with IE we can use [polyfill implementation of resizeObserver ](https://www.npmjs.com/package/resize-observer-polyfill)

Issue #863, opened 2022-09-23
design update to "show course info" button for course-v2 theme


pdpinch:
#### Designs and Mockups

See https://projects.invisionapp.com/share/NB11D8TFZX28/#/screens/466127728

#### Acceptance Criteria:

- [x] replace textual button with simple left-pointing arrow in desktop view
- [x] no changes to mobile view 
- [x] anchor the button within the body div, not on the edge of the viewport 
- [x] include screenshots with the PR: desktop, extra-wide desktop, mobile

PR #872, opened 2022-09-29
style: "show course info" button replaced with an icon

MAbdurrehman12:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/ocw-hugo-themes/issues/863

#### What's this PR do?
- Removes the old `show course info` button.
- Adds a new icon (left arrow) button. 

- Checkout this branch
- Verify that when you close the course info drawer, a left arrow icon is shown instead of text button.
- Verify that this icon works as expected and open the course info drawer on clicking.
- Verify that this change is only applied on large screens and mobile view and functionality is unaffected.
- Test this on multiple screens sizes and on multiple browsers. 
- Verify that all the points mentioned in the acceptance criteria of issue ticket are fullfilled.

https://user-images.githubusercontent.com/93309234/193028305-0d18aff5-796c-44fb-9f73-5c6c823c1f95.mov

### DIFF
--- a/course-v2/assets/css/course-info.scss
+++ b/course-v2/assets/css/course-info.scss
@@ -34,10 +34,13 @@ $panel-course-info-text-color: #464646;
 
 .show-course-info-left-arrow {
   position: absolute;
-  right: 0;
-  padding: 8px;
-  margin-top: -17px;
   background-color: #e4e4e4;
+  margin-top: -17px;
+  padding-top: 5px;
+  right: 0;
+  width: 32px;
+  height: 32px;
+  text-align: center;
   cursor: pointer;
 }
 
--- a/course-v2/assets/css/course-info.scss
+++ b/course-v2/assets/css/course-info.scss
@@ -42,6 +42,10 @@ $panel-course-info-text-color: #464646;
   height: 32px;
   text-align: center;
   cursor: pointer;
+
+  .left-arrow {
+    margin-bottom: 2px;
+  }
 }
 
 .show-course-info-left-arrow:hover {
--- a/course-v2/assets/css/course-info.scss
+++ b/course-v2/assets/css/course-info.scss
@@ -34,13 +34,13 @@ $panel-course-info-text-color: #464646;
 
 .show-course-info-left-arrow {
   position: absolute;
-  right:0;
+  right: 0;
   padding: 8px;
   margin-top: -17px;
   background-color: #e4e4e4;
   cursor: pointer;
 }
 
-.show-course-info-left-arrow:hover{
+.show-course-info-left-arrow:hover {
   background-color: #c0c0c0;
 }
--- a/course-v2/assets/css/course-info.scss
+++ b/course-v2/assets/css/course-info.scss
@@ -35,8 +35,8 @@ $panel-course-info-text-color: #464646;
 .show-course-info-left-arrow {
   position: absolute;
   background-color: #e4e4e4;
+  border: none;
   margin-top: -18px;
-  padding-top: 5px;
   right: 0;
   width: 32px;
   height: 32px;
--- a/course-v2/assets/css/course-info.scss
+++ b/course-v2/assets/css/course-info.scss
@@ -31,3 +31,16 @@ $panel-course-info-text-color: #464646;
 .close-course-info:hover {
   background-color: #c0c0c0;
 }
+
+.show-course-info-left-arrow {
+  position: absolute;
+  right:0;
+  padding: 8px;
+  margin-top: -17px;
+  background-color: #e4e4e4;
+  cursor: pointer;
+}
+
+.show-course-info-left-arrow:hover{
+  background-color: #c0c0c0;
+}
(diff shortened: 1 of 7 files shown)


pdpinch:
That looks really good. 

Note that there's another issue about updating the mobile view buttons:  https://github.com/mitodl/ocw-hugo-themes/issues/877

course-v2/layouts/_default/baseof.html
```diff
@@ -42,6 +42,11 @@
               <div class="d-flex">
                 <div class="col-12 p-0" id="main-course-section">
                   <div class="card-body">
+                    {{ if not $isCourseHomePage }}
+                      <div id="show-desktop-course-drawer" class="show-course-info-left-arrow large-and-above-only">
+                        <img src="/images/left_arrow.svg" alt=""/>
+                      </div>
```
ChristopherChudzicki, line 48: **Requested Change:** This should be a button not a div (Right not it's not keyboard focusable as far as i can tell; more generally, its behavior is a button so it should be a button), and should have a reasonable aria-label. ("Show course info", perhaps?)1

ChristopherChudzicki (changes requested):
One request about using semantic html elements.

course-v2/layouts/_default/baseof.html
MAbdurrehman12: Done. Please have another look

ChristopherChudzicki (changes requested):
This does look really good.

One request: I think it also makes sense to change the X from a div to a button as part of this work.

If you do that, you probably won't need the `cursor: pointer` css rule anymore (on either the X or or the left arrow).

ChristopherChudzicki (changes requested):
Also, I just noticed that the buttons have no visual indication of focus. 

This seems to be because of a bootstrap 4 rule
```css
button:focus:not(.focus-visible) {
  outline: 0;
}
```
So apparently in Bootstrap 4, buttons only have visual indication of focus if they use `.btn` class (which uses box shadow, not outline). So Bootstrap 4 messes up the default browser behavior. :/

I think it would be reasonable to include this CSS somewhere:
```css
/* revert the bootstrap 4 rule */
button:focus:not(.focus-visible) {
  outline: revert;
}
/* re-add it so that it only applies to btn buttons */
button.btn:focus:not(.focus-visible) {
  outline: 0;
}
```

Revert only has ~94% support, but it won't break people's browsers who don't have support.

The right way to fix this issue is probably to update to Bootstrap 5.

course-v2/layouts/_default/baseof.html
```diff
@@ -42,6 +42,16 @@
...
                 <div class="col-12 p-0" id="main-course-section">
                   <div class="card-body">
+                    {{ if not $isCourseHomePage }}
+                      <button 
+                        type="button" 
+                        id="show-desktop-course-drawer" 
+                        class="show-course-info-left-arrow large-and-above-only"
+                        aria-label="Show Course Info"
+                        >
+                        <img class="left-arrow" src="/images/left_arrow.svg" alt=""/>
```
gumaerc, line 52: I was able to drag over and highlight the x icon, which looks a little odd:
image

Can we prevent this from happening with the `user-select: none;` CSS rule?

gumaerc (changes requested):
This looks good and functions well, just noticed one weird thing:

course-v2/assets/css/course-info.scss
```diff
@@ -31,3 +31,23 @@ $panel-course-info-text-color: #464646;
 .close-course-info:hover {
   background-color: #c0c0c0;
 }
+
+.show-course-info-left-arrow {
+  position: absolute;
+  background-color: #e4e4e4;
+  border: none;
+  margin-top: -18px;
```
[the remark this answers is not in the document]
gumaerc: We should definitely look into upgrading to Bootstrap 5, although I think that will be pretty involved. For now these CSS overrides will work fine I think.

course-v2/layouts/_default/baseof.html
MAbdurrehman12: Since we are replacing the div wrappers with buttons so now this should be fine

course-v2/layouts/partials/course_info.html
```diff
@@ -8,15 +8,19 @@
...
+        type="button"
+        class="btn close-course-info large-and-above-only"
+        id="{{ if $isDesktopCourseDrawer }}hide-desktop-course-drawer{{ end }}"
+        aria-label="Close Course Info"
+        >
         <img
           class="toggle navbar-toggle offcanvas-toggle"
           src="/images/close_small.svg"
           alt=""
+          width="12px"
```
MAbdurrehman12, line 21: `close_small.svg` was a bit bigger than `left_arrow.svg` hence looked inconsistent when closing/opening the drawer so reducing its size here to minimize this inconsistency

MAbdurrehman12:
@ChristopherChudzicki @gumaerc Thank you for your feedback guys. Please have another look.
Let me know if you find any issue.

ChristopherChudzicki (approved):
👍

gumaerc (approved):
:+1:

Issue #864, opened 2022-09-23
"Read More" link instead of hide/reveal arrows on CHP


pdpinch:
The up/down hide/reveal arrows are difficult to understand. Let's replace them in the course-v2 theme with a simple Read More link that expands to display all the text. 

#### Designs and Mockups

https://projects.invisionapp.com/share/NB11D8TFZX28/#/screens/469178576

#### Acceptance Criteria:

- [x] When the text overflows, end it with ... and a "Read More" Link
- [x] clicking the link should reveal the text, similar to what happens now
- [x] remove the horizontal rule 
- [x] include screenshots of desktop and mobile with the PR ([15.s21](https://live-qa.ocw.mit.edu/courses/15-s21-nuts-and-bolts-of-business-plans-january-iap-2014/) can be used for testing)

#### Out of Scope

- Are there any other places that we used the hide/reveal arrows that can be switched to "Read More" ?

pdpinch:
@gumaerc wrote [elsewhere](https://github.com/mitodl/hq/discussions/278#discussioncomment-3757517):
> Using text-overflow: ellipsis and some webpack specific CSS rules I can get multi-line truncation to work with pure CSS and the ellipsis is automatically appended to the end. The obvious pro of this is simplicity, you only need the CSS rules and then a basic show / hide button to adjust the height of the container. The main drawback is that there isn't really a polyfill for this functionality and while it works in basically all modern browsers, older browsers might not display this correctly.

Which browser aren't supported, and what's the failure mode for older browsers?

Is there a belt-and-suspenders solution where you:
1. Use truncate to set a max length in characters
2. Use the CSS rule to truncate on 3 lines for the browsers that support it

PR #330, opened 2022-01-10
feat: expand/collapse enhancement

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-hugo-themes/issues/329

#### What's this PR do?
New enhanced design for expanding and collapsing course description, course meta-data and image description.

#### How should this be manually tested?
- Verify that image description, course description and course info (meta-data) expands and collapse correctly when arrow icon is clicked/toggled.
- Change the size of the content or test them on different courses having different sizes/lengths of content.
- Test the above on multiple browsers and multiple screen sizes.

#### Screenshots (if appropriate)
Video link: https://drive.google.com/file/d/1R8yhyXMAMVmjZn9dphS0kjAjpxwbkwm2/view?usp=sharing

PR #876, opened 2022-09-29
show more / show less for course description

gumaerc:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/864

#### What's this PR do?
In https://github.com/mitodl/ocw-hugo-themes/pull/330 we added some functionality to add a collapse / expand arrow to the course description using a JS / CSS based solution. This ended up being confusing for users, and a new approach (invision design link in the issue above) to use an inline link with the text "Show more" at the end of truncated text. This PR implements that using Hugo's [`truncate`](https://gohugo.io/functions/truncate/) function to truncate the description if it's over 320 characters long. This was the character threshold determining whether the expand / collapse should be initialized on the description in the CSS based method, so I stuck with that. Both the truncated and full description (with expand / collapse links) are written to the page with the expanded description hidden at first. Clicking the links simply toggles the `d-none` class on both the descriptions, effectively toggling them.

 - Clone https://github.mit.edu/mitocwcontent/15.s21-january-iap-2014
 - Point to it by setting the following in your `.env.`:
```
COURSE_CONTENT_PATH=/path/to/mitocwcontent/
OCW_TEST_COURSE=15.s21-january-iap-2014
```
 - Run `npm run start:course`
 - Visit http://localhost:3000/
 - Click the "Show more" link on the course description and confirm that the description is expanded
 - Click the "Show less" link and ensure that the description collapses
 - Go into the course content repo, open the `data/course.json` file and edit `course_description` to a short string like "test"
 - Verify that the short description loads and that there are no show / hide links
 - Repeat the above for as many browsers as you have access to and simulate mobile as well

image
image
image
image

### DIFF
--- a/base-theme/assets/css/button.scss
+++ b/base-theme/assets/css/button.scss
@@ -45,3 +45,7 @@ a.red-btn {
   cursor: pointer;
   white-space: nowrap;
 }
+
+button.btn-link.text-black {
+  color: black;
+}
--- a/base-theme/assets/css/button.scss
+++ b/base-theme/assets/css/button.scss
@@ -46,6 +46,7 @@ a.red-btn {
   white-space: nowrap;
 }
 
-button.btn-link.text-black {
-  color: black;
+button.btn-link {
+  color: inherit;
+  vertical-align: baseline;
 }
--- a/base-theme/assets/css/button.scss
+++ b/base-theme/assets/css/button.scss
@@ -48,5 +48,6 @@ a.red-btn {
 
 button.btn-link.link-button {
   color: inherit;
+  text-decoration: underline;
   vertical-align: baseline;
 }
--- a/base-theme/assets/css/button.scss
+++ b/base-theme/assets/css/button.scss
@@ -46,7 +46,7 @@ a.red-btn {
   white-space: nowrap;
 }
 
-button.btn-link {
+button.btn-link.link-button {
   color: inherit;
   vertical-align: baseline;
 }
--- a/course-v2/assets/js/course_expander.ts
+++ b/course-v2/assets/js/course_expander.ts
@@ -43,11 +43,19 @@ const initCourseInfoExpander = document => {
 const initCourseDescriptionExpander = document => {
   const courseDescription = document.getElementById("course-description")
   if (courseDescription) {
-    const collapsedDescription = courseDescription.querySelector("#collapsed-description")
-    const expandedDescription = courseDescription?.querySelector("#expanded-description")
+    const collapsedDescription = courseDescription.querySelector(
+      "#collapsed-description"
+    )
+    const expandedDescription = courseDescription?.querySelector(
+      "#expanded-description"
+    )
     if (collapsedDescription && expandedDescription) {
-      const expandLink = collapsedDescription.querySelector("#expand-description")
-      const collapseLink = expandedDescription.querySelector("#collapse-description")
+      const expandLink = collapsedDescription.querySelector(
+        "#expand-description"
+      )
+      const collapseLink = expandedDescription.querySelector(
+        "#collapse-description"
+      )
       expandLink.addEventListener("click", () => {
         collapsedDescription.classList.add("d-none")
         expandedDescription.classList.remove("d-none")
--- a/course-v2/assets/js/course_expander.ts
+++ b/course-v2/assets/js/course_expander.ts
@@ -45,8 +45,8 @@ const initCourseDescriptionExpander = container => {
   const collapsedDescription = courseDescription.find("#collapsed-description")
   const expandedDescription = courseDescription.find("#expanded-description")
   if (collapsedDescription && expandedDescription) {
-    const expandLink = collapsedDescription.find("a")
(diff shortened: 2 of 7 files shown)


gumaerc:
@pdpinch One thing to keep in mind with this method is that since `truncate` runs in the Hugo build and not on the client's browser, the "split point" is pre-determined by a number of characters. Therefore, there is no way to guarantee that the "Show more" link will end up at the very end of a line every time. As you can see in the screenshots, it will appear in a different spot depending on window width / DPI / client zoom / text wrapping / etc.

pdpinch:
I left a [comment on the issue](https://github.com/mitodl/ocw-hugo-themes/issues/864#issuecomment-1262983305) because I want to understand the options better:

course-v2/layouts/partials/course_description.html
```diff
@@ -3,21 +3,25 @@
 {{ with $courseData.course_description }}
   {{ $shouldCollapseDescription = gt (len .) 320 }}
 {{ end }}
-
-<div class="course-description">
+{{ $truncatedDescription := $courseData.course_description }}
+{{ $expandedDescription := print $courseData.course_description " [Show less](#collapse_description)" }}
+{{ if $shouldCollapseDescription }}
+  {{ $truncatedDescription = truncate 320 "...[Show more](#expand_description)" $courseData.course_description }}
```
ChristopherChudzicki, line 9: **Requested Change:** IMO "Show more" / "Show less". should but a button styled as a link, not a link. I realize this is a little bit of a pain to style, but we might want such link-styled buttons else. Maybe a link_button shortcode or partial would be useful? Also, using a link has consequences for the URL:


https://user-images.githubusercontent.com/9010790/193831417-8ea04d0f-dcce-4361-ad25-a37379ac8a5e.mov



_If we do need to use a link, there are probably ways to work around the URL issue._

ChristopherChudzicki (changes requested):
Looks good. One request around links vs buttons, and one suggestion.

gumaerc:
@ChristopherChudzicki back to you for another look

base-theme/assets/css/button.scss
```diff
@@ -45,3 +45,7 @@ a.red-btn {
   cursor: pointer;
   white-space: nowrap;
 }
+
+button.btn-link.text-black {
+  color: black;
+}
```
ChristopherChudzicki, line 51: **Requested Change:** We also want `vertical-align: baseline`, which will align the button text with surrounding text.

**Suggestion:** Since the "textual" buttons might not always be black, I think it would be better to ditch the color in class name and instead inherit the color from parent:
```css
button.btn-link.text-button {
  color: inherit;
  vertical-align: baseline;
}
```

course-v2/assets/js/course_expander.ts
```diff
@@ -40,4 +40,22 @@ const initCourseInfoExpander = document => {
   }
 }
 
-export { initCourseInfoExpander }
+const initCourseDescriptionExpander = document => {
+  const courseDescription = document.getElementById("course-description")
+  const collapsedDescription = courseDescription.querySelector("#collapsed-description")
+  const expandedDescription = courseDescription?.querySelector("#expanded-description")
```
ChristopherChudzicki, line 46: Are we safe to assume `courseDescription` always exists?

I stand by the "ditch jquery" suggestion, but I had forgotten (until workin on #887) that the jquery versions are a little more forgiving if things don't exist.

Also: ... Typescript should have caught this issue, except there's a `ts-nocheck` at the top of the file. Not necessarily requesting you change it as part of this PR, but we should get rid of it.

ChristopherChudzicki (changes requested):
One styling request, one question.

base-theme/assets/css/button.scss
gumaerc: That makes sense, I just set these on `button.btn-link` though as that's the bootstrap style for link buttons. If you don't do this Bootstrap will try and set it to bootstrap alert style colors.

gumaerc:
@ChristopherChudzicki Thanks, I think I addressed your concerns.

ChristopherChudzicki (approved):
@gumaerc A few questions for you
- I may have been wrong to suggest `color: inherit`... We seem to use black links everywhere, which I did not realize, and our text is not black... it's some lighter color. What do you think?
- Should the button text be unerlined? Thats how our normal links are, right?

And also....I know I suggested a `<button />`, but maybe an anchor with `role="button"` would be good enough. We'd have an easier time styling it like our other in-line links.

apologies for waffling =/

gumaerc:
@ChristopherChudzicki Good call on the underline style, I added that. As far as the color, the design shows it all the same color but also shows all the text as pure black. I'll leave that as a design question for later. If you look back at the original issue, I discussed our options with @pdpinch and we're going to stick with the Hugo based `truncate` solution for now and see if we like it. We can revisit this later if we want to re-work it.

Issue #877, opened 2022-09-30
update mobile design: menu button and more info link


pdpinch:
#### Designs and Mockups

See https://projects.invisionapp.com/d/main?origin=v7#/console/21537330/469289599/preview?scrollOffset=2287

#### Acceptance Criteria:

- [ ] replace "browse materials" button with "menu"
- [ ] replace "course info" tab with "more info" link

PR #885, opened 2022-10-04
style: menu button & more info link mobile design

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/ocw-hugo-themes/issues/877

#### What's this PR do?
- Updates buttons for mobile design for course-v2.

- Checkout this branch
- Set screen size of tablet or mobile.
- Verify that `Menu` button and `More Info` link are according to the design provided in issue ticket.
- Verify that both work as expected.
- Verify that they are hidden on large screens.
- Verify that they are completely responsive.
- Verify that there is no change/affect/impact on course-v1 theme.
- Any other use-case that needs to be tested. 

image

image

### DIFF
--- a/course-v2/layouts/partials/course_info_toggle.html
+++ b/course-v2/layouts/partials/course_info_toggle.html
@@ -1,10 +1,10 @@
 {{ $isCourseHomePage := .isCourseHomePage | default "" }}
 
 {{ if not $isCourseHomePage }}
-  <button class="btn btn-link float-right">
-    <a href="javascript:void(0)" class="mobile-course-info-toggle-btn toggle navbar-toggle offcanvas-toggle"
-      data-toggle="offcanvas" data-target="#course-info-drawer">
-      More Info
-    </a>
+  <button 
+    class="btn btn-link float-right mobile-course-info-toggle-btn toggle navbar-toggle offcanvas-toggle" 
+    data-toggle="offcanvas" 
+    data-target="#course-info-drawer">
+    More Info
   </button>
 {{ end }}
--- a/course-v2/layouts/partials/course_info_toggle.html
+++ b/course-v2/layouts/partials/course_info_toggle.html
@@ -1,0 +1,10 @@
+{{ $isCourseHomePage := .isCourseHomePage | default "" }}
+
+{{ if not $isCourseHomePage }}
+  <div class="d-flex justify-content-end">
+    <a href="javascript:void(0)" class="mobile-course-info-toggle-btn toggle navbar-toggle offcanvas-toggle"
+      data-toggle="offcanvas" data-target="#course-info-drawer">
+      More Info
+    </a>
+  </div>
+{{ end }}
--- a/course-v2/layouts/partials/course_info_toggle.html
+++ b/course-v2/layouts/partials/course_info_toggle.html
@@ -1,10 +1,10 @@
 {{ $isCourseHomePage := .isCourseHomePage | default "" }}
 
 {{ if not $isCourseHomePage }}
-  <div class="d-flex justify-content-end">
+  <button class="btn btn-link float-right">
     <a href="javascript:void(0)" class="mobile-course-info-toggle-btn toggle navbar-toggle offcanvas-toggle"
       data-toggle="offcanvas" data-target="#course-info-drawer">
       More Info
     </a>
-  </div>
+  </button>
 {{ end }}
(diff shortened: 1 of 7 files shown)


course-v2/layouts/partials/course_info_toggle.html
```diff
@@ -0,0 +1,10 @@
+{{ $isCourseHomePage := .isCourseHomePage | default "" }}
+
+{{ if not $isCourseHomePage }}
+  <div class="d-flex justify-content-end">
```
ChristopherChudzicki, line 4: **Requested Change:** A button element seems more semantic here than a div-wrapping anchor. 

Same comment below for the menu button (which certainly visually looks like a button).

ChristopherChudzicki (changes requested):
One request about using button elements.

ChristopherChudzicki:
_Responding because I almost included this in the review, then decided it was out-of-scope for the button issue..._

Screen Shot 2022-10-04 at 12 08 00 PM Screen Shot 2022-10-04 at 12 08 05 PM

@pdpinch those are the drawers as they currently are. Two thoughts:
- we should add a "Close" (or "X") button when the drawer is open. Right now they're only dismissable (w/o navigation) via tappinga  fairly narrow portion of the screen.
- We might want to set a max-width on the drawer. They're a bit wide on tablets:

Screen Shot 2022-10-04 at 12 11 33 PM

course-v2/layouts/partials/course_info_toggle.html
MAbdurrehman12: I think this `More Info` is supposed to be a link rather than a button, and wrapping it up with a button will also require some extra styling to hide/remove the button background, border, etc (since button has its own default styles)
But I agree with the `Menu`, it should be a button. 

So I think we can keep `More Info` as it is since it's a link.
and we can replace `div` with `button` for `Menu`.

What do you think?

MAbdurrehman12:
> we should add a "Close" (or "X") button when the drawer is open. Right now they're only dismissable (w/o navigation) via tappinga fairly narrow portion of the screen.

> We might want to set a max-width on the drawer

Yes, I agree. I also had this in mind.
Thank you for pointing it out.
Maybe @pdpinch can also share his thoughts on this

course-v2/layouts/partials/course_info_toggle.html
ChristopherChudzicki: My inclination is that both "Menu" and "More info" perform similar actions (open a drawer) and neither one changes the URL, so they should both be buttons not links. The fact that one is styled differently from the other shouldn't really affect the tag / role. 

As far as styling, you can make a button look like a link with Bootstrap's `btn btn-link` class. https://getbootstrap.com/docs/4.0/components/buttons/#examples  _Another option would be the use `role="button"` on a link, but using an actual button element is probably better if we want a button role._

For reference...[WAI-ARIA says](https://www.w3.org/TR/wai-aria-1.1/#link)
> NOTE
> If pressing the link triggers an action but does not change browser focus or page location, authors are advised to consider using the button role instead of the link role.

Which...isn't really the clearest, imo. In this case the page location (which I think means URL) does not change. I guess maybe browser focus changes? I'm not sure. Anyway, "Does the URL change" seems like a good test to me.

ChristopherChudzicki (changes requested):
Sending this back once more, but if the button reasoning doesn't sway you, then I'll defer to your judgement 😄

course-v2/layouts/partials/course_info_toggle.html
MAbdurrehman12: I think this makes sense, you're right!
Let's wrap this up with button too

MAbdurrehman12: Done.
Please have another look

ChristopherChudzicki (approved):
Agree with peter. It looks great! Thanks for fixing up the tags.

MAbdurrehman12:
> Yes. Do we have a design for a "X" button that we can repurpose?

We had a design for closing course info 
image
For these drawers, I think we can place the close (x) icon at the right and left most sides, something like this:
image

What do you and @ChristopherChudzicki think?

> I agree, but I think that's out of scope. @MAbdurrehman12 can you open an issue for that?

Sure

> I agree. I think maybe half the width in the screenshot would be good, but I defer to your judgement @MAbdurrehman12

I agree, let's set the max-width for these drawers.

Another quick question: Should we place close (x) icons and set max-width in this same PR or have another follow-up PR?

MAbdurrehman12:
I have opened issues for the things discussed here.
Let's create separate PRs for the stuff discussed above according to the issues opened and merge this one

PR #887, opened 2022-10-04
Fix table code erroring on non-table pages

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

#875 followup to #844 

#### What's this PR do?
This PR:
- fixes an issue with `responsive_tables.js` where code was assuming that every page had at least one table.
- fixes an issue where only the first table on each page was respnsive
- converts course_info_toggle.js and responsive_tables.js to typescript

First, visit any page on ocw.mit.edu that does *not* have a table. Observe the console errors.

Now on this branch using the original course theme:
1. Visit any page without a table. Check that there are no errors.
2. Now check a page with large (wide) tables...
3. Open the content markdown for any page. Replace it with https://gist.github.com/ChristopherChudzicki/dd0d177987b2baf219dfb9b2f351a097
4. View the hugo build for that page. Check that clicking "Show course info / hide course info" expands/collapses the tables appropriately.

https://user-images.githubusercontent.com/9010790/193902339-fc545751-f681-49e5-a4d8-8087483961bb.mov


course/assets/js/course_info_toggle.ts
```diff
@@ -29,30 +31,37 @@ export const initDesktopCourseInfoToggle = () => {
 }
 
 function calculateTableResponsiveness() {
-  const table = document
-    .getElementById("main-content")
-    .getElementsByTagName("table")[0]
-  const mainContentWidth = document.getElementById("main-content").clientWidth
-  if (table.clientWidth > mainContentWidth) {
```
ChristopherChudzicki, line 36: Two issues here:
1. This was running whether or not the table existed. So every non-table page was throwing an error.
2. Also, we were only being responsive for the first table on each page.

Fix is:
- be responsive for all tables on the page (all might be zero).

We should use typescript. Typescript would have caught this error. (At least error 1)

course/assets/js/course_info_toggle.ts
```diff
@@ -11,6 +10,9 @@ export const initDesktopCourseInfoToggle = () => {
       const desktopCourseInfoToggle = document.getElementById(
         "desktop-course-info-toggle"
       )
+      if (!(desktopCourseInfoToggle && desktopCourseInfo && mainContent)) {
+        return
+      }
```
ChristopherChudzicki, line 15: I'm pretty sure `mainContent` should always exist; the other two do not exist on the homepage, though this code still seems to run.

gumaerc (approved):
:+1: LGTM

Issue #881, opened 2022-09-30
local errors being logged to sentry


MAbdurrehman12:
### Steps to Reproduce

- Local errors logged to sentry: https://sentry.io/organizations/mit-office-of-digital-learning/issues/3630699848/?project=5304953&query=is%3Aunresolved
- To manually reproduce:
  - Try to break something on your local setup
      -  1 idea: Pass null/0 to this new courses template [here](https://github.com/mitodl/ocw-hugo-themes/blob/74ba4b7953548bf21b9549555c72b9befd8e01da/www/layouts/home.html#L94) OR manually throw an error in the console.
  - Now note that the new courses API has failed you have some console errors. 
  - Check the network tab that these errors have been logged to sentry (sentry API calls can been seen in network)
  - Go to sentry and see your local errors.

### Expected Behavior

- Local errors shouldn't be logged to sentry

### Actual Behavior

- Local errors are being logged to sentry

### Screenshot or Screencast

<img width="1201" alt="image" src="https://user-images.githubusercontent.com/93309234/193275229-68c7e94f-08c0-4fcd-aba4-453f34f49ce0.png">

PR #903, opened 2022-10-10
fix: sentry DSN configured via env

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/ocw-hugo-themes/issues/881

#### What's this PR do?
- Configures Sentry DSN from env.
- If `SENTRY_DSN` env is not found or not set, then exceptions are ignored/not logged to sentry.

- Checkout this branch
- Generate errors locally (how to do it? Please check the description of the issue ticket)
- Verify that these errors are not logged to Sentry.
- Now set `SENTRY_DSN` in your .env 
- Rebuild.
- Repeat the above steps to generate errors locally.
- Verify that these errors are logged to Sentry.
- Test any other use case that needs to be tested in order to verify that these changes have no side-effects

### DIFF
--- a/base-theme/assets/js/sentry.ts
+++ b/base-theme/assets/js/sentry.ts
@@ -1,11 +1,9 @@
 import * as Sentry from "@sentry/browser"
 
 export const initSentry = (): typeof Sentry => {
-  const sentryDSN = process.env.SENTRY_DSN
   Sentry.init({
-    release:    RELEASE_VERSION,
-    dsn:        sentryDSN,
-    beforeSend: event => (sentryDSN ? event : null)
+    release: RELEASE_VERSION,
+    dsn:     process.env.SENTRY_DSN
   })
   return Sentry
 }
--- a/base-theme/assets/js/sentry.ts
+++ b/base-theme/assets/js/sentry.ts
@@ -3,8 +3,7 @@ import * as Sentry from "@sentry/browser"
 export const initSentry = (): typeof Sentry => {
   Sentry.init({
     release: RELEASE_VERSION,
-    dsn:
-      "https://eee58f41dda54d2b814296e12dced4b7@o48788.ingest.sentry.io/5304953"
+    dsn:     SENTRY_DSN
   })
   return Sentry
 }
--- a/base-theme/assets/js/sentry.ts
+++ b/base-theme/assets/js/sentry.ts
@@ -1,10 +1,11 @@
 import * as Sentry from "@sentry/browser"
 
 export const initSentry = (): typeof Sentry => {
+  const sentryDSN = process.env.SENTRY_DSN
   Sentry.init({
-    release: RELEASE_VERSION,
-    dsn:
-      "https://eee58f41dda54d2b814296e12dced4b7@o48788.ingest.sentry.io/5304953"
+    release:    RELEASE_VERSION,
+    dsn:        sentryDSN,
+    beforeSend: event => (sentryDSN ? event : null)
   })
   return Sentry
 }
--- a/base-theme/assets/js/sentry.ts
+++ b/base-theme/assets/js/sentry.ts
@@ -1,6 +1,6 @@
 import * as Sentry from "@sentry/browser"
 
-export const initSentry = () => {
+export const initSentry = (): typeof Sentry => {
   Sentry.init({
     release: RELEASE_VERSION,
     dsn:
--- a/base-theme/assets/js/sentry.ts
+++ b/base-theme/assets/js/sentry.ts
@@ -3,7 +3,8 @@ import * as Sentry from "@sentry/browser"
 export const initSentry = (): typeof Sentry => {
   Sentry.init({
     release: RELEASE_VERSION,
-    dsn:     SENTRY_DSN
+    dsn:
+      "https://eee58f41dda54d2b814296e12dced4b7@o48788.ingest.sentry.io/5304953"
   })
   return Sentry
 }
--- a/base-theme/assets/js/utils.ts
+++ b/base-theme/assets/js/utils.ts
@@ -1,5 +1,3 @@
-declare var isHugoDevServer: boolean
-
 export const setLocalStorageItem = (key: string, value: any) => {
   return setOrGetLocalStorageItemHelper("set", key, value)
 }
@@ -49,4 +47,4 @@ export const isLocalServerRunning = (): boolean =>
   location.hostname === "localhost" ||
   location.hostname === "127.0.0.1" ||
   location.hostname === "" ||
-  isHugoDevServer
+  IS_HUGO_DEV_SERVER
--- a/base-theme/assets/js/utils.ts
+++ b/base-theme/assets/js/utils.ts
@@ -1,3 +1,5 @@
+declare var isHugoDevServer: boolean
+
 export const setLocalStorageItem = (key: string, value: any) => {
   return setOrGetLocalStorageItemHelper("set", key, value)
 }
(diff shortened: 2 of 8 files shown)


base-theme/assets/js/utils.ts
```diff
@@ -39,3 +39,11 @@ const setOrGetLocalStorageItemHelper = (
     return false
   }
 }
+
+/**
+ * @return {boolean} returns true is app is running on local server, else false.
+ */
+export const isLocalServerRunning = () =>
```
gumaerc, line 46: This will be okay to cover most cases, but a potential edge case could exist if someone has decided to add a `hosts` entry for something like `127.0.0.1 ocw.local`. Something we could add here to make this more robust would be to use the Hugo variable [`site.IsServer`](https://gohugo.io/variables/site/). We don't template our JS files through Hugo, so we wouldn't be able to pull it directly into this function, but we could inline something on the base template like:

```
    <script type="text/javascript">
    function isHugoDevServer() {
      return {{ site.IsServer }}
    }
    </script>
```
We could then add an extra line to the or statements here like:
```
  location.hostname === "localhost" ||
  location.hostname === "127.0.0.1" ||
  location.hostname === "" ||
  isHugoDevServer()
```

This wouldn't cover sites built through the Minio based local dev setup in `ocw-studio` because that all happens through nginx. We could maybe do something to detect that in the future, but I think checking `site.IsServer` is enough of an easy win here that we should do it.

gumaerc (changes requested):
This looks good, but I have a suggestion:

base-theme/assets/js/sentry.js
```diff
@@ -1,10 +1,14 @@
...
-  Sentry.init({
-    dsn:
-      "https://eee58f41dda54d2b814296e12dced4b7@o48788.ingest.sentry.io/5304953"
-  })
-  return Sentry
+  if (!isLocalServerRunning()) {
+    Sentry.init({
+      dsn:
+        "https://eee58f41dda54d2b814296e12dced4b7@o48788.ingest.sentry.io/5304953"
+    })
```
ChristopherChudzicki, line 10: Sentry has a [`beforeSend`](https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-platformidentifier-namebefore-send-) configuration option that can be used to discard the event:
> ...  If you return null, the event will be discarded.

IMO, it would be a little better to do
```ts
beforeSend: event => isLocalServerRunning() ? null : event
```
because that way `window.Sentry` will be a sentry object in all environments—we won't need modifications to `base-theme/assets/index.ts` or  `base-theme/layouts/partials/sentry_capture_message.html`

ChristopherChudzicki (changes requested):
I was looking at sentry stuff today for another reason... One small request, unless you disagree.

Also, we should convert the sentry file to typescript.

base-theme/assets/js/sentry.js
MAbdurrehman12: Nice suggestion
Thanks! @ChristopherChudzicki

ChristopherChudzicki (approved):
👍

MAbdurrehman12:
btw, sometimes we need to deliberately log errors to sentry from our local setup, but once this PR is merged then we might not be able to do that without making a change in `sentry.ts` file.
So do we need to document this somewhere?

ChristopherChudzicki:
> btw, sometimes we need to log errors to sentry deliberately from our local setup, but once this PR is merged then we might not be able to do that without making a change in sentry.ts file.
So do we need to document this somewhere?

Personally, I think it's ok to leave it out, but others might disagree. I'm always dubious that adding stuff to an already-long readme makes things more discoverable than just searching the JS code for sentry. If you're working on sentry, you're probably aware of Sentry.init 🤷 

None of {Studio, Open Discussions, MITxOnline} mention sentry in their readmes and I suspect they all supress local sentry errors. Studio does, certainly. (Although it's done differently... controlling the DSN via an env variable which people usually omit in their local setups.)

@gumaerc @MAbdurrehman12 what do you think?

pdpinch:
Instead of omitting the error if it's a local server, is there a way to only capture the errors if they are from: 

ocw.mit.edu => Sentry production environment
live-qa.mit.edu => Sentry QA environment

MAbdurrehman12:
> Instead of omitting the error if it's a local server, is there a way to only capture the errors if they are from:
ocw.mit.edu => Sentry production environment
live-qa.mit.edu => Sentry QA environment

yeah, I think we can place a filter on `location.hostname` (i.e: domain) and log errors only if they are any of the following.

ChristopherChudzicki:
> Instead of omitting the error if it's a local server, is there a way to only capture the errors if they are from:
>
> ocw.mit.edu => Sentry production environment
> live-qa.mit.edu => Sentry QA environment

@MAbdurrehman12 would it be better to just set the sentry `{ dsn, env }` config properties from env variables configured in RC / prod? Then people can override them if they want locally. (*I think my earlier comment about "people might leave them in local config accidentally" is not that big a deal, and env variable seems better than inspecting hosts.*)

If we go env variable route, I think we'd need to insert them with webpack.DefinePlugin like here: https://github.com/mitodl/ocw-hugo-themes/pull/907/files#diff-de5d52c1ed871f53248e58f559733c95f29131fbaba9e5653eda8746ae6a9c7dR146

MAbdurrehman12:
> would it be better to just set the sentry { dsn, env } config properties from env variables configured in RC / prod? 

Yup, sounds good!

MAbdurrehman12:
@pdpinch Shall we follow this approach then? 
i.e: Configure sentry DSN via env variable and set the env variable only on:
- ocw.mit.edu
- live-qa.mit.edu

gumaerc:
@MAbdurrehman12 @ChristopherChudzicki On the topic of env variables, I think we didn't originally go with that for client side sentry in our Hugo build because the value is set inside a JS file that's part of the webpack build and isn't templated by Hugo. However, nowadays [Hugo Pipes](https://gohugo.io/hugo-pipes/resource-from-template/) potentially offers a way to do this. It allows you to treat anything in the `assets` folder as a Go template and execute "template magic" against it. There would also need to be a change in the config in `ocw-hugo-projects` under `security.funcs.getenv` to allow the specific env variable you reference.

ChristopherChudzicki (changes requested):
Arg! I forgot to click "Submit" on this review yesterday. Sorry for the delay.

Looks good. One small request.

base-theme/assets/js/sentry.ts
```diff
@@ -1,10 +1,11 @@
...
-export const initSentry = () => {
+export const initSentry = (): typeof Sentry => {
+  const sentryDSN = process.env.SENTRY_DSN
   Sentry.init({
-    release: RELEASE_VERSION,
-    dsn:
-      "https://eee58f41dda54d2b814296e12dced4b7@o48788.ingest.sentry.io/5304953"
+    release:    RELEASE_VERSION,
+    dsn:        sentryDSN,
+    beforeSend: event => (sentryDSN ? event : null)
```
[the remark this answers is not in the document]
gumaerc: I can confirm trying it without the `beforeSend` is fine. Node renders this as `var sentryDSN = "MISSING_ENV_VAR".SENTRY_DSN;` which would make me think that it would throw a connection error or something, but even after removing the `beforeSend` I didn't have any errors in my console.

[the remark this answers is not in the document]
MAbdurrehman12: beforesend removed. Please have a look

ChristopherChudzicki (approved):
👍 Looks good...do we a plan for getting the env vars into....the concourse envs? *I think that's where they need to go... @gumaerc ?*

gumaerc:
@ChristopherChudzicki @MAbdurrehman12 The env var needs to go in `ocw-studio` so it can be templated into pipelines upserted by the management commands there.

PR #910, opened 2022-10-13
add footer to seach page

ChristopherChudzicki:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

See https://mitodl.slack.com/archives/GQ89D0Z5M/p1665687170526699?thread_ts=1665685641.605519&cid=GQ89D0Z5M

#### What's this PR do?
This PR sets a max height for the search area of 150% of user's view port.  for the search area of 150% of the user's viewport height.

1. Run `yarn run start:www`
2. Visit the search page
3. Check that you can (a) infinite scroll and also (b) scroll to the footer, even without viewing a bazillion results.
4. Check that if you load the page zoomed out, search results still load.

    *To see what I mean, go to https://ocw.mit.edu/search/, zoom waaay out so you can see 10+ search results on one page on the screen, and reload. That's not very realistic, but this PR exacerbates the issue, so I addressed it here.*

#### Screenshots

https://user-images.githubusercontent.com/9010790/195717607-b2a35927-49a6-4014-a84e-f9f4a3675e99.mov

Mobile view is unchanged. It's significantly harder to scroll to the bottom without exhausting search results on mobile, but we can live with that for now.

### DIFF
--- a/www/assets/js/components/SearchPage.tsx
+++ b/www/assets/js/components/SearchPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from "react"
+import React, { useState, useCallback, useRef, useEffect } from "react"
 import InfiniteScroll from "react-infinite-scroller"
 import {
   Aggregations,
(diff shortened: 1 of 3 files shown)


www/assets/js/components/SearchPage.tsx
```diff
@@ -157,6 +157,37 @@ export default function SearchPage(props: SearchPageProps) {
...
+   * then no new request is made.
+   *
+   * This is a work-around to address that.
+   *
+   * This returns a ref to be attached to the child of InfiniteScroll that fixes
+   * the issue.
+   *
+   * See https://github.com/danbovey/react-infinite-scroller/issues/294
+   */
+  const checkIfNeedMore = useCallback(() => {
```
ChristopherChudzicki, line 174: We have an issue on OCW before this PR where
- IF search page fits 10+ results on initial load (or after resize)
- THEN new results are never loaded.

That's not very realistic (you have to zoom waaaaay out to make it happen on ocw currently).

But this PR exacerbates the issue because no longer do 10 items need to fit on the visible search area at once, 10 items just need to fit within 150% of the viewport height at once (see CSS changes above). 

So I decided to address it.

We could also increase the search page size a bit, which would make the workaround here less relevant.

collinpreston (approved):
Looks good!  Tested and this works for me.  Oops now I see the "needs review" tag wasn't actually on here.

ChristopherChudzicki:
@collinpreston Well, thanks for looking :) Yeah, we ended up deciding to do something different with the search footer.

Issue #919, opened 2022-10-18
Integration Testing for OCW Courses


ChristopherChudzicki:
As a developer, I'd like to be able to add tests when I fix a bug to have confidence future PRs due not regress the behavior. 

_ocw-hugo-themes currently has some limited tests for JS code, but nothing that renders a course. See https://github.com/mitodl/hq/discussions/45 for more background_

#### Acceptance Criteria:

- [ ] A CI job that
    - builds a course with hugo
    - makes assertions about course behavior (See https://github.com/mitodl/hq/discussions/45 for some examples)

#### Out of Scope

- Not going to do

ChristopherChudzicki:
Things to test:
- #924 
- (no) extra spaces around resource links

PR #921, opened 2022-10-19
update to yarn 3.2.2

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested *Insofar as...tests still pass.*

On the road toward #919 

#### What's this PR do?
This PR updates ocw-hugo-themes to modern yarn. (In yarn's lingo, I believe yarn v1 is "yarn" and higher major versions, v2, and v3 right now, are "yarn berry" 🤷).

This brings ocw-hugo-themes in line with open discussions, ocw-studio, and mitxonline, and generally seems good. (*My immediate motivation is possibly using yarn workspaces toward #919; workspaces have better support in yarn berry.)

1. yarn install
2. yarn start:course
3. yarn start:www

They should work.


.github/workflows/deploy.yml
```diff
@@ -41,7 +41,7 @@ jobs:
             ${{ runner.os }}-yarn-
 
       - name: Install dependencies
-        run: yarn install --frozen-lockfile --ignore-engines --prefer-offline
+        run: yarn install --immutable
```
ChristopherChudzicki, line 44: This is what Studio uses. 

immutable replaces frozen-lockfile;  --ignore-engines and --prefer-offline are no longer necessary.

ChristopherChudzicki (commented):
The enormous lockfile diff is because yarn changed their lockfile format

package.json
```diff
@@ -5,7 +5,7 @@
   "repository": "mitodl/ocw-hugo-themes",
   "main": "base-theme/assets/index.js",
   "scripts": {
-    "lint": "eslint **/assets",
+    "lint": "eslint {base-theme,course,course-v2,fields,util,www}",
```
ChristopherChudzicki, line 8: This sort of annoys me and boggles my understanding.

After updating to yarn v3, `npm run lint` worked fine, but `yarn run lint` complained about no matches in `node_modules/@jest/core/build/assets`. Eslint should not have been touching those anyway, since eslint ignores node_modules by default. It boggles my mind that `run lint` worked with `npm` but not `yarn`.

Most (all?) of our other repos specify specific directories to lint. Switching to that here fixed the issue. (You can check things are still being linted by adding an unused variable `const cat = "meow"` to some js file.)

I find this a little annoying, but I don't think it should stop us from updating to yarn 3. I'm also looking to update eslint at some point soon, and maybe that will help resolve this, too 🤷

gumaerc (approved):
👍 LGTM

MAbdurrehman12:
@ChristopherChudzicki theme assets [pipeline](https://cicd-qa.odl.mit.edu/teams/ocw/pipelines/ocw-theme-assets/jobs/build-theme-assets/builds/127?vars.branch=%22release-candidate%22) is failing due to 
```
selected worker: ip-10-1-1-45
+ cd ocw-hugo-themes
+ yarn install --pure-lockfile
Unknown Syntax Error: Unsupported option name ("--pure-lockfile").

$ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--mode #0]
```

Is it relevant to this PR?

PR #927, opened 2022-10-24
Update webpack devserver; convert webpack conifgs to ts

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing

On the road to #919 

#### What's this PR do?
This PR updates webpack-dev-server to version 4 and changes the configuration files from js to typescript, because then we get waaaay better autocomplete on webpack stuff.

Run `yarn start:course`, `yarn start:www`, yarn `build:webpack` locally. They should work.

I think the test setup for #919 will want to run webpack programmatically (ie, not through CLI). The docs I found easily for that were for webpack-dev-server v4. Rather than finding the old docs, I figure let's update the dependency.


base-theme/assets/webpack/webpack.prod.ts
```diff
@@ -20,7 +20,7 @@ module.exports = merge(common, {
       new TerserPlugin({
         parallel:      true,
         terserOptions: {
-          ecma: 6
+          ecma: 5
```
ChristopherChudzicki, line 23: When converting the config files to typescript, I discovered that `6` is not a valid valid for this configuration. The valid values are:
```
export type ECMA = 5 | 2015 | 2016 | 2017 | 2018 | 2019 | 2020;
```
This configuration option controls what sort of optimizations the terser plugin will do. The higher ES version you specify, the more optimizations it does, but the optimizations might be valid in fewer browsers. For example, the terser docs say that specifying 2015 can turn `{ x: x }` into `{ x }`. (I couldn't actually get this optimization to happen in their [playground](https://try.terser.org/), but whatever.)

Anyway, 6 is invalid. (The thing people call ES6 is formally ES2015).

5 is the default, which is what I suspect was being used before.

pt2302:
Not running currently, @ChristopherChudzicki will push up another commit.

www/assets/css/home.scss
```diff
@@ -381,7 +381,7 @@ $searchbox-height: 50px;
     width: 100%;
     background-size: cover !important;
     display: flex;
-    align-items: end;
+    align-items: flex-end;
```
ChristopherChudzicki, line 384: We were getting warnings about this issue previously. (`"end" has mixed support... use "flex-end" instead.`)

The new version of dev server cares more (won't load page). Hence fixing it here.

www/assets/css/learning-resource-card.scss
```diff
@@ -83,7 +83,7 @@
       }
 
       .cover-image-resource {
-        width: fill;
+        width: stretch;
```
ChristopherChudzicki, line 86: Same as https://github.com/mitodl/ocw-hugo-themes/pull/927/commits/b19c0b38bf14f38bf20d1854320bc214c60c55ca#r1005766005, but a different css warning.

pt2302 (approved):
This looks good to me!

One warning that comes up, but is not impacting functionality:
```
Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db
```

Issue #895, opened 2022-10-06
course menu and more info drawer closing by swiping and x button


MAbdurrehman12:
As a mobile user, I'd like to close the `course menu drawer` and `course info drawer` either by swiping or by clicking the close (x) button.

#### Designs and Mockups

None.
[Reference](https://github.com/mitodl/ocw-hugo-themes/pull/885#issuecomment-1268451604)

#### Acceptance Criteria:

- [ ] Make the `course menu drawer` swipe-able such that it can be closed by swiping
- [ ] Make the `course info drawer` swipe-able such that it can be closed by swiping
- [ ] Add close (x) button to `course menu drawer` which closes the drawer when clicked.
- [ ] Add close (x) button to `course info drawer` which closes the drawer when clicked.

Note: Both drawers will be opened by their buttons only.

PR #926, opened 2022-10-21
Start linting typescript code

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
https://github.com/mitodl/eslint-config/issues/15, https://github.com/mitodl/eslint-config/issues/14

#### What's this PR do?
Updates the linting configuration to support typescript. 

#### How should this be manually tested?
There should be no functional changes, although I did change a few small pieces of application code (e.g., adding missing dependencies to a useEffect hook). So, run any checks you like on OCW.

PR #928, opened 2022-10-25
feat: close mobile course drawers via swiping

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Testing
  - [x] Code is tested

closes https://github.com/mitodl/ocw-hugo-themes/issues/895

#### What's this PR do?
- Makes the course menu drawer swipe-able such that it can be closed by swiping
- Makes the course info drawer swipe-able such that it can be closed by swiping
- Does it by adding `touchstart` and `touchend` event listeners

- Checkout this branch
- Build any course on v2 theme or view [netlify deployed version](https://ocw-hugo-themes-course-v2-pr-928--ocw-next.netlify.app/)
- Go to mobile screen on any course page.
- open course menu drawer, swipe it to left, and verify that it is closed smoothly by swiping
- open course info drawer, swipe it to right, and verify that it is closed smoothly by swiping
- Ideally: Open the netlify deployed version on your mobile, and test it there.

https://user-images.githubusercontent.com/93309234/197763281-65f261d4-349e-4bdb-be10-655bad4007ac.mov

### DIFF
--- a/course-v2/assets/js/mobile_course_drawers.ts
+++ b/course-v2/assets/js/mobile_course_drawers.ts
@@ -1,6 +1,19 @@
+enum SWIPE_DIRECTION {
+  Left = "L",
+  Right = "R"
+}
+
 export const initCourseDrawersClosingViaSwiping = () => {
-  elementSwiping("mobile-course-nav", "mobile-course-nav-toggle", "L")
-  elementSwiping("course-info-drawer", "mobile-course-info-toggle", "R")
+  elementSwiping(
+    "mobile-course-nav",
+    "mobile-course-nav-toggle",
+    SWIPE_DIRECTION.Left
+  )
+  elementSwiping(
+    "course-info-drawer",
+    "mobile-course-info-toggle",
+    SWIPE_DIRECTION.Right
+  )
 }
 
 /**
@@ -13,7 +26,7 @@ export const initCourseDrawersClosingViaSwiping = () => {
 const elementSwiping = (
   elementId: string,
   buttonId: string,
-  swipeDirection: string
+  swipeDirection: SWIPE_DIRECTION
 ) => {
   const element = document.getElementById(elementId)
   let touchstartX = 0
@@ -25,11 +38,11 @@ const elementSwiping = (
 
   element?.addEventListener("touchend", e => {
     touchendX = e.changedTouches[0].screenX
-    if (swipeDirection === "R") {
-      // checking if its a right swipe
-      if (touchendX > touchstartX) document.getElementById(buttonId)?.click()
-    } else {
-      if (touchendX < touchstartX) document.getElementById(buttonId)?.click()
+    if (
+      (swipeDirection === SWIPE_DIRECTION.Right && touchendX > touchstartX) ||
+      (swipeDirection === SWIPE_DIRECTION.Left && touchendX < touchstartX)
+    ) {
+      document.getElementById(buttonId)?.click()
     }
   })
 }
--- a/course-v2/assets/js/mobile_course_drawers.ts
+++ b/course-v2/assets/js/mobile_course_drawers.ts
@@ -1,0 +1,36 @@
+export const initCourseDrawersClosingViaSwiping = () => {
+  elementSwiping("mobile-course-nav", "mobile-course-nav-toggle", "L")
+  elementSwiping("course-info-drawer", "mobile-course-info-toggle", "R")
+}
+
+/**
+ * It adds touch event listener to the element and clicks the button when swiped in the mentioned direction.
+ *
+ * @param {string} elementId element on which touch eventlistenter is added.
+ * @param {string} buttonId This button will be clicked on swiping
+ * @param {string} swipeDirection L= Swipe left, R= Swipe Right
+ * @return {any} data fetched from localstorage
+ */
+const elementSwiping = (
+  elementId: string,
+  buttonId: string,
+  swipeDirection: string
+) => {
+  const element = document.getElementById(elementId)
+  let touchstartX = 0
+  let touchendX = 0
+
+  element?.addEventListener("touchstart", e => {
+    touchstartX = e.changedTouches[0].screenX
+  })
+
+  element?.addEventListener("touchend", e => {
+    touchendX = e.changedTouches[0].screenX
+    if (swipeDirection === "R") {
+      // checking if its a right swipe
+      if (touchendX > touchstartX) document.getElementById(buttonId)?.click()
+    } else {
+      if (touchendX < touchstartX) document.getElementById(buttonId)?.click()
+    }
+  })
+}
--- a/course-v2/assets/js/mobile_course_drawers.ts
+++ b/course-v2/assets/js/mobile_course_drawers.ts
@@ -1,18 +1,18 @@
-enum SWIPE_DIRECTION {
+enum SwipeDirection {
   Left = "L",
   Right = "R"
 }
 
 export const initCourseDrawersClosingViaSwiping = () => {
-  elementSwiping(
+  enableSwiping(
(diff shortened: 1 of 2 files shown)


course-v2/assets/js/mobile_course_drawers.ts
```diff
@@ -0,0 +1,35 @@
...
+  })
+
+  element?.addEventListener("touchend", e => {
+    touchendX = e.changedTouches[0].screenX
+    if (swipeDirection === "R") {
+      // checking if its a right swipe
+      if (touchendX > touchstartX) document.getElementById(buttonId)?.click()
+    } else {
+      if (touchendX < touchstartX) document.getElementById(buttonId)?.click()
+    }
```
Wassaf-Shahzad, line 33: I might be missing something but why not use use "!==" rather than checking inequalities.

Wassaf-Shahzad (changes requested):

course-v2/assets/js/mobile_course_drawers.ts
MAbdurrehman12: inequalities are used to identify the swipe direction. 
if `touchendX > touchstartX` then its a right swipe.
if `touchendX < touchstartX` then its a left swipe.

We can't use  `!==`, as it will enable the drawer to be closed in both, left and right directions but actually, it should be closed only in one direction. (course menu drawer should only be closed in left direction whereas course info drawer should only be closed in right direction).

Try replacing the inequalities with "!==" and you will notice what I'm talking about

course-v2/assets/js/mobile_course_drawers.ts
```diff
@@ -0,0 +1,35 @@
...
+  let touchstartX = 0
+  let touchendX = 0
+
+  element?.addEventListener("touchstart", e => {
+    touchstartX = e.changedTouches[0].screenX
+  })
+
+  element?.addEventListener("touchend", e => {
+    touchendX = e.changedTouches[0].screenX
+    if (swipeDirection === "R") {
```
Wassaf-Shahzad, line 28: Overall looks good, this is just a nitpick and you can choose to implement this. Since you are already passing a swipeDirection parameter into the function, we can just pass the "<" and ">" as parameters into the function and evaluate the valid direction.
Generally eval function is considered `evil` but since we are generating strings ourselves, we would not face any issues.
[Reference for eval function](https://stackoverflow.com/questions/197769/when-is-javascripts-eval-not-evil) 
```suggestion
   if (eval(`${touchendX} ${swipeDirection} ${touchstartX}`)) document.getElementById(buttonId)?.click()
    
```

MAbdurrehman12: Yeah, I had this in mind but did not use it deliberately:
1. since it's considered quite bad ("evil") due to potential security concerns (though our string is hardcoded)  
2. to have better readability of the code. 

But  I agree with you that it will make the code simpler. 
@ChristopherChudzicki Do you have any thoughts on this?

course-v2/assets/js/mobile_course_drawers.ts
```diff
@@ -0,0 +1,35 @@
...
+ * It adds touch event listener to the element and clicks the button when swiped in the mentioned direction.
+ *
+ * @param {string} elementId element on which touch eventlistenter is added.
+ * @param {string} buttonId This button will be clicked on swiping
+ * @param {string} swipeDirection L= Swipe left, R= Swipe Right
+ */
+const elementSwiping = (
+  elementId: string,
+  buttonId: string,
+  swipeDirection: string
```
Wassaf-Shahzad, line 16: You can also make swipeDirection an Enum or a Union Type with values of left and right. which would make it more readable and a bit more error safe.

Wassaf-Shahzad (changes requested):

course-v2/assets/js/mobile_course_drawers.ts
Wassaf-Shahzad: @MAbdurrehman12, You also dont need to nest if blocks since you can just use the `&&` operator with the child if block.

Wassaf-Shahzad (changes requested):

course-v2/assets/js/mobile_course_drawers.ts
Wassaf-Shahzad: @MAbdurrehman12 We can also dont need to use a nested if since we can just use the "&&" with the child if and make it single level block. This would make the code a bit for readable.

Wassaf-Shahzad (changes requested):

course-v2/assets/js/mobile_course_drawers.ts
```diff
@@ -0,0 +1,35 @@
...
+  let touchendX = 0
+
+  element?.addEventListener("touchstart", e => {
+    touchstartX = e.changedTouches[0].screenX
+  })
+
+  element?.addEventListener("touchend", e => {
+    touchendX = e.changedTouches[0].screenX
+    if (swipeDirection === "R") {
+      // checking if its a right swipe
```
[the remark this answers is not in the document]
MAbdurrehman12: > Using an enum (or at union, 'left' | 'right' for the type as Wassaf suggests below would also be good.

yeah, I have now added an enum

> Question: Does bootstrap support swiping? Or is there some minimal library we could use? I don't think this approach would animate the closing / opening, which is a little strange for swipe behavior.

I couldn't find any default/builtin swiping for the navbar which we are using, however, I think I did find a 3rd party for this, but I was also reluctant to add it for just 'swiping'.

MAbdurrehman12:
@Wassaf-Shahzad @ChristopherChudzicki Thanks for the review and feedback guys, can you please have another look?

Wassaf-Shahzad (approved):

course-v2/assets/js/mobile_course_drawers.ts
```diff
@@ -0,0 +1,48 @@
+enum SWIPE_DIRECTION {
```
ChristopherChudzicki, line 1: **Suggestion:** enums and their members are usually PascaleCase, so `SwipeDirection`.

But apparently #926 doesn't enforce this, hence just a suggestion.

course-v2/assets/js/mobile_course_drawers.ts
```diff
@@ -0,0 +1,48 @@
...
+const elementSwiping = (
+  elementId: string,
+  buttonId: string,
+  swipeDirection: SWIPE_DIRECTION
+) => {
+  const element = document.getElementById(elementId)
+  let touchstartX = 0
+  let touchendX = 0
+
+  element?.addEventListener("touchstart", e => {
```
ChristopherChudzicki, line 35: Rather than using optional chaining, I usually throw an error in situations like this if `element` is null.

Rationale: `element` should not be null. If it is, that's an error. We can either throw the error and get a better message and typing below, or we exit the function / use optional chaining. Then there's no visible error, but the site still behaves wrong.

course-v2/assets/js/mobile_course_drawers.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/**
+ * It adds touch event listener to the element and clicks the button when swiped in the mentioned direction.
+ *
+ * @param {string} elementId element on which touch eventlistenter is added.
+ * @param {string} buttonId This button will be clicked on swiping
+ * @param {string} swipeDirection L= Swipe left, R= Swipe Right
+ */
+const elementSwiping = (
```
ChristopherChudzicki, line 26: **Suggestion:** "enableSwiping" might be a better name.

ChristopherChudzicki (approved):
👍 A few suggestions, up to you.

Issue #932, opened 2022-10-27
improper context in offline build for mathjax url


gumaerc:
### Steps to Reproduce

Build a site with MathJax using the `course-offline` theme

### Expected Behavior

The course builds successfully

### Actual Behavior

There is an error during the build

### Stacktrace

`Error: Error building site: failed to render pages: render of "page" failed: "/tmp/build/d1fb8792/ocw-hugo-themes/course-v2/layouts/_default/baseof.html:186:5": execute of template failed: template: pages/single.html:186:5: executing "pages/single.html" at <partial "mathjax_if_necessary.html" (dict "context" .)>: error calling partial: "/tmp/build/d1fb8792/ocw-hugo-themes/base-theme/layouts/partials/mathjax_if_necessary.html:21:16": execute of template failed: template: partials/mathjax_if_necessary.html:21:16: executing "partials/mathjax_if_necessary.html" at <partial "webpack_url.html" (dict "context" . "url" "/static/mathjax/tex-svg.js")>: error calling partial: "/tmp/build/d1fb8792/ocw-hugo-themes/course-offline/layouts/partials/webpack_url.html:1:18": execute of template failed: template: partials/webpack_url.html:1:18: executing "partials/webpack_url.html" at <partial "path_to_root.html" .context>: error calling partial: "/tmp/build/d1fb8792/ocw-hugo-themes/course-offline/layouts/partials/path_to_root.html:1:29": execute of template failed: template: partials/path_to_root.html:1:29: executing "partials/path_to_root.html" at <.RelPermalink>: can't evaluate field RelPermalink in type []string`

PR #906, opened 2022-10-11
only load mathjax if necessary

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Changes have been manually tested

#### What are the relevant tickets?
https://github.com/mitodl/ocw-hugo-themes/issues/905

#### What's this PR do?
This PR switches us to including Mathjax only on pages that use it.

#### How should this be manually tested?
First, load https://ocw.mit.edu/courses/11-382-water-diplomacy-spring-2021/ and view network tab. Observe that a file called `tex-svg.js` is downloaded.

Now on this branch:
1. Load any course page without math. No `tex-svg.js` should be downloaded.
2. Edit the markdown to include `\\(E=mc^2\\)` OR `\\[E=mc^2\\]`. `tex-svg.js` should now be downloaded.
3. Repeat (2) for site pages, site homepage, on v1, v2, and www theme.

PR #933, opened 2022-10-27
pass .context to webpack_url in mathjax_if_necessary

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/932

#### What's this PR do?
In https://github.com/mitodl/ocw-hugo-themes/pull/906, we introduced a change that conditionally loads MathJax. This new partial, `mathjax_if_necessary.html`, calls `webpack_url.html`. The Hugo `.` object is passed as the `context` parameter, when the context being passed to `mathjax_if_necessary.html` itself is stored in `.context`. This same object should be passed on as the context object, not the parameter dict itself.

 - Clone https://github.mit.edu/mitocwcontent/res.8-004-january-iap-2015
 - Clone [`ocw-hugo-projects`](https://github.com/mitodl/ocw-hugo-projects)
 - Run `npm run build /path/to/mitocwcontent/res.8-004-january-iap-2015 /path/to/ocw-hugo-projects/ocw-course-v2/config-offline.yaml
 - Verify the build completes successfully
 - Browse to the course content folder and look in the `dist` folder for `index.html`. Double click it and verify that the page loads and you can click around the site. Some resources like the course image will not load unless you pull down the static assets prior to building, but that's not necessary for this test.

### DIFF
--- a/base-theme/layouts/partials/mathjax_if_necessary.html
+++ b/base-theme/layouts/partials/mathjax_if_necessary.html
@@ -18,5 +18,5 @@
       }
     }
 </script>
-<script src="{{ partial "webpack_url.html" (dict "context" . "url" "/static/mathjax/tex-svg.js") }}" defer></script>
+<script src="{{ partial "webpack_url.html" (dict "context" .context "url" "/static/mathjax/tex-svg.js") }}" defer></script>
 {{ end }}
--- a/base-theme/layouts/partials/mathjax_if_necessary.html
+++ b/base-theme/layouts/partials/mathjax_if_necessary.html
@@ -1,3 +1,4 @@
+{{ $context := .context }}
 {{ with findRE `\\\(.*?\\\)|\\[.*?\\]` .context.Page.Content 1 }}
 <script>
   window.MathJax = {
@@ -18,5 +19,5 @@
       }
     }
 </script>
-<script src="{{ partial "webpack_url.html" (dict "context" .context "url" "/static/mathjax/tex-svg.js") }}" defer></script>
+<script src="{{ partial "webpack_url.html" (dict "context" $context "url" "/static/mathjax/tex-svg.js") }}" defer></script>
 {{ end }}


ChristopherChudzicki (changes requested):
I have not tested this with the offline build yet, but the regular v2 course build seems to fail...

Using `COURSE_HUGO_CONFIG_PATH=/Users/cchudzicki/dev/ocw-hugo-projects/ocw-course-v2/config.yaml
`, then `yarn start:course` gives
```
 failed to render pages: render of "page" failed: "/Users/cchudzicki/dev/ocw-hugo-themes/course-v2/layouts/_default/baseof.html:186:5": execute of template failed at <partial "mathjax_if_necessary.html" (dict "context" .)>: error calling partial: "/Users/cchudzicki/dev/ocw-hugo-themes/base-theme/layouts/partials/mathjax_if_necessary.html:21:59": execute of template failed at <.context>: can’t evaluate field context in type []string 
```

gumaerc:
@ChristopherChudzicki Right, so the issue here was that the `{{ with findRE ...` statement remaps the `.` context object to be a string instead of a page object. I modified the PR to store the context object and then pass that along.

ChristopherChudzicki (approved):
👍 Looks good to me!

For (internal) posterity, Carey answered a lot of questions about this: https://mitodl.slack.com/archives/GQ89D0Z5M/p1666885665553749?thread_ts=1666884310.310079&cid=GQ89D0Z5M

gumaerc:
Summarizing our conversation linked above that will not be available publicly, we discussed how this issue come to be and why it only affects the offline build. When you view an HTML page offline by double clicking it in any given operating system, a `file://` directive is sent to your browser with an absolute path to the page based on the folder you've stored the content in locally. Since our webpack modules (like MathJax) are not content that Hugo is aware of like the way it is with `Page` objects generated from the `content` folder, Hugo cannot automatically generate relative URLs to these assets as far as I know. Therefore, the `path_to_root.html` partial in the `course-offline` theme takes a Page context passed in (the page currently being rendered) and a root-relative URL that you wish to link to. It then generates a series of `../` tracing back to the site root and appends the URL to the end. It's relatively simple, but I'd like to come up with a better way of handling this in the future.

PR #941, opened 2022-10-31
set up e2e tests with playwright

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#919 

#### What's this PR do?
This PR adds e2e testing via Playwright to OCW Hugo Themes.

There are a variety of other significant changes that I will try to call out in individual comments.

**First:** Run `yarn install` and read the readme:  https://github.com/mitodl/ocw-hugo-themes/tree/cc/ci2/tests-e2e  If you use VS Code, I highly recommend installing the Playwright extension. (See the Readme.).

1. Run `yarn test:e2e`. It should fail. Add the required environment variable. (Use value 3001; see https://github.com/mitodl/ocw-hugo-themes/pull/941#discussion_r1010893053 ).
2. Run `yarn test:e2e` again. It should suceed but it will also be kinda slow.
3. Run `yarn start:webpack`. In a separate shell, run `yarn test:e2e`. It should be pretty fast. (Now Playwright is reusing the webpack server.)
4. Stick a `.only` on one of the tests, and run `yarn test:e2e --debug`. You can step through the test. 
5. Run `yarn start:course` and `yarn start:www`, too. I changed some stuff with environment variables and scripts. Those should both still work. You could also try `OCW_TEST_COURSE=different-from-value-in-env-file yarn start:course`. That should work now, whereas previously it would not. 

I suggest starting with [the readme](https://github.com/mitodl/ocw-hugo-themes/tree/cc/ci2/tests-e2e), **then** reading some of the test files. The test files should more or less make sense without context.

Then look at all the other junk.


course-v2/layouts/partials/course_detail.html
```diff
@@ -1,15 +1,15 @@
 <div class="card">
   <div class="card-body">
-    <div class="course-detail-section">
+    <section class="course-detail-section">
       {{ partial "course_description.html" (dict "context" .) }}
-    </div>
+    </section>
```
ChristopherChudzicki, line 5: ⚠️⚠️ Site change! *very minor*

Changing div -> section here made seemed reasonable and made selecting "Course Info" section in Playwright a little easier. See https://github.com/testing-library/dom-testing-library/issues/403#issuecomment-554536248 for a related issue. (I believe that issue is where I picked up the pattern of using `.closest()` for stuff like this. 🤷 )

package.json
```diff
@@ -5,29 +5,31 @@
...
   "main": "base-theme/assets/index.js",
   "scripts": {
-    "lint": "eslint {base-theme,course,course-v2,fields,util,www}",
+    "lint": "eslint {base-theme,course,course-v2,fields,util,www,tests-e2e}",
     "start": "echo \"Run 'npm run start:www' or 'npm run start:course'\"",
-    "start:webpack": "webpack serve --config base-theme/assets/webpack/webpack.dev.ts --hot --host 0.0.0.0 --stats-children",
-    "start:www": "concurrently \"npm:start:webpack\" \"npm:start:www:hugo\" --kill-others",
-    "start:course": "concurrently \"npm:start:webpack\" \"npm:start:course:hugo\" --kill-others",
-    "start:fields": "concurrently \"npm:start:webpack\" \"npm:start:fields:hugo\" --kill-others",
+    "start:webpack": "dotenv -- webpack serve --config ./base-theme/assets/webpack/webpack.dev.ts --hot --host 0.0.0.0 --stats-children",
```
ChristopherChudzicki, line 10: This is [dotenv-cli](https://www.npmjs.com/package/dotenv-cli). It loads env vars on the cli.

base-theme/assets/webpack/webpack.dev.ts
```diff
@@ -5,6 +5,11 @@ import { CleanWebpackPlugin } from "clean-webpack-plugin"
 import MiniCssExtractPlugin from "mini-css-extract-plugin"
 import common from "./webpack.common"
 import "webpack-dev-server" // this import tells webpack's typings about the devServer type
+import * as envalid from "envalid"
+
+const env = envalid.cleanEnv(process.env, {
+  WEBPACK_PORT: envalid.port()
+})
```
ChristopherChudzicki, line 12: This is https://www.npmjs.com/package/envalid. It's nice for (a) validating that env variables are set and providing good error messages if they are not, and (b) typescript. Since `process.env.WHATEVER` is `string | undefined` by default.

I've scattered a few `envalid` calls into the repo. I *think* one common usage of envalid is to have a single file that validates the environment variables and sets defaults and stuff. (Like `settings.py` in Django, I guess). I think that would be really nice, but we'd have to have some centralized js-based cli through which we call hugo (Rather than our current shell scripts). I briefly explored doing something like that with [oclif](https://oclif.io/), but it seemed like way too big a change for the moment.

*Aside:  [oclif](https://oclif.io/) is cool. And it's built by Salesforce / it's what the heroku CLI is built upon. And while I like the idea of a centralized CLI for this repo, like django provides us elsewhere, I would not recommend `oclif` because https://npmtrends.com/commander-vs-oclif-vs-yargs*

package_scripts/build_all_courses.sh
```diff
@@ -3,7 +3,6 @@
 set -euo pipefail
```
ChristopherChudzicki, line 3: @gumaerc Do you think we need this script anymore?

test-sites/ocw-ci-test-www/README.md
```diff
@@ -0,0 +1 @@
+# ocw-ci-test-www
```
ChristopherChudzicki, line 1: I have not written very many tests for `www` yet, and this is the only file in the test site.

Really I included `www` mostly as a proof-of-concept for building / running tests against multiple sites using multiple hugo configs.

tests-e2e/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+To run the tests:
+1. `yarn start:webpack` in one shell, and
+2. `yarn test:e2e` in another. *Or `yarn test:e2e --debug` to step through tests.
+
+Running `yarn test:e2e` by itself will also work. However, if you start a webpack server in a separate process first, Playwright will use that server instead rather than starting a new one, resulting in much less overhead if you need to re-run tests.
+
+### Is there a `--watch` mode for re-running tests?
+
+No. For VS Code users, a similar experience can be achieved with [Playwright for VS Code](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright). Individual tests can be run from your editor with the click of a button.
```
ChristopherChudzicki, line 19: > For VS Code users, ... Individual tests can be run from your editor with the click of a button.

I swear I'm not _just_ a shill for Microsoft.

base-theme/layouts/partials/webpack_url.html
```diff
@@ -1,9 +1,10 @@
+{{- $port := 3001 -}}
```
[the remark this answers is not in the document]
ChristopherChudzicki: This is the hugo projects PR: https://github.com/mitodl/ocw-hugo-projects/pull/222, and https://github.com/mitodl/ocw-hugo-themes/tree/cc/ci3 branch of themes includes the required change.

package_scripts/build_all_courses.sh
gumaerc: A while back I had been using this script to build all courses locally, and even before that it was actually being used in an EC2 box to deploy the whole site before we had `ocw-studio` and Concourse. Now that we have a full dev setup for `ocw-studio` that can build the whole site locally, I think we can safely remove this without consequence.

pt2302:
During testing, found out we first need to run `npx playwright install`.

pt2302:
Runs in my tests with https://github.com/mitodl/ocw-hugo-themes/blob/f5ba169b213236cb0e7d3aec60a75f32cc0ec8ed/playwright.config.ts#L41 changed to `http://0.0.0.0:${env.WEBPACK_PORT}/static/css/main.css`.

pt2302 (approved):
Looks good to me!

playwright.config.ts
```diff
@@ -0,0 +1,47 @@
...
+  },
+  fullyParallel: true,
+  forbidOnly:    !!process.env.CI,
+  retries:       process.env.CI ? 2 : 0,
+  workers:       process.env.CI ? 1 : undefined,
+  reporter:      'html',
+  use:           {
+    actionTimeout: 0,
+    trace:         'on-first-retry',
+  },
```
ChristopherChudzicki, line 30: BTW: Most of these settings are what Playwright recommends. I created this config file via (https://playwright.dev/docs/intro#installing-playwright)
```
yarn create playwright
```
and these were the settings it gave me. My customizations were:
- removing some comments because it seemed cluttery and TS docstrings / playwright docs have the same information.
- removing the extra browsers (firefox, webkit)
- customizing the `webServer` portion. 

We could restore firefox + webkit. I'm indifferent. I'm not sure how useful the extra browsers would be at the moment; maybe if we were doing visual or very interactive tests.

package.json
```diff
@@ -155,5 +162,8 @@
     "webpack-dev-server": "^4.11.1",
     "webpack-merge": "^5.0.9"
   },
-  "packageManager": "yarn@3.2.2"
+  "packageManager": "yarn@3.2.2",
+  "devDependencies": {
+    "@playwright/test": "^1.27.1"
```
ChristopherChudzicki, line 167: BTW: I mentioned on a zoom call once that there is [playwright-testing-library](https://github.com/testing-library/playwright-testing-library) akin to other libs we use like [react-testing-library](https://testing-library.com/docs/react-testing-library/intro/) (all part of the [Testing Library](https://github.com/testing-library) family).

I was initially using it, but decided to drop it. Playwright itself introduced some features inspired by testing-library, and the creator of playwright-testing-library says [^1]:

> Start out without playwright-testing-library and see how the new Testing Library API in 1.27 serves y

[^1]: https://github.com/testing-library/playwright-testing-library/issues/558#issuecomment-1274195339w

package_scripts/build_all_courses.sh
ChristopherChudzicki: Unless someone wants it here, I think I'll drop that script in a separate PR.

base-theme/assets/webpack/webpack.dev.ts
```diff
@@ -18,7 +23,7 @@ const devOverrides: Configuration = {
   devtool: "eval-source-map",
 
   devServer: {
-    port:    process.env.PORT || 3001,
+    port:    env.WEBPACK_PORT,
```
gumaerc, line 26: Why not `env.WEBPACK_PORT || 3001`? I get an error if I try to run `yarn test:e2e` without `WEBPACK_PORT` explicitly in my env. We default it to 3001 when running a site for dev, so why wouldn't we do that for tests?

package.json
gumaerc: I feel like an idiot now because I didn't realize this package was required for NPM to pass env variables to package scripts. I thought all you needed install was the `dotenv` package and that would happen, but apparently I was mistaken.

package_scripts/start_course.sh
```diff
@@ -3,31 +3,21 @@
 set -euo pipefail
 
 source package_scripts/common.sh
-load_env --require-dot-env
```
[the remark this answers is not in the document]
gumaerc: Yea, I think we added this a while back because we were having trouble with `dotenv`. Like I mentioned above, I didn't realize that the `dotenv-cli` package was required and that you had to prefix any commands that needed the `.env` file with it. They don't mention it anywhere on the `dotenv` Github and it's kind of puzzling why the "CLI" component of that is a separate package. Either way, it works much better  so yea it makes sense to get rid of this.

gumaerc (changes requested):
All of this looks great and is a good start to getting some actual testing done here. Everything works, I just have one request regarding the `WEBPACK_PORT` environment variable:

package.json
ChristopherChudzicki: I actually had never used this package before; I have used
```
 node -r dotenv/config your_script.js
```
(See https://www.npmjs.com/package/dotenv#Preload) but that only works if you're calling node.

base-theme/assets/webpack/webpack.dev.ts
ChristopherChudzicki: > We default it to 3001 when running a site for dev

If we merge this PR as-is, that is not longer true.

But, to be sure, having a required WEBPACK_PORT env variable also annoys me. But I do not think we currently have a good way to set the default, and the current situation on `main` illustrates this.

Currently on main:
- [webpack.dev.ts](https://github.com/mitodl/ocw-hugo-themes/blob/main/base-theme/assets/webpack/webpack.dev.ts#L21) uses `process.env.PORT || 3001` 
- but [base-theme/layout/partials/webpack_url.html](https://github.com/mitodl/ocw-hugo-themes/blob/main/base-theme/layouts/partials/webpack_url.html#L4) always uses `3001`.

So you can see the inconsistency: if you have `PORT` env var set, webpack will respect it but hugo will not. 💥.

Here are some options that occur to me:
1. Hardcode 3001 everywhere. *I don't really want to do this, but it doesn't actually seem terrible*
2. Use an environment variable with no default. If it's not set, you get an error message. *A little annoying.*
3. Use an environment variable with inlined-defaults. *This is the current situation, and IMO it is worse than (1) or (2), because of the possibility for inconsistency, as demonstrated by current `maint` branch.*
4. Somehow have a "single source of truth" for environment variables that also allows setting defaults... I can think of two ways:
    - **4A**. Build a CLI... call it `ocw`. It would be like `./manage.py` in django. We'd use it like:

        ```
        yarn ocw --help
        yarn ocw test
        yarn ocw test-e2e
        yarn ocw build course        <-- build a course; we'd use this in pipelines
        yarn ocw build course --help <-- get help info for all the flags/args to build
        yarn ocw serve course        <-- serve a course in dev mode
        ``` 
    
    where the `ocw` CLI would do something like load environment variables, set default values for some, provide useful error messages, have docstrings...like `manage.py` does.

    **This is appealing to me for a variety of reasons. Setting env variables is one. Another is it's a way to provide documentation without cluttering the README.**
    
    - **4B**. Another option is to have `.env` and `.env.defaults` and

        ```sh
        # replace this
        yarn dotenv <cmd>
        # with this:
        yarn dotenv -e .env -e .env.defaults <cmd>
        ```

        I don't dislike this idea. But then do we have three files? (env, env.defaults, and env.sample)? Or do we just use env.sample in place of env.defaults? I'm not sure all the values in .env.sample are reasonable as defaults.

ChristopherChudzicki: oops, submitted that too earlier...

In summary: 4A is my longer-term preference, but the approach of this PR is (2) which seems ok. Maybe 4B is good.

What do you think? or any other ideas?

gumaerc: I actually think 3 is the best option specifically referring to the port. There aren't that many places where it's used to generate inconsistency, and I have never heard of software that spins up a service that binds to a port requiring you to specify said port without a default.

ChristopherChudzicki: > There aren't that many places where it's used to generate inconsistency,

But even with not many, we already have the inconsistency on `main`. Run `PORT=3456 yarn start:course` and observe that OCW does not load webpack assets.

ChristopherChudzicki: Actually, I'm going to try (3)... if I can set all the js defaults in one place, that's good enough. If the env var is needed in Hugo, too, we'll have to set the default in two places. If we can keep it to two places, I think that's ok.

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "lint": "eslint {base-theme,course,course-v2,fields,util,www,tests-e2e}",
     "start": "echo \"Run 'npm run start:www' or 'npm run start:course'\"",
-    "start:webpack": "dotenv -- webpack serve --config ./base-theme/assets/webpack/webpack.dev.ts --hot --host 0.0.0.0 --stats-children",
+    "start:webpack": "dotenv -- webpack serve --config ./base-theme/assets/webpack/webpack.dev.ts --hot --stats-children",
```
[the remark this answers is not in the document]
gumaerc: @ChristopherChudzicki Hitting `http://localhost:3001` on your phone is going to try and access port 3001 on your phone, but your phone isn't running a web server. Binding to `0.0.0.0` means that you can hit the *local IP address of your dev machine* from any other device on your network. Say your dev machine has a local IP of `192.168.1.100`. That means that if you run this with `--host 0.0.0.0` then on your phone you can go to `http://192.168.1.100:3001` and you will get your webpack instance.

base-theme/assets/webpack/webpack.dev.ts
gumaerc: However it's accomplished, Node and Hugo don't require you to set a port when you run them and will run on a default port. We shouldn't require that it be set either, but allow it to be overridden. I'd rather hard code 3000 for Hugo and 3001 for Webpack than require the env variable to be set to spin up the server.

package.json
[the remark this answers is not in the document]
ChristopherChudzicki: @gumaerc Awesome, thank you for that explanation!  Sounds like we do want 0.0.0.0. I've reverted the "replace 0.0.0.0 with localhost" for now.

ChristopherChudzicki (commented):
@gumaerc I believe that the most recent commit (https://github.com/mitodl/ocw-hugo-themes/pull/941/commits/c9cc0eeb164b263d2e8b644bf9eaeccd4a2e1dcb) has all changes since your last review. _Aside from a rebase with no conflicts._

I went ahead and set a default port in a 4 places. I'm not super happy with this and I plan to make a small folllowup PR that reorganizes a few env variable things, but I'd rather not add even more env-reorganization to this PR.

Thanks again for your explanations about 0.0.0.0 (I reverted the 0.0.0.0 --> localhost change for now; will also address that in my incoming PR).

gumaerc (approved):
👍 LGTM

PR #953, opened 2022-11-02
Ab/set compact UI live

abeglova:
#### Pre-Flight checklist

Screen Shot 2022-11-02 at 4 23 46 PM
Screen Shot 2022-11-02 at 4 24 11 PM
Screen Shot 2022-11-02 at 4 31 06 PM
Screen Shot 2022-11-02 at 4 31 14 PM

- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/hq/issues/79

#### What's this PR do?
This pr unhides the toggle for the compact search view to enable the feature. This pr is currently up to test https://github.com/mitodl/course-search-utils/pull/45 which needs to be merged first. The test failure should be fixed by updating course-search-utils.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "browserslist": "> 0.25%, not dead",
   "dependencies": {
-    "@mitodl/course-search-utils": "^2.0.1",
+    "@mitodl/course-search-utils": "^2.0.4",
     "@mitodl/ocw-to-hugo": "1.36.0",
     "@sentry/browser": "^5.19.0",
     "@testing-library/react-hooks": "^7.0.2",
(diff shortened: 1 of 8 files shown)


ChristopherChudzicki (commented):
Working well! Two things I noticed:

1. Clicking the list view sets `u=list` in the URL params. I think it would be better to just remove the URL param. Should be able to do this by replacing `onClick={() => updateUI(SEARCH_LIST_UI)}` with `onClick={() => updateUI(null)}`.
2. IMO, we should change the terminology a little... Currently the two values of `ui` are `list | compact`. But both views are lists.

    I would suggest "compact" and "relaxed"; or "compact" and "default"; or "compact" and "comfortable". _gmail has a similar option and uses comfortable for the spacier one.

ChristopherChudzicki (approved):

ChristopherChudzicki:
Looks good to me!

Issue #937, opened 2022-10-31
only show course download if ZIP file exists


gumaerc:
As a user of OCW, I'd like to only be presented with course download links that work

#### Designs and Mockups

https://projects.invisionapp.com/share/PA131XTPEK76#/screens/470225774_Page_-_Download_Options_-_V1-1
https://invis.io/UX132JLIWE43#/470179934_OCW_-_Course_Home_V2-1
https://invis.io/6N12T6YA43PW#/467901625_OCW_-_Course_Detail_-_Sidebar_Alt_-__v-1-4

#### Acceptance Criteria:

- [ ] If the ZIP file for a course has not been generated yet or otherwise doesn't exist, don't show the download link

gumaerc:
I've tried a couple different methods of doing this today and haven't been happy with any of them:

 - Firstly, I tried using Hugo's `resources.GetRemote` with `"method" "head"` to make a `HEAD` request during the build at the URL to check for its existence. This method will make the HEAD request but the object it returns is a Hugo `Resource` object that only contains the response body, not the response headers. Logging the response, I actually strangely enough get a blank string when requesting an invalid URL testing against QA. When requesting a valid ZIP file, I get an error like `error calling resources.GetRemote: failed to read remote resource "https://live-qa.ocw.mit.edu/courses/2-57-nano-to-macro-transport-processes-spring-2012/2.57-spring-2012.zip": unexpected EOF`. All in all, this method seems unusable. As with most things in Hugo, I found a thread where Bjorn shines some light on this and says "we're not adding some low-level HTTP API to Hugo" here: https://github.com/gohugoio/hugo/issues/7699#issuecomment-698753361
 - Secondly, I tried making the `HEAD` request client side and checking for a 404. This method is pretty simple and does work, but presents two issues:
   - Since this is on the client side, even using raw native JS and making sure it loads quickly as possible in-line like we have done in the past for the mobile table styling, there is a quick "flicker" while the HTTP request is being made to determine if the file exists. I'm not sure that there's a way around this as the HTTP request is always going to take a bit of time.
   - CORS is not a problem right now since the ZIP files will be hosted in the same bucket alongside the rest of the content, however if we decide to move them in the future it could become a problem.

So yea, neither of these options are really that great. With the Hugo method we could maybe tweak the Fastly response so that something shows up in response to a HEAD request that can be usable there but I'm not certain. At this point, I'm thinking a better strategy would be something along the lines of this:

[...]

PR #931, opened 2022-10-26
course download button desktop

gumaerc:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-hugo-themes/issues/929

#### What's this PR do?
This PR does the following:

 - Adds a "download course" button to the `/resources` page on every course with the URL being the base URL plus the course short id as the ZIP filename
 - Turns the resource lists into collapsible lists
 - Makes sure that the Lecture Videos are always on the top and expanded by default
 - Make sure other resource lists are collapsed by default

#### How should this be manually tested?
This PR is best tested by running `ocw-studio` locally and mocking out the publishing environment, so you can generate the ZIP files and the live course sites the same way they would be in RC and production:

 - Clone `ocw-studio` on the `cg/offline-build-short-id-zip-filename` branch
 - Make sure you've taken care of the following prerequisites:
   - Minio support configured
   - Concourse support configured
   - Custom Github organization configured
   - Some test courses in your database using the `ocw-course` starter, I'm using 9-40-introduction-to-neural-computation-spring-2018 in my example below but you should test any courses you want that have resources
   - In addition to configuring your `.env` for the above, also set:
```
RESOURCE_BASE_URL_DRAFT=https://draft.ocw.mit.edu
RESOURCE_BASE_URL_LIVE=https://live.ocw.mit.edu
STATIC_API_BASE_URL=https://live.ocw.mit.edu
OCW_HUGO_PROJECTS_BRANCH=main
OCW_HUGO_THEMES_BRANCH=cg/course-download
```
 - Make a temporary edit to `websites/views.py` to limit the sites seen by `mass-build-sites`:
```
        ...
        sites = Website.objects.filter(name="9-40-introduction-to-neural-computation-spring-2018")
        serializer = WebsiteMassBuildSerializer(instance=sites, many=True)
        return Response({"sites": serializer.data})
```
 - Bring up `ocw-studio` with `docker-compose up`
 - Run the following management commands one by one:
   - `./manage.py upsert_theme_assets_pipeline`
   - `./manage.py backpopulate_pipelines --filter 9-40-introduction-to-neural-computation-spring-2018`
[...]

PR #960, opened 2022-11-03
only show download button if zip exists

gumaerc:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/937

#### What's this PR do?
In https://github.com/mitodl/ocw-hugo-themes/pull/931, we added a "course download" button to the resources download page with a URL based off the work done in `ocw-studio` to create Concourse pipelines that build offline versions of each course and ZIP them up. This PR makes the visibility of that download conditional on Hugo being able to make a `HEAD` request against the ZIP URL to confirm its existence.

 - Clone any course you want from https://github.mit.edu/mitocwcontent/ and point to it in your env using `COURSE_CONTENT_PATH` and `OCW_TEST_COURSE`
 - Clone `ocw-hugo-projects` on the `cg/allow-head-zip` branch
 - In `ocw-hugo-projects/course-v2/config.yaml`, we need to temporarily set the `baseUrl` setting to http://localhost:3000/. Hugo automatically sets it to `//localhost:3000` which ironically causes `resources.GetRemote` to throw an error about the protocol not being specified.
 - Download the ZIP file for the course you're testing using the formula https://ocw.mit.edu/courses/9-40-introduction-to-neural-computation-spring-2018/9.40-spring-2018.zip, replacing the `name` and `short_id` properties with whatever course you're using, or create a dummy ZIP file with the same name matching the `short_id`
 - Place the ZIP file in the root of the `public` folder inside your course content repo
 - Run `yarn start:course`
 - Visit http://localhost:3000/download and verify that you see a download button
 - In `course-v2/layouts/partials/resources_header.html`, make a temporary edit to spoof requesting a file that doesn't exist by adding a `-2` to the filename like this: `{{- $downloadCourseUrl := printf "%v/%v-2.zip" $trimmedBaseUrl $shortId -}}`
 - Save the file to refresh the page, and the download button should disappear

chrome_2022-11-03_19-47-48
chrome_2022-11-03_19-48-25

### DIFF
--- a/course-v2/layouts/partials/resources_header.html
+++ b/course-v2/layouts/partials/resources_header.html
@@ -1,5 +1,5 @@
-{{- $courseData := site.Data.course --}}
-{{- $zenDeskUrl := "https://mitocw.zendesk.com/hc/en-us/articles/4414681093659-I-have-downloaded-an-MIT-OpenCourseWare-course-but-I-can-t-access-the-materials-How-do-I-get-started-" --}}
+{{- $courseData := site.Data.course -}}
+{{- $zenDeskUrl := "https://mitocw.zendesk.com/hc/en-us/articles/4414681093659-I-have-downloaded-an-MIT-OpenCourseWare-course-but-I-can-t-access-the-materials-How-do-I-get-started-" -}}
 {{- $trimmedBaseUrl := strings.TrimSuffix "/" site.BaseURL -}}
 {{- $shortId := printf "%v-%v-%v" $courseData.primary_course_number (lower $courseData.term) $courseData.year -}}
 {{- $downloadCourseUrl := printf "%v/%v.zip" $trimmedBaseUrl $shortId -}}
--- a/course-v2/layouts/partials/resources_header.html
+++ b/course-v2/layouts/partials/resources_header.html
@@ -1,25 +1,37 @@
-{{ $courseData := site.Data.course }}
-{{ $zenDeskUrl := "https://mitocw.zendesk.com/hc/en-us/articles/4414681093659-I-have-downloaded-an-MIT-OpenCourseWare-course-but-I-can-t-access-the-materials-How-do-I-get-started-" }}
-{{ $trimmedBaseUrl := strings.TrimSuffix "/" site.BaseURL }}
-{{ $shortId := printf "%v-%v-%v" $courseData.primary_course_number (lower $courseData.term) $courseData.year }}
-{{ $downloadCourseUrl := printf "%v/%v.zip" $trimmedBaseUrl $shortId }}
-{{ $headResponse := resources.GetRemote $downloadCourseUrl }}
-{{ if $headResponse }}
+{{- $courseData := site.Data.course --}}
+{{- $zenDeskUrl := "https://mitocw.zendesk.com/hc/en-us/articles/4414681093659-I-have-downloaded-an-MIT-OpenCourseWare-course-but-I-can-t-access-the-materials-How-do-I-get-started-" --}}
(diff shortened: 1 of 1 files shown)


course-v2/layouts/partials/resources_header.html
```diff
@@ -1,22 +1,37 @@
...
-{{ $trimmedBaseUrl := strings.TrimSuffix "/" site.BaseURL }}
-{{ $shortId := printf "%v-%v-%v" $courseData.primary_course_number (lower $courseData.term) $courseData.year }}
-{{ $downloadCourseUrl := printf "%v/%v.zip" $trimmedBaseUrl $shortId }}
+{{- $courseData := site.Data.course -}}
+{{- $zenDeskUrl := "https://mitocw.zendesk.com/hc/en-us/articles/4414681093659-I-have-downloaded-an-MIT-OpenCourseWare-course-but-I-can-t-access-the-materials-How-do-I-get-started-" -}}
+{{- $trimmedBaseUrl := strings.TrimSuffix "/" site.BaseURL -}}
+{{- $shortId := printf "%v-%v-%v" $courseData.primary_course_number (lower $courseData.term) $courseData.year -}}
+{{- $downloadCourseUrl := printf "%v/%v.zip" $trimmedBaseUrl $shortId -}}
+{{- $headResponse := resources.GetRemote $downloadCourseUrl (dict "method" "head") -}}
+<!--
```
gumaerc, line 7: This is a lengthy explanation but I figured it was worth leaving. The problem here in a nutshell is that if we leave off the `(dict "method" "head")` and instead make a normal `GET` request on the ZIP file, the entire thing will be downloaded during the course of the build. This could have a bad effect when building all 2700+ OCW courses and greatly increase the time it takes to rebuild the whole site. This is why I'm making a `HEAD` request here, to just get the headers on the file. Ideally I would have just liked to inspect the headers, check the type and content length, but Hugo does not support reading the response headers. It only tries to fetch the body and present that as a resource. The potential downside to doing it this way is that it's possible other URL's could return the "unexpected EOF" error if the request returns a status code of 200 with an empty body. The alternative is to do a `GET` instead and see what the fallout is from that on build times in RC.

MAbdurrehman12 (approved):
LGTM 👍

MAbdurrehman12:
For any other reviewer testing this: 
> Clone ocw-hugo-projects on the cg/allow-head-zip branch

Hugo projects branch is `cg/allow-head-and-zip`

course-v2/layouts/partials/resources_header.html
```diff
@@ -1,22 +1,37 @@
...
+  the response headers. If we just perform a normal GET here the entire ZIP file will be 
+  downloaded during the build which could vastly balloon the build time or make the build fail.
+
+  A 404 on the OCW site returns an empty body with a 404 status code, which Hugo properly 
+  interprets as nil. In this case the "unexpected EOF" means "I found a file and got a 200 
+  in the response, but the file was empty."
+  
+  The file is empty because we only made a HEAD request.
+-->
+{{- if in $headResponse "unexpected EOF" -}}
```
MAbdurrehman12, line 19: Not sure if this is an optimal approach but I thought about this and couldn't come up with any other robust/better solution.

gumaerc: Yea, I'm not super happy with it either as you can see but I think this will work for now. I detailed another potential solution in the issue involving tracking the offline site publish date in `ocw-studio` which is probably more realistically the way to go here but this will prevent the link being shown when the ZIP doesn't exist yet for now.

Issue #955, opened 2022-11-02
more use of "Show more" 


pdpinch:
In https://github.com/mitodl/ocw-hugo-themes/pull/876 we replaced the line/arrow UI for truncating text on the CHP with a "show more"/"show less" UI. 

But we still have the old UI on some resource pages, like https://live-qa.ocw.mit.edu/courses/res-tll-008-social-and-ethical-responsibilities-of-computing-serc-fall-2021/resources/res-tll-008f21-sts047-1/

#### Screenshots

![image](https://user-images.githubusercontent.com/430126/199603581-730fd944-6876-40e1-b2de-aaa29ce6ad6f.png)

![image](https://user-images.githubusercontent.com/430126/199603688-636976a6-1cba-42c1-9090-aa66889f8991.png)

#### Acceptance Criteria:

- [ ] Use the same show more UI everywhere
- [ ] get rid of the lines and arrows

Ferdi:
I would simply show the full title.

MAbdurrehman12:
For reference/comparison: `show more/show less` looks something like this:

https://user-images.githubusercontent.com/93309234/199702631-547208c7-0eed-48eb-989a-2ca3735553cd.mov

pdpinch:
What does the longest resource title look like, if we go with showing the full title? 

The longest resource title is #3 in this search result: https://live-qa.ocw.mit.edu/search/?q=%22Theory%20of%20Discounted%20Problems%22&type=resourcefile&u=compact

MAbdurrehman12:
> What does the longest resource title look like, if we go with showing the full title?
> 
> The longest resource title is #3 in this search result: https://live-qa.ocw.mit.edu/search/?q=%22Theory%20of%20Discounted%20Problems%22&type=resourcefile&u=compact

<img width="1377" alt="image" src="https://user-images.githubusercontent.com/93309234/199920325-8d0682d0-4c05-4056-8903-33cb62067fc5.png">

not sure but I think such long titles are not very common, usually titles are of normal length. So as per @Ferdi's suggestion, I think it would be okay to simply show the full title

PR #961, opened 2022-11-04
fix: expand/collapse arrow removed and full title shown

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

closes https://github.com/mitodl/ocw-hugo-themes/issues/955
ref: https://github.com/mitodl/ocw-hugo-themes/issues/955#issuecomment-1301280575

#### What's this PR do?
- Removes expand/collapse arrow
- Full content title shown

- Checkout this branch or view netlify deployed version
- Go to resources/courses page which have long titles
- Verify that full title is being shown 

image

image

### DIFF
--- a/base-theme/layouts/partials/content_header_v2.html
+++ b/base-theme/layouts/partials/content_header_v2.html
@@ -4,7 +4,7 @@
       {{ with .Params.parent_title }}
       <div class="course-content-parent-title">{{ . }}</div>
       {{ end }}
-      <h2 class="font-weight-bold pb-1 mb-1">{{ .Title }}</h2>
+      <h2 class="pb-1 mb-1">{{ .Title }}</h2>
     </div>
   </div>
 </header>
--- a/base-theme/layouts/partials/content_header_v2.html
+++ b/base-theme/layouts/partials/content_header_v2.html
@@ -1,27 +1,10 @@
-{{ $shouldCollapseTitle := gt (len .Title) 120 }}
 <header>
   <div class="course-section-title-container ">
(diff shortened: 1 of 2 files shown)


Ferdi:
All caps makes it hard to read

MAbdurrehman12:
> All caps makes it hard to read

I agree. Thanks for pointing it out.
Please have another look.

Ferdi:
Looks good

Unrelated : titles look too strong imo

gumaerc (approved):
@Ferdi I think this is fine to merge, but I noticed that long titles take up a whole lot of the screen on mobile:

chrome_2022-11-07_11-50-24

Maybe the right approach is to only collapse on mobile? Either way, I don't think long titles like this are super common and to be honest it seems like a content issue i.e. why does the title of this page need to be so long?

MAbdurrehman12:
> Unrelated : titles look too strong imo

We are using `h2` tag for titles and it has it's own font-weight that is I think around 600, and if we reduce it, it looks something like this: 
image

and if we do this then I think we will also have to override the `font-weight` of other headings (h1, h3, h4 etc) to maintain consistency, else `h2` will look lighter than `h3` as it can be seen here:
image

@Ferdi Let me know what are your thoughts on this?

MAbdurrehman12:
> to be honest it seems like a content issue i.e. why does the title of this page need to be so long?

haha, I agree. I also had this in mind but just couldn't say it :P

pdpinch:
Since Ferdi's comment about title strength is "unrelated" I suggest we hold it for a separate issue / PR. 

That said, I agree that titles could be lighter across the board.

Issue #938, opened 2022-10-31
add links to the course download page


gumaerc:
As a user of OCW, I'd like to see a link to download course materials

#### Designs and Mockups

https://invis.io/UX132JLIWE43#/470179934_OCW_-_Course_Home_V2-1
https://invis.io/6N12T6YA43PW#/467901625_OCW_-_Course_Detail_-_Sidebar_Alt_-__v-1-4

#### Acceptance Criteria:

- [ ] The course home page has a link to the resources index as indicated in the design above in the first link
- [ ] In the course info sidebar on other pages, there is also a link to the resources index as indicated in the design in the second link above
- [x] The existing page at `/resources` is instead rendered at `/download`

Ferdi:
I think we should rename /resources page to /download

gumaerc:
@Ferdi What do we expect to see at the root of `/resources` vs `/download`?

Ferdi:
I was hoping we can have `/download`  replaces `/resources`   but leave  `/resources/lecture-videos` etc. as is.   Is that possible ?

gumaerc:
Yep, that's controlled by [this line](https://github.com/mitodl/ocw-hugo-projects/blob/9e5d311441b9c6b4d8af67fd951009e32ded61b1/ocw-course-v2/config.yaml#L48) in the Hugo config, so as long as we don't change that the "see all" pages will stay the same at `/resources`. I can certainly move the `/resources` page to `/download`, but my question is what do we expect to be at `/resources` in its place? Since we have content at `/resources`, the `single.html` template renders the individual resource pages and Hugo automatically tries to render an index page there. It ends up being blank because the only templates matching it are the base template, like this:

![chrome_2022-10-31_15-43-33](https://user-images.githubusercontent.com/12089658/199096587-e63c667a-61fd-472a-ac52-c33650f79851.png)

So, I'm wondering what we should expect here. A few options that come to mind:

 - 301 Redirect to `/download`
 - A separate template with the resource lists minus the download button
 - 404 / don't render a page

PR #966, opened 2022-11-08
add course download page link button

gumaerc:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/938

#### What's this PR do?
This PR finalizes the work done on course downloads and makes the feature public by adding the buttons to the download page on the course home page and the course info drawer. The link points to `/download`, taking `baseUrl` into account.

 - Configure your `.env` file to spin up a test course (preferably one with a variety of resources) and do so with `yarn start:course`
 - Visit http://localhost:3000 and ensure that the download button on the right side looks correct and links to the download page
 - Visit any other page in the nav (syllabus, lecture notes, etc.) and verify that the link in the course info drawer also looks correct and links to the download page
 - Shut down the dev server and make a temporary edit to `package_scripts/start_course.sh`, adding `--baseUrl http://localhost:3000/test/ \` to the `hugo server` arguments list
 - Visit http://localhost:3000/test/ and follow the same procedure, ensuring that the links all go where they should

chrome_2022-11-09_01-09-07
chrome_2022-11-09_01-09-30
chrome_2022-11-09_01-09-59
chrome_2022-11-09_01-10-17


Ferdi:
In the home page (desktop) there is too much space above the download button

course-v2/layouts/partials/download_course_link_button.html
```diff
@@ -0,0 +1,10 @@
+{{- $trimmedBaseUrl := strings.TrimSuffix "/" site.BaseURL -}}
+{{- $downloadPageUrl := print $trimmedBaseUrl "/download" -}}
+{{- $downloadPageUrl = partial "page_url.html" (dict "context" .context "url" $downloadPageUrl) -}}
+<div class="mt-4">
```
MAbdurrehman12, line 4:
```suggestion
<div class="my-4">
```

Download button sticks to the bottom for pages having less content.  
image

MAbdurrehman12 (changes requested):
Overall it looks good. Just a couple of small things:

1. To address Ferdi's point of too much space above download button on CHP, maybe we can pull it up so it looks something like this: 
image

2. For any course, which doesn't have resources, we are redirected to the download page which is blank:
image
Do we also need to hide this download button too?

gumaerc:
My bad guys, I was mistaken. The design in Invision uses a course that has "other versions" listed in it which pushes the download button down to the bottom of that window, and I incorrectly assumed that we would want it pinned to the bottom.

gumaerc:
@MAbdurrehman12 This is ready for another look. I ended up just putting bottom padding on the button in the course info drawer, as the horizonal rule already adds a margin of 1rem of spacing to the top of the button. Setting `pb-3` on the button makes the spacing uniform.

Ferdi:
The front page screenshot still looks the same (too much space) 

> For any course, which doesn't have resources, we are redirected to the download page which is blank:

is this fixed ?

gumaerc:
@Ferdi should be fixed now

gumaerc:
@MAbdurrehman12: @Ferdi and I discussed this today and we're going to make some further changes. We talked about how it doesn't make sense to fully hide the link to the download page on courses that "have no resources." This is because what that really means is that the course in question doesn't have its `learning_resource_types` tags set to properly create the taxonomies, even if resources do exist in the course. Nonetheless, the offline course ZIP bundle will still be available for download. The plan is to still link to the page, but ensure that only the course download button is shown. There will be a larger effort to see if there's a way we can correct the issue with some courses missing their `learning_resource_types` tags like this, but that will be the strategy for now. I'm going to do this work tomorrow.

gumaerc:
@MAbdurrehman12: This is ready for another look, but I have temporarily hard coded the ZIP URL to demonstrate how the page is supposed to look for purposes of demonstration. You may also want to read below.

@Ferdi: Question for you:

You can view the deploy preview for this PR here: https://ocw-hugo-themes-course-v2-pr-966--ocw-next.netlify.app/

As you can see in the deploy preview, the download course link button is there in the sidebar and it brings you to the download page, which has the download course button at the top and the usual individual resource drawers below that. As I mentioned above I've hard coded the ZIP URL for purposes of demonstration on Netlify. The code in this PR will now not hide the link button if `learning_resource_types` are not set, like we discussed. However, in the case that the ZIP is also not available, I still have it set up to display "no resources found." This is what it looks like:

No `learning_resource_types`, but ZIP exists:
chrome_2022-11-16_15-16-39

No `learning_resource_types` and ZIP also does not exist:
chrome_2022-11-16_15-23-40

My questions:
 - Instead of showing "No resources found", should we hide the "download course" link buttons in the sidebar in the case that no `learning_resource_types` are set and no ZIP exists?
 - If we don't hide the button, should we have a more "human friendly" message here?
 - Both?

MAbdurrehman12:
@gumaerc It looks good! 

Couple of thoughts:

> I've hard coded the ZIP URL for purposes of demonstration on Netlify

I think it's fine for demonstration purposes as far as we change it back, though we can also change the course built on netlify in `deploy.yml`, this seems a bit safer approach to me.

> Instead of showing "No resources found", should we hide the "download course" link buttons in the sidebar in the case that no learning_resource_types are set and no ZIP exists

I think we should. Otherwise clicking the download button from the sidebar will not be valuable/useful for the user since it will take the user to an empty page.

Another random thought: Just wondering how will it look, if, in case of no ZIP, we make the download button blurred and unclickable and have a message next to it "Download not available yet" etc. 
Will this avoid the confusion (if any) among the users that some courses have a download button and some don't?

gumaerc:
@MAbdurrehman12 I implemented the following changes:

 - Broke out the code that checks if a file exists at a URL into its own partial at `base-theme/layouts/partials/file_exists_at_url.html`
 - Put the code for generating the course zip download URL into its own partial as well under `course-v2/layouts/partials/get_course_download_url.html`
 - Download page link in the sidebar now only shows up if the ZIP is available *or* there are resources detected through the taxonomy
 - On the download page itself, the download course button will only appear if the ZIP file actually exists
 - If the ZIP is unavailable *and* there are no resources detected, the "no resources found" message will still be rendered to the download page, but there will be no link to it

I think this is good enough for now. If we want anything more fancy looking when no resources are detected we will need a design for that and it can be done in a later PR. I removed the hard coded URL I had in there for demonstration purposes, but if you want to test this locally the easiest way is to manipulate what's returned from the `file_exists_at_url.html` and `show_no_resources_found.html` partials. If you change the former to return false and the latter to return true, you should not see the buttons linking to the download page. If either of them have their values flipped, the link should be there  with the appropriate part that's missing being hidden.

MAbdurrehman12 (approved):
@gumaerc Great stuff 👍

pdpinch:
Sorry, I'm not seeing a download page with a download button on https://ocw-hugo-themes-course-v2-pr-966--ocw-next.netlify.app/ -- should I be?

Issue #949, opened 2022-11-02
Lots of links have extra spaces around them


ChristopherChudzicki:
*Re-wrote this to make more clear that multiple things are affected.*

A **lot** of content in OCW has extra whitespace (try viewing the raw HTML). This usually isn't visible to users, but in a few places it is. Here are some examples:

- **Instructors in Featured Course Cards:** on https://ocw.mit.edu

    <img width="200" alt="Screen Shot 2022-11-02 at 10 13 39 AM" src="https://user-images.githubusercontent.com/9010790/199512681-584a6ca9-ebd3-4d87-8e3e-fbbbc96e0f05.png">
- **After "FAQs"on Download page:** https://ocw.mit.edu/courses/res-18-010-a-2020-vision-of-linear-algebra-spring-2020/download/

    <img width="500" alt="Screen Shot 2022-11-07 at 9 58 42 AM" src="https://user-images.githubusercontent.com/9010790/200341917-64074690-3fd9-44b9-9636-a2a67efb1b75.png"> 
- Add more examples...

### Scope
Basically everything using `link.html` partial has this issue, though it's only noticeable if there's a non-whitespace character after or before the link (punctuation or a parenthesis, usually).

Probably an issue with other shortcodes, too.

gumaerc:
@ChristopherChudzicki I think this is reasonable. We should likely have a PR that goes through the whole repo and sets the whitespace stripping delimiters wherever it makes sense to do so.

Wassaf-Shahzad:
@ChristopherChudzicki Should the PR to convent `{{ }}` to `{{- -}}` be a part of this issue or a separate issue ?

In regards to this issue, I have identified that the cause for extra spaces was the delimiter we were adding through hugo rather then adding it using CSS. I have a PR up for this issue resolving issue in `course_card_topics` and `course_card_instrutors`

ChristopherChudzicki:
> Should the PR to convent {{ }} to {{- -}} be a part of this issue or a separate issue ?

@Wassaf-Shahzad I would *expect* yes, because I think that's really the underlying problem.

I'd be happy to look a branch with your CSS fix, but I suspect it's specific to the instructor list, whereas the instructor list issue really just illustrative of a bigger issue.

For example, below there is an extra space between "FAQs" the period ".", and we probably don't want to use CSS in this case.

[...]

PR #968, opened 2022-11-08
Fixed issue with course card links caused by delimiter.

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Part of #949 

#### What's this PR do?
- Removes delimiter "," which were added in html template.
- Added css class to add delimiter

- Run ocw-www
- Verify topics in course card have valid spacings.
- Verify instructors in course card ave valid spacings.
- Verify course levels ave valid spacing.
- Verify any other partial using link.html remains same

As per [the discussion](https://github.com/mitodl/ocw-hugo-themes/issues/949#issuecomment-1305732892), This PR just fixes the problem in course cards which is not caused by using ` {{ }} ` but because of the delimiter.
(Optional)
Screenshot 2022-11-10 at 12 33 26 PM

### DIFF
--- a/base-theme/layouts/partials/course_cards_instructors.html
+++ b/base-theme/layouts/partials/course_cards_instructors.html
@@ -4,9 +4,6 @@
     {{ $title := $instructorObject.title | default "" }}
     {{ $instructorName := $instructorObject.instructor | default $title }}
     {{ $instructorSearchUrl := partial "get_search_url.html" (dict "key" "instructors" "value" $instructorName) }}
-    {{ partial "link.html" (dict "href" $instructorSearchUrl "text" $instructorName "stripLinkOffline" true) }}
-    {{ if ne (add $index 1) $numInstructors }}
-    ,&nbsp;
-    {{ end }}
+    {{ partial "link.html" (dict "href" $instructorSearchUrl "text" $instructorName "stripLinkOffline" true "class" "link-items") }}
   {{ end }}
 {{ end }}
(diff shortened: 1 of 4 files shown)


ChristopherChudzicki (changes requested):
The CSS approach is very cool, but I have two concerns:

First: I don't really think it addresses the underlying issue, namely that the way we've been using Go templates in our Hugo partials/shortocdes is adding extra whitespace. 

Second: Using CSS to insert the commas also influences how the site behaves a bit. For example, try highlighting the text on the card and you'll notice that the commas are not highlighted:

Screen Shot 2022-11-10 at 12 06 06 PM

The commas aren't selected because they aren't part of the HTML content.

The fact that the commas aren't part of the HTML content also has other consequences. E.g., try having a screenreader read that text and its inflection will be wrong. It rushes through the text rather than pausing at the commas.

All in all, I think it's probably better to address this by changing how we're using Go templates. I think the relevant templates to change are:
- base-theme/layouts/partials/link.html
- base-theme/layouts/partials/course_cards_instructors.html
- and probably whatever partial does the "topics" part of the card, too.

I suggest changing those to use {{- and -}}.

BTW: I am finding this site to be helpful... https://gotemplate.io/

Wassaf-Shahzad:
Closing this PR since this issue is resolved with merge of [this PR](https://github.com/mitodl/ocw-hugo-themes/pull/969)

Issue #973, opened 2022-11-10
Update Prettier for better TS support.


ChristopherChudzicki:
We should use Prettier v2, not v1. There are a few Typescript syntax things that Prettier v1 does not support, e.g., `import type { } ... from ...`.

PR #981, opened 2022-11-16
Update Prettier for better TS support.

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing

#973 

#### What's this PR do?
This PR is a major version change of Prettier (v2)
- v2 supports more typescript syntax (notably, `import type` and `export type`, which v1 does not support)
- There are some formatting changes from v1 to v2 of prettier. I sort of prefer some of the v1 formatting, but mostly I don't care and I want support for importing types. If anyone does care, we can presumably use some eslint rule to override prettier's choices. But not worth the effort imo.

1. Introduce a formatting error somewhere and save (depending on editor + extensions, may need to "Save without formatting").
2. Run `yarn fmt:check`
3. Run `echo $?` to view the exit code of the previous process. It should be 1.


collinpreston (approved):
Works well for me.  I tested the happy path as well as creating some formatting errors.  Looks good!

Issue #982, opened 2022-11-16
Video player overflows duration


ChristopherChudzicki:
### Steps to Reproduce

1. Visit https://ocw.mit.edu/courses/8-03sc-physics-iii-vibrations-and-waves-fall-2016/pages/part-i-mechanical-vibrations-and-waves/lecture-2/, and view lower left of video player:

<img width="200" alt="Screen Shot 2022-11-16 at 11 39 53 AM" src="https://user-images.githubusercontent.com/9010790/202240891-91f4dadb-c333-442d-9a10-2c725f5f5763.png">

### Expected behavior:
Duration "1:16:33" is on a single line.

### Actual behavior
The duration is confusingly split across multiple lines.

### Scope
Unsure how many courses this affects, but seems like it might be a lot.

PR #984, opened 2022-11-17
Display time on single line, don't clash with vol

collinpreston:
Removes defined width so that the timestamp will always display on a single line in the video player.  Also removes the padding specified which made the volume slider and time elapsed hit each other.

#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

https://github.com/mitodl/ocw-hugo-themes/issues/982

#### What's this PR do?
Ensures that the video duration is completely displayed when the time is X:XX:XX.

1. Pull this branch.
2. gh repo clone github.mit.edu/ocw-content-rc/18.s096-fall-2013.
3. Set `OCW_TEST_COURSE=18.s096-fall-2013` in your .env file for `ocw-hugo-themes`.
4. Run `yarn start course` from the root directory of `ocw-hugo-themes`.
5. Navigate to http://localhost:3000/resources/lecture-1-introduction-financial-terms-and-concepts/ and ensure that the video duration is correctly displayed.
6. Activate the volume slider and ensure it does not overlap with the video duration.

The duration is hidden when viewed on a mobile screen.

#### Screenshots

**Before:**
Screen Shot 2022-11-17 at 7 26 30

**After**
Screen Shot 2022-11-17 at 7 26 24

### DIFF
--- a/course/assets/css/videojs_player.scss
+++ b/course/assets/css/videojs_player.scss
@@ -32,10 +32,6 @@ $slider-color: $pink;
     margin: 2.1em 0.45em;
   }
 
-  .vjs-control {
-    width: 3.5em;
-  }
-
   .vjs-volume-panel {
     min-width: 50px;
   }
@@ -74,7 +70,6 @@ $slider-color: $pink;
 
   .vjs-time-control {
     font-size: 1.4em;
-    padding-left: 0.3em;
     padding-right: 0.3em;
 
     @include media-breakpoint-down(xs) {
--- a/course/assets/css/videojs_player.scss
+++ b/course/assets/css/videojs_player.scss
@@ -32,6 +32,10 @@ $slider-color: $pink;
     margin: 2.1em 0.45em;
   }
 
+  .vjs-control {
+    min-width: 3.5em
+  }
+
   .vjs-volume-panel {
     min-width: 50px;
   }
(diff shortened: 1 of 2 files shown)


course/assets/css/videojs_player.scss
```diff
@@ -32,10 +32,6 @@ $slider-color: $pink;
     margin: 2.1em 0.45em;
   }
 
-  .vjs-control {
-    width: 3.5em;
```
ChristopherChudzicki, line 36: I certainly don't fully understand the purpose of all the CSS here, but I *suspect* that the `width` value was supposed to guarantee some sort of minimum width.

I tried `min-width: 3.5em` and that seemed to negate the need for the padding change below. (And avoiding the padding change might be good. It seemed to me that with the padding change dropped, the space around the "/" got bigger, and it's already weirdly big.)

Consider both of the points above minor suggestion... Priority is certainly making the whole time visible. 👍

ChristopherChudzicki (changes requested):
Looking good. I left one suggestion. In addition to that:

We have two course themes: "course" and "course-v2". The latter, "course-v2", is the theme that we actually use. We should get rid of the old one. (See https://github.com/mitodl/ocw-hugo-themes/issues/986)

**Requested Change:** Change the css in course-v2. *I guess change it in "course" also, but I don't think we really care right now.*.

#### Additional Comments
- your should also change your env variable that controls the course theme (COURSE_HUGO_CONFIG_PATH).
- Depending on your directory structure, you can probably just delete your env file. See https://github.com/mitodl/ocw-hugo-themes#running-sites-courses-and-ocw-www
- Re the testing instructions, no need to clone or change an env variable. Just run `yarn start course 18.s096-fall-2013` and you will be prompted to clone and the cli arg will override the env variable

ChristopherChudzicki (approved):

Issue #994, opened 2022-11-21
Filtering Search Results by High School level Doesn't Work


pt2302:
### Steps to Reproduce

Navigate to https://ocw.mit.edu/search, and attempt to filter courses by "High School" level. It is currently not possible to do so.

### Expected Behavior

It should be possible to filter courses by "High School" level, since this was added as a level in https://github.com/mitodl/ocw-hugo-projects/pull/221.

PR #995, opened 2022-11-21
Adding High School to allowed list for filtering search results

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested
  - [X] Changes have been manually tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/994.

#### What's this PR do?
Allows for filtering search results by "High School" level.

Set `SEARCH_API_URL=https://open.mit.edu/api/v0/search/` in the `.env` file for `ocw_hugo_themes`, and start an instance of OCW Studio locally. Start a browser with CORS security disabled, and navigate to `http://localhost:3000/search/`. Verify that filtering by `High School` level works as intended.

### DIFF
--- a/www/assets/js/lib/constants.ts
+++ b/www/assets/js/lib/constants.ts
@@ -468,7 +468,7 @@ export const FACET_OPTIONS: Facets = {
   department_name: Object.values(departments).map(
     department => department.title
   ),
-  level:               ["Undergraduate", "Graduate", "Non Credit"],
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki (approved):
👍 

In the future, we should look into changing how `sanitizeActiveFacets` works so as not to need these hardcoded values. (Rather, we should need to hardcode the value in either ocw-hugo-themes or ocw-hugo-projects, not both.)

Issue #980, opened 2022-11-16
Make ocw-www test course more realistic


ChristopherChudzicki:
Right now it is of limited usefulness because it has e.g., no featured courses or stories.

OCW WWW test course should have:
- feaured course cards
- instructors
- build without pointing env vars (e.g., STATIC_API_BASE_URL) to RC

ChristopherChudzicki:
@gumaerc I hit a bit of a snag with this. I was trying to NOT use the rc environment variables in the playwright tests so that `ocw-ci-test-course` could be one of the featured courses in `ocw-ci-test-www`. But it seems to me that:

- building ocw-www requires all featured courses to already be built
- building courses requires ocw-www to already be built (e.g., the course build pulsl instructor data from ocw-www)

So... it seems like there's some circular dependency in the build. I think maybe we would see it in production if you (1) added new instructors, (2) added a new course with the new instructor, (3) made the new course a featured course on www, and (4) then tried to publish both.

Does the above sound accurate?

gumaerc:
@ChristopherChudzicki This situation would be prevented by `ocw-studio` because it does not allow you to assign an instructor to a course if that instructor data hasn't been published. Basically, if there hasn't been a live publish since the instructor was added, then it won't show up in the list

The same goes with featured courses, the won't show up in the relation widget unless they have been built and published live. This does present a conundrum for this particular situation though where we are starting from nothing and want to just run two builds. If this were done in `ocw-studio`, you would build the `ocw-www` site with the instructor but without the featured course and publish it, then publish the featured course with the instructor linked, then add the featured course in `ocw-www` and republish.

ChristopherChudzicki:
> This situation would be prevented by ocw-studio because it does not allow you to assign an instructor to a course if that instructor data hasn't been published. Basically, if there hasn't been a live publish since the instructor was added, then it won't show up in the list

Ah, that makes sense!

I will ponder this conundrum a little more...

ChristopherChudzicki:
<img src="https://i.imgflip.com/715415.jpg" />

ChristopherChudzicki:
@gumaerc Isn't this potentially problematic for our "sync prod db --> rc db, republish all of rc" plan? I guess rc was close enough to prod last time you did the sync that we didn't run into this issue?

gumaerc:
[...]

PR #969, opened 2022-11-09
Replaced hugo {{ }} tag with {{- -}}

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Closes #949 

#### What's this PR do?
Replaces `{{}}` with `{{- -}}`

#### How should this be manually tested?
Verify partials with replaces hugo tags remains unchanged.

#### Any background context you want to provide?
As per this [discussion](https://github.com/mitodl/ocw-hugo-themes/issues/949#issuecomment-1305732892), This PR replaces all `{{ }}` with `{{- -}}`. The following regex was used to determine the valid tags `>\{\{\s*\$\w+\s*\}\}\s*<`.

PR #1001, opened 2022-11-22
Better isolate e2e sites from RC + add featured courses

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#980 

#### What's this PR do?
Previously the e2e tests were pulling in a fair amount of data from RC (e.g., featured courses, instructors, images, new courses). Now we set `STATIC_API_BASE_URL` to a local value, so the e2e test sites are built more in isolation.

Some data is still being pulled from RC (e.g., "New Courses" and images). The infrastructure in this PR should also be helpful for dealing with that. _Although, using RC images may be desirable._

I also changed a bit how the `global-setup.ts` file serves the sites. Previously each site was served by a separate http server. Now all sites are served by one server, so behaves much more like rc/prod.

`yarn test:e2e` should pass locally and on its github action.

Start with [test-sites/__fixtures__/README.md](https://github.com/mitodl/ocw-hugo-themes/blob/a92d46f0b077aefd3b87a6052ec418159a512041/test-sites/__fixtures__/README.md), which explains a thorny issue the PR tackles.

### DIFF
--- a/test-sites/__fixtures__/README.md
+++ b/test-sites/__fixtures__/README.md
@@ -22,7 +22,7 @@ then publishing will be deadlocked: the www build will fail for lack of course X
 
 Thus, it is likely that a **snapshot**[^1] of the OCW Markdown at a single point in time cannot be built in isolation by Hugo.
 
-[^1]: OCW Studio forces you to build the sites incrementally: Course A cannot be published until a WWW version has been published which includes the relevant instructors, and a course cannot be featured on WWW until it has been published.
+[^1]: OCW Studio forces you to build the sites incrementally: Course A cannot be published until a www version has been published which includes the relevant instructors, and a course cannot be featured on www until it has been published.
 
 ## Building a snapshot for e2e testing: `test-sites/__fixtures`
 
--- a/test-sites/__fixtures__/README.md
+++ b/test-sites/__fixtures__/README.md
@@ -1,0 +1,31 @@
+# Static API Mocks
+
+## Static API
+Our Hugo builds output json data files, and other builds consume these data
+files. For example:
+ - **Building ocw-www:**
+   - **produces** instructor data, e.g., https://ocw.mit.edu/instructors/a32373bc-8c78-e37a-f79e-1f15686d1a9b/index.json
+   - **consumes** course summary data (e.g., for the "Featured Courses" carousel)
+- **Building course sites:**
+  - **produces:** course summary data, e.g., https://ocw.mit.edu/courses/8-01sc-classical-mechanics-fall-2016/data.json
+  - **consumes:** instructor data (e.g., for the "Course Info" on course homepage)
+
+These data comprise our "static API". When Hugo consumes this data during the build, it does so via requests through `STATIC_API_BASE_URL`.
+
+## Circular dependencies when building a snapshot
+Because course builds depend on www data, and www builds depend on course data, it is possible (and likely) for builds to contain circular dependencies. For example, if
+ - Course A uses instructor X, and
+ - ocw-www markdown describes instructor X
+ - ocw-www features Course A as a featured course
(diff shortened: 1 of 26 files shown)


tests-e2e/ocw-ci-test-www/featured_courses.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { test, expect } from "@playwright/test"
+import { WwwPage } from "../util"
+
+test("Course Card lists Instructors", async ({ page }) => {
+  const www = new WwwPage(page)
+  await www.goto()
+  const card = www.getCourseCard("OCW CI Test Course")
+  await expect(card).toContainText(
+    "Instructors(s): Prof. Tester One, Dr. Tester Two, Another Tester Three"
```
ChristopherChudzicki, line 9: @Wassaf-Shahzad Here's the e2e test I was thinking of adding from your work in #969!

Previously there were no featured courses in ocw-ci-test-www`. 😦

test-sites/__fixtures__/README.md
```diff
@@ -0,0 +1,31 @@
...
+ - neither Course A nor ocw-www have ever been published
+then publishing will be deadlocked: the www build will fail for lack of course X, and the course X build will fail for lack of instructor data.
+
+Thus, it is likely that a **snapshot**[^1] of the OCW Markdown at a single point in time cannot be built in isolation by Hugo.
+
+[^1]: OCW Studio forces you to build the sites incrementally: Course A cannot be published until a WWW version has been published which includes the relevant instructors, and a course cannot be featured on WWW until it has been published.
+
+## Building a snapshot for e2e testing: `test-sites/__fixtures`
+
+For e2e testing we *do* want to build a snapshot of OCW markdown in isolation. But there's really only a circular dependency if the `STATIC_API_BASE_URL` is the Hugo output. So for e2e testing, we pre-construct the relevant JSON files (stored in `test-sites/__fixtures__`) and serve them for use during the Hugo builds.
```
ChristopherChudzicki, line 29: I have two overall thoughts here:
1. This circular dependency seems bad for OCW in general, not just the e2e tests. I understand that Studio mitigates it, but I suspect that if we created a new "staging" environment (as opposed to "rc"), and synced the database from prod -> staging, we'd have trouble publishing.  @gumaerc I think we did NOT run into this when you did the prod -> rc sync because rc was already "close enough" to prod. Does that sound right?

    (I'm not suggesting this is a big problem or something we should tackle right now.)
2. There are probably better ways to solve this issue for both the e2e tests AND "real" ocw. One idea would be:  Make the build tolerate missing data more consistently, then we could build twice: once would produce the json files (and not-quite-correct html), then the second pass would produce correct results for both.

tests-e2e/global-setup.ts
```diff
@@ -1,75 +1,145 @@
...
-const buildSite = (name: string, configPath: string) => {
+const buildSite = (name: string, destInTmp: string, configPath: string) => {
   return hugo(
     {
       themesDir:   fromRoot("./"),
-      destination: fromRoot(`./test-sites/${name}/dist`),
-      baseURL:     `http://localhost:${env.WEBPACK_PORT}`,
-      config:      configPath
+      destination: fromTmp("dist", destInTmp),
+      cacheDir:    fromTmp("hugo_cache"),
```
ChristopherChudzicki, line 31: At one point when I was working on this, Hugo was caching nil responses. I think I had messed up one of the servers so, e.g., `localhost:4321/instructors/some-uuid/index.json` was returning nil incorrectly. Then Hugo cached that, so when I fixed the server, Hugo was still using the old response.

I went insane, then started deleting the cache, and then started putting the cache in the `tmp` directory (so I could delete it easily), and sanity returned.

PR #1008, opened 2022-12-08
fix: array-flat-polyfill added

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

might or might not close https://github.com/mitodl/ocw-hugo-themes/issues/901

#### What's this PR do?
- Uses `array-flat-polyfill` for `flat()`

- I'm having a hard time testing this change since we need to reproduce and test this locally on older browser [versions](https://caniuse.com/array-flat) that do not support flat and then verify that this change fixes that

### DIFF
--- a/base-theme/assets/index.ts
+++ b/base-theme/assets/index.ts
@@ -14,6 +14,7 @@ import "./js/utils"
 import * as Sentry from "@sentry/browser"
 import { initSentry } from "./js/sentry"
 import PDFObject from "pdfobject"
+import "./js/polyfill"
 
 export interface OCWWindow extends Window {
   $: JQueryStatic
--- a/base-theme/assets/js/polyfill.ts
+++ b/base-theme/assets/js/polyfill.ts
@@ -1,0 +1,1 @@
+import "array-flat-polyfill"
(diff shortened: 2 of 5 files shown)


Wassaf-Shahzad (approved):
LGMT!
although unable to verify it locally, May need another review

ChristopherChudzicki (changes requested):
One request, but feel free to disagree:

IMO, we should put this in the webpack bundle `base-theme/assets/index.ts` so that it's included everywhere. Like mentioned, it's very small. Additionally, I think it would be nice to make a `polyfill.ts` file that handles all polyfills. (Right now, I think that's just this.)

_Rationale: I think it should go in the base theme because part of the point of using a polyfill is so that we don't have to be so careful about whether a browser supports `.flat()` and `.flatMap()`; if we put it in `search.ts`, it will only be supported if that file is loaded, not generally available, so we would still need to be careful._

**Re Testing:** The way I tested this locally was by adding 

```ts
// @ts-expect-error emulating older browser
Array.prototype.flat = undefined
// @ts-expect-error emulating older browser
Array.prototype.flatMap = undefined
```
on `main`, then checking that http://localhost:3000/collections/energy/ throws error on main, but not on this branch.

ChristopherChudzicki (approved):
👍

PR #1010, opened 2022-12-15
fail build on get_instructor errors

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested *We could test this behavior. Opening PR first for some discussion.*

*I should make a ticket, but toward discussion https://github.com/mitodl/hq/discussions/366*

#### What's this PR do?
This PR changes / fixes how we handle errors that occur during `get_instructors.html` getRemote call. See comments for details.

On this branch:
1. `yarn start course`. The course should build successfully.
2. **Simulate an error other than 'not found':** (not an error in response) Edit https://github.com/mitodl/ocw-hugo-themes/blob/ccacdd1d4d6969b490481fc2277f58b8d6a0dfbb/course-v2/layouts/partials/get_instructors.html#L5  Change $url to "meow" or some other invalid url. Your course should now fail to build, and you should get a meaningful error. (You might get one per instructor.) *Alternatively, cause an http error any other way you can think of.*
3. Revert the change from step (2).
4. **Now simulate a 'not found' response:** For example, in your local copy of the course repository you can edit `course.json` and change one of the instructor UUIDs to something bogus.

### DIFF
--- a/course-v2/layouts/partials/get_instructors.html
+++ b/course-v2/layouts/partials/get_instructors.html
@@ -4,8 +4,7 @@
   {{ $url := (print (strings.TrimSuffix "/" $staticApiBaseUrl) "/instructors/" . "/index.json") }}
   {{ with resources.GetRemote $url }}
     {{ with .Err }}
-      {{ $errorMessage := printf "Something went wrong while fetching instructors on %v via %v with error: %v" site.BaseURL $url .Err }}
-      {{ partial "sentry_capture_message.html" $errorMessage }}
+      {{ errorf "Something went wrong while fetching instructors on %v via %v with error: %v" site.BaseURL $url . }}
     {{ else }}
       {{ $data := (. | unmarshal).data }}
       {{ $searchUrl := partial "get_search_url.html" (dict "key" "instructors" "value" (title $data.title)) }}
@@ -13,8 +12,7 @@
       {{ $instructors = $instructors | append $instructor }}
     {{ end }}
   {{ else }}
-    {{ $errorMessage := printf "Failed to fetch course instructors through %v on %v" $url site.BaseURL }}
-    {{ partial "sentry_capture_message.html" $errorMessage }}
+    {{ errorf "Failed to fetch course instructors through %v on %v" $url site.BaseURL }}
   {{ end }}
 {{ end }}
 {{ return $instructors }}


course-v2/layouts/partials/get_instructors.html
```diff
@@ -4,17 +4,15 @@
   {{ $url := (print (strings.TrimSuffix "/" $staticApiBaseUrl) "/instructors/" . "/index.json") }}
   {{ with resources.GetRemote $url }}
     {{ with .Err }}
-      {{ $errorMessage := printf "Something went wrong while fetching instructors on %v via %v with error: %v" site.BaseURL $url .Err }}
```
ChristopherChudzicki, line 7: Because we're in a `with .Err` block, `.` refers to the error already. So we shouldn't be using `.Err`, we should be using just `.`.

course-v2/layouts/partials/get_instructors.html
```diff
@@ -4,17 +4,15 @@
...
-      {{ partial "sentry_capture_message.html" $errorMessage }}
```
ChristopherChudzicki, line 8: Two issues here:
1. This doesn't work at all. The intent of calling `{{ partial "sentry_capture_message.html }}` here is to include a <script> tag in the markup that pings sentry. This isn't clearly documented, but as far as I can tell, Hugo partials can EITHER return a value OR they can include markup. Partials cannot do both, which we were attempting to do here. So it didn't work at all.
2. Even if this did work, including a <script> tag is not really that great. It means we get the error when a user visits the page, not when the hugo build occurs. We can detect the error at build time, so let's do that.

course-v2/layouts/partials/get_instructors.html
```diff
@@ -4,17 +4,15 @@
...
+      {{ errorf "Something went wrong while fetching instructors on %v via %v with error: %v" site.BaseURL $url . }}
```
ChristopherChudzicki, line 7: Calling `errorf` here will cause Hugo to fail.

**Not a behavior change:** We were already failing if this block got hit. (But we'd fail with a message like `can't evaluate field Err in type resource.ResourceError`.)

course-v2/layouts/partials/get_instructors.html
```diff
@@ -4,17 +4,15 @@
...
     {{ else }}
       {{ $data := (. | unmarshal).data }}
       {{ $searchUrl := partial "get_search_url.html" (dict "key" "instructors" "value" (title $data.title)) }}
       {{ $instructor := merge $data (dict "url" $searchUrl) }}
       {{ $instructors = $instructors | append $instructor }}
     {{ end }}
   {{ else }}
-    {{ $errorMessage := printf "Failed to fetch course instructors through %v on %v" $url site.BaseURL }}
-    {{ partial "sentry_capture_message.html" $errorMessage }}
+    {{ errorf "Failed to fetch course instructors through %v on %v" $url site.BaseURL }}
```
ChristopherChudzicki, line 15: This block is reached if Hugo successfully makes the HTTP call, but the response is 404. Hugo treats such cases as "empty", not as errors. [^1]

[^1]: @gumaerc @pt2302 does my description jive with your understanding?

**This is a behavior change:** Previously, we did nothing[^2] when this block was hit. Now we fail the build.

[^2]: We were trying to alert sentry. But we were unsuccessful because, again, partials cannot both return a value and render markup.

ChristopherChudzicki:
@gumaerc @pt2302 Do you feel that the error handling here is reasonable? It is **more aggressive**: More builds will fail now than previously did. *Hopefully not many more.* But IMO that is appropriate. Before 404s on instructors were silently ignored. Causing a build failure is maybe aggressive, but at least now we'll know.

If you think this is too aggressive... Can you think a reliable way to alert sentry (or our ocw publications slack channel) *without* failing the build? 

cc @pdpinch

gumaerc (approved):
Aside from the obvious bugfix with `.Err`, I think given our current build structure it makes sense to fail the build in the given scenario. However, we should adjust the mass build so that when a single course site fails that the fastly cache purge at the end still runs. If the build fails it won't hit the S3 sync, but the whole build gets marked as failed and the Fastly cache purge step gets skipped.

ChristopherChudzicki:
I'm going to wait till https://github.com/mitodl/ocw-studio/pull/1623 is merged to merge this.

PR #1017, opened 2022-12-20
Test build failures

ChristopherChudzicki:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/ocw-hugo-themes/issues/1018

#### What's this PR do?
Adds some tests along the lines of "In scenario X, our build fails with particular error". In particular, for missing instructors. These tests go with the functionality from #1010 

Tests should pass.

Start tests-e2e/jest/build-failures.test.ts

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,4 +1,6 @@
-module.exports = {
+import type { Config } from "@jest/types"
+
+const config: Config.InitialOptions = {
   clearMocks: true,
 
   // A list of paths to modules that run some code to configure or set up the testing framework before each test
@@ -10,7 +12,9 @@ module.exports = {
   ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  testPathIgnorePatterns: ["/node_modules/", "tests-e2e"],
+  testPathIgnorePatterns: ["/node_modules/", "tests-e2e/"],
   preset:                 "ts-jest",
   testEnvironment:        "jsdom"
 }
+
+export default config
--- a/tests-e2e/LocalOcw.ts
+++ b/tests-e2e/LocalOcw.ts
@@ -209,3 +209,4 @@ class LocalOCW {
 }
 
 export default LocalOCW
+export { fromRoot }
--- a/tests-e2e/LocalOcw.ts
+++ b/tests-e2e/LocalOcw.ts
@@ -1,0 +1,211 @@
+import * as path from "node:path"
+import * as http from "node:http"
+import { SpawnOptions } from "node:child_process"
+import execShCb from "exec-sh"
+import handler from "serve-handler"
+import Table from "cli-table3"
+import * as color from "ansi-colors"
+
+import {
+  TEST_SITES,
+  LOCAL_OCW_PORT,
+  siteUrl,
+  TestSiteAlias
+} from "./util/test_sites"
+import SimpleServer, { RedirectionRule } from "./util/SimpleServer"
+import { env, cleanEnv, stringifyEnv } from "../env"
+import { hugo } from "../package_scripts/util"
+
+const execSh = execShCb.promise
+
+const fromRoot = (...pathFromRoot: string[]) => {
+  return path.join(__dirname, "../", ...pathFromRoot)
+}
+
+/**
+ * Redirect all requests /path/to/thing to /ocw-ci-test-www/path/to/thing,
+ * except requests to /courses/.
+ */
+const OCW_WWW_REDIRECT: RedirectionRule = {
+  type:      "rewrite",
+  match:     /^\/(?!(courses\/))/,
+  transform: url => `/ocw-ci-test-www${url}`
+}
+
+interface LocalOCWOptions {
+  /**
+   * All sites will be built to subdirectories of this directory.
+   */
+  rootDestinationDir: string
+  staticApiPort: number
+}
+
+/**
+ * Helps build and serve OCW sites locally.
+ */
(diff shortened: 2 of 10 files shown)


jest.config.ts
```diff
@@ -10,7 +12,9 @@ module.exports = {
   ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  testPathIgnorePatterns: ["/node_modules/", "tests-e2e"],
+  testPathIgnorePatterns: ["/node_modules/", "tests-e2e/(?!jest)"],
```
ChristopherChudzicki, line 15: Slightly awkward: `tests-e2e` now has:
- playwright tests, which are really e2e
- a few jest tests which build with hugo + assert build failed as expected

The jest tests share enough in common with the playwright tests (namely, building via hugo) I think it makes sense to keep them there.

tests-e2e/LocalOcw.ts
```diff
@@ -0,0 +1,199 @@
...
+
+interface LocalOCWOptions {
+  /**
+   * All sites will be built to subdirectories of this directory.
+   */
+  rootDestinationDir: string
+  staticApiPort: number
+}
+
+class LocalOCW {
```
ChristopherChudzicki, line 46: Almost all of this code was previously in `global-setup.ts`, with one exception noted below. I moved it here for re-use in the jest tests. Aside from that, I think this is better organization.

tests-e2e/LocalOcw.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  private makeStaticApiServer = () => {
+    let patchedHandler: http.RequestListener | undefined
+    const server = new SimpleServer(
+      (request, response) => {
+        if (patchedHandler) {
+          patchedHandler(request, response)
+          if (response.headersSent) return
+        }
```
ChristopherChudzicki, line 87: This is the new code that was not in `global-setup.ts` before, our playwright test setup file.

One of the primary things to test with build failures is how our hugo theme reacts to static api request failures. `patchedHandler` let's us override a static api request for a single test.

gumaerc (approved):
:+1: 

This PR does what it says and effective tests the scenario of a Hugo build failure, at least the most common scenario of "I can't fetch the instructor data." The only suggestions I have are organizational / formatting related. A small formatting question is below, but more largely I'm wondering about the organization of the different types of tests we have here and how we might make things less confusing, maybe in a separate refactor.

Currently, we have:
 - Jest tests that test the search app in `www/assets/js`
 - Playwright tests that assert HTML output of Hugo sites in `tests-e2e`
 - Jest tests that assert things that may happen during the Hugo build in `tests-e2e/jest`

The above tests are run through the following package scripts, respectively:
 - `yarn test`
 - `yarn test:e2e`
 - `yarn teste:e2e:jest`

I wonder if we can do a better job organizing this? I feel like `yarn test` should be a command that runs all of the tests in the project, and the search app tests should have their own command. I can imagine something like `yarn test`, `yarn test:search`, `yarn:test:e2e` and `yarn test:e2e:jest`. I'm also not certain that `e2e:jest` is descriptive enough about why those tests are separated out from Playwright and run in Jest instead. The name describes the technology used to run the tests, not what they're actually for. We don't call the Playwright tests `yarn test:e2e:playwright`, but if we're going to call out Jest specifically like that, maybe calling out playwright as well would make the separation more clear? Maybe a `yarn test:e2e` command should run something like `yarn test:e2e:playwright && yarn test e2e:jest`? Either way, it's a bit beyond the scope of this PR and testing Hugo build failures is a big win.

tests-e2e/global-setup.ts
```diff
@@ -1,157 +1,16 @@
...
-      [
-        "\n",
-        color.bold("Now serving the following sites:\n"),
-        table.toString(),
-        "\n"
-      ].join("")
-    )
-  }
-}
+import LocalOCW, { fromRoot } from "./LocalOcw"
```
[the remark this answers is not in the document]
ChristopherChudzicki: Re PascalCase:
1. PascalCase is common for classes. (And also for react components, even though they are more commonly functions now adays).
2. It's common for filenames to have the same name as their default export *I probably picked up the convention from https://github.com/airbnb/javascript#naming--filename-matches-export)

> ... so we don't have a mix of both because that's confusing.

We do generally follow the two rules above. It's most obvious in our react code, because that's where we use default exports the most. The kebab-cased js files that we have do strike me as a little odd.


BTW: The above applies to JS or TS equally. (Two TS-specific places where PascalCase is common are the names of types and the names of enum members).

ChristopherChudzicki:
> The above tests are run through the following package scripts, respectively:
>
> `yarn test`
> `yarn test:e2e`
> `yarn teste:e2e:jest`
> I wonder if we can do a better job organizing this? 

Yes...we can. I definitely agree that `yarn test` should run all the tests. 

I will probably merge this before changing the names, but I should have written down:

### Why separate the two sets of jest tests?

- The "search" tests are testing frontend react code (they may test more than just search; I'm not sure). So they run in JSDom; the tests added here do not need to run in JSDom
- The jest tests added here share **a lot** in common with the playwright e2e tests. (abstracted into LocalOCW). But they shouldn't be playwright tests themselves: they don't actually use the browser, and we probably don't want to run the same globalSetup as for the browser-based playwright tests

which led me to the current `yarn test` and `yarn test:jest`

Issue #1036, opened 2023-01-18
Add production sourcemaps


ChristopherChudzicki:
Sentry errors are harder to debug because we lack sourcemaps in production. Sentry says we should add sourcemaps:

> Sentry can un-minify your code to show you more readable stack traces
[Upload Source Maps](https://docs.sentry.io/platforms/javascript/sourcemaps/)

Let's do that.

### Acceptance Criteria:
- [ ] Our build process should automatically upload source maps to sentry
- [ ] This should not impact the JS transferred to users

PR #1042, opened 2023-01-19
produce source-maps in prod build

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Changes have been manually tested *I have not tested that sentry actually reads these source maps; I think we can wait till for QA servers for that.*

#1036 

#### What's this PR do?
This PR changes our ts+webpack configs to generate sourcemaps during the production builds.

**First:** Verify that this does not noticeably affect the size of the production webpack build.
1. Generate an env file if you don't have one. You can do this with `yarn with-env --dev --print-env '' > .env`. *Most of our commands, e.g., `yarn start course`, run with default environment variables, but **only in development**, and we want to run some production-like commands.
2. Clear your webpack dist folder: `rm -rf base-theme/dist`
3. On `main` branch: run production webpack build `yarn build:webpack`
4. On this branch, run production webpack build: `yarn build:webpack`
5. view the file sizes, `ls -lh base-theme/dist/static/js`. The `.js` file sizes should not be noticeably different.

_There should be no noticeable difference in filesizes. The sourcemap build does add a few dozen characters to the end of the `.js` file specifying where the sourcemaps are located in relation to the js file itself._

**Now:** verify that sourcemaps work
1. Build a site for production. Assuming your ocw-content-rc and ocw-hugo-projects directories adjacent to ocw-hugo-themes, `yarn build abs/path/to/9.40-spring-2018  abs/path/to/ocw-hugo-projects/ocw-course-v2/config.yaml`
2. Serve the built course: `npx serve path/to/9.40-spring-2018/dist`.
3. Open the url that `serve` displayed in your terminal and check that the soruce code is viewable:

Screenshot 2023-01-19 at 2 19 17 PM

#### Two places to Turn on Source Maps
There are two places we "turn on" sourcemaps:
1.  Setting `devtool: source-map` in the webpack config tells webpack to generate separate sourcemaps for the multiple *JS* files that it concatenates/minifies/bundles into one (or a few) JS files. This setting affects the JS that is output by the Typescript compiler.
3. Setting `sourceMap: true` in the TS config tells the TS compiler to generate sourcemaps for the compiled JS back to the original TS.

So if you only turn on the webpack sourceMaps, you end up with sourcemaps to the compiled TS, not to the real source code.

#### Should we upload them to Sentry?
We would need to upload sourcemaps if, say, Sentry was monitoring a backend written in Typescript but running as compiled JS. With a browser environment, Sentry can read publicly hosted sourcemaps. They do *recommend* uploading the sourcemaps. From [Sentry Docs, "Hosting Publicly"](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/hosting-publicly/)
> While making source maps available to Sentry from your servers is the most natural integration, it is not always advisable:
>
>  - Sentry may not always be able to reach your servers.
> - If you do not specify versions in your asset URLs, there may be a version mismatch
> - The additional latency may mean that source mappings are not available for all errors. 

I think we should start with hosting publicly: (1) our assets are versioned, (2) fastly/s3 is reliable, and (3) Avoids us having to generate/manage auth tokens to do the uploading.

### DIFF
--- a/base-theme/assets/webpack/webpack.prod.ts
+++ b/base-theme/assets/webpack/webpack.prod.ts
@@ -7,7 +7,8 @@ import MiniCssExtractPlugin from "mini-css-extract-plugin"
 import common from "./webpack.common"
 
 const prodOverrides: Configuration = {
-  mode: "production",
+  mode:    "production",
+  devtool: "source-map",
 
   output: {
     path:       path.join(__dirname, "../../dist/static"),
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,12 +7,8 @@
       "dom"
     ],
     "jsx": "react",
-    "module": "commonjs",                                /* Specify what module code is generated. */
-    // "rootDir": "./",                                  /* Specify the root folder within your source files. */
-    "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */
-    // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
-    // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
-    // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
+    "module": "commonjs",
+    "moduleResolution": "node",
     "typeRoots": [
       "node_modules/@types",
       "base-theme/assets/types"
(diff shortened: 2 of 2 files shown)


ChristopherChudzicki:
This is ready for review, but I am going to hold on merging till I check with @gumaerc about whether we want to exclude the sourcemaps from the offline build zips. They are a bit big (they add ~9MB total to the static/ folder). Browser don't download them (unless dev console is open), online users shouldn't care.

Issue #1037, opened 2023-01-18
Update Sentry


ChristopherChudzicki:
Sentry says we should update it:
> We recommend you [update your SDK from sentry.javascript.browser@v5.30.0 to sentry.javascript.browser@v7.31.1](https://docs.sentry.io/platforms/javascript) (All sentry packages should be updated and their versions should match)

(E.g., here https://sentry.io/organizations/mit-office-of-digital-learning/issues/3683365210/?project=5304953&referrer=github_integration)

May want to do this before #1036

Issue #1034, opened 2023-01-17
Do not load large dependencies (videojs, nanogallery) unless they are needed


ChristopherChudzicki:
Videojs and nanogallery make up a large fraction of the JS bundle that users load when viewing course pages, but those libraries are used on relatively few pages.
```
                      size, gzipped
course_v2 bundle      243kb
nanogallery2          54kb            (22%)
videojs               126kb           (52%)
```

We should only load these where they are needed so that the page is interactive faster.

Issue #1033, opened 2023-01-17
Reduce duplicate css/js in webpack output


ChristopherChudzicki:
We bundle the js/css assets that a user downloads with webpack. Currenlty, users download a few different "bundles". Mostly:
- visiting a course --> download `main` and `course_v2`
- visiting site homepage (ocw.mit.edu) --> download `main` and `www`

This currently results in users downloading a fair amount of duplicated code. A few examples are jquery, popper, and sentry:

<img width="1728" alt="Screenshot 2023-01-23 at 2 42 10 PM" src="https://user-images.githubusercontent.com/9010790/214135770-0d557ab4-1a75-4cee-aaba-3808f6b92c11.png">

### Acceptance Criteria
- [ ] External assets like jquery that we use in multiple themes should only be loaded once
- [ ] avoid manual configuration; e.g., if we add a new dependency, it would be nice for it to automatically go in the appropriate chunk.

ChristopherChudzicki:
tldr: We should use `optimization.splitChunks` to automatically share code for us, not a manually made `main.js`. And we should configure `splitChunks` to ONLY consider "course-v2" and "www" when doing the splitting. That way experimenting with other themes won't affect our production bundles.

### Webpack terminology:
My understanding:
- **entrypoint:** File or files that tell Webpack where to start when following dependency graph to create bundles.
- **bundle:** A collection of files, the final output of webpack. One bundle per entrypoint. 
- **output chunk:** I will use this to mean an individual output file in the bundle *chunk is a webpack-specific term; it also refers to the internal representation of modules while webpack runs. I'll try to say "output chunk".*

The approach we've taken in OCW to sharing common assets between the course theme and the www theme has been to manually specify the shared code in `base-theme/assets/index.ts`, then use multiple entry points:
```js
{
  entry: {
    main:      "/base-theme/assets/index.ts",     //we manually put shared code here
    course_v2: "/course-v2/assets/course-v2.ts",
    www:       ".www/assets/www.tsx"
  }
// plus some other bundles for "fields
}
```
then we load two bundles on each page, "main+www" for www and "main+course_v2" for course sites. 

**Disadvantages:** This has the obvious disadvantage that if we accidentally import (or [provide](https://github.com/mitodl/ocw-hugo-themes/blob/main/base-theme/assets/webpack/webpack.common.ts#L139), in the case of jQuery and Popper) then the dependency in two different entrypoints then it will be included twice.

[...]

PR #1044, opened 2023-01-20
better tree-shaking... compile ts to ESM (before bundling)

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing
  - [x] Changes have been manually tested *I have not tested that sentry actually reads these source maps; I think we can wait till for QA servers for that.*

Related to #1033, #1034

#### What's this PR do?
This PR changes the way we compile TS *before* being bundled by webpack: In our tsconfig, we had `module` set to CommonJS, whereas it should be set to ES6 As the Webpack Typescript docs say,

> **Warning**
> `ts-loader` uses `tsc`, the TypeScript compiler, and relies on your `tsconfig.json` configuration. Make sure to avoid setting [module](https://www.typescriptlang.org/tsconfig#module) to "CommonJS", or webpack won't be able to [tree-shake your code](https://webpack.js.org/guides/tree-shaking).

### How should this be tested

There should be no noticeable changes, other than that the output of `yarn build:webpack` should be a bit smaller.

You could...
1. Check the site still looks ok at on the [netlify builds](https://github.com/mitodl/ocw-hugo-themes/pull/1044#issuecomment-1398735619)
2. If you want:
    - Run `yarn build:wepback`. It should succeed and the js file sizes (`ls -lh base-theme/dist/js`) should be a bit smaller than they are on `main` branch.
    - Run a site in production mode on this branch, e.g., by
        1. Build a site for production. Assuming your ocw-content-rc and ocw-hugo-projects directories adjacent to ocw-hugo-themes, `yarn build abs/path/to/9.40-spring-2018  abs/path/to/ocw-hugo-projects/ocw-course-v2/config.yaml`
        2. Serve the built course: `npx serve path/to/9.40-spring-2018/dist`.
        3. Open the url that `serve` displayed in your terminal and check that the soruce code is viewable:

        Again, there should be no noticeable changes.

### Background Info
The "module" config setting tells Typescript what module syntax to use when compiling TS files into JS files. Specifying ES6 means use ES6 `import`/`export`. **Note:** This ES6 import syntax **does not end up in the final bundles**. Again, from Webpack docs](https://webpack.js.org/concepts/manifest/):

> No matter which [module syntax](https://webpack.js.org/api/module-methods) you have chosen, those import or require statements have now become __webpack_require__ methods that point to module identifiers.

The advantage of compiling TS to JS is better tree-shaking (removal of unused code). It's not a huge difference at the moment:
```
                commonjs    es6
main.js	        323 kb	    313 kb																							
www.js	        580 kb	    516 kb      <-- biggest difference																				
course_v2.js	866 kb	    864 kb																							
```
but it's free, and I expect in the future it could matter more. (For example: I came across this issue while working on #1037, which I have decided to set aside. While working on #1037 , I found that because of [changes](https://github.com/getsentry/sentry-javascript/blob/master/MIGRATION.md#restructuring-of-package-content) in how Sentry organizes their code in v7, we Sentry v7's footprint on our bundle would be 54kb instead of 20kb without this config change.)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
       "dom"
     ],
     "jsx": "react",
-    "module": "commonjs",
+    "module": "ES6",
     "moduleResolution": "node",
     "typeRoots": [
       "node_modules/@types",
@@ -41,6 +41,9 @@
   ],
   "ts-node": {
     "files": true,
-    "swc": true
+    "swc": true,
+    "compilerOptions": {
+      "module": "commonjs"
+    },
   }
 }


ChristopherChudzicki:
If you're interested and want to see this in action with our config, you can check out https://github.com/mitodl/ocw-hugo-themes/tree/cc/ts-esm-experiment. That branch defines three new files (just for testing):

```js
// example_dog.js
export const dog = 'dog'
export const woof = (n: number) => 'woof'.repeat(n)

// example_cat.js
export const cat = 'cat'
export const meow = (n: number) => 'meow'.repeat(n)

// example_entry.js imports from cat and dog
// but does not use all of cat
import { cat } from "./example_cat"
import { dog, woof } from "./example_dog"

console.log(cat, "...?")

console.log(dog, woof(3))
```

Note that the `meow` function is unused. If you run the webpack build:
- if tsconfig `module: commonjs`, `meow` is included in the build output
- if tsconfig `module: ES6`, `meow` will be omitted

Issue #1045, opened 2023-01-20
Remove "Coming Soon" popover code


ChristopherChudzicki:
We have removed all "Coming Soon" popovers (#524 , #530) but the code and dependency never got removed.

Issue #530, opened 2022-03-17
More Educator links


pdpinch:
#### Acceptance Criteria:

- [ ] link "Find open online textbooks from MIT" to `/search/?f=Online%20Textbook`
- [ ] link "Browse MIT instructor interviews by department" to `https://ocwnext.odl.mit.edu/search/?f=Instructor%20Insights`

#### Out of Scope

- Zero Textbook Cost (ZTC) initiatives

Wassaf-Shahzad:
@pdpinch  This can be closed, as it was resolved in PR https://github.com/mitodl/ocw-hugo-themes/pull/522

Issue #524, opened 2022-03-16
Home: remove all "coming soon" pop-ups


pdpinch:
We should have working hrefs for all links on the home page, about page and educator page

#### Acceptance Criteria:

- [ ] "Get Started" on the home page should link to ZenDesk, https://mitocw.zendesk.com/hc/en-us/categories/4414679122971-Get-Started-with-MIT-OpenCourseWare

Wassaf-Shahzad:
@pdpinch 
The PR https://github.com/mitodl/ocw-hugo-themes/pull/522 will handle coming soon pop ups for the one link in home page.
However there are still anchor tags  in educator page which have the coming soon pop up which are 
- Zero Textbook Cost” (ZTC) initiatives
- Find open online textbooks from MIT
- Browse MIT instructor interviews by department

pdpinch:
I opened a new issue #530 for 2 our of 3 of the missing educator links. You can fix them in this PR if you like.

PR #1046, opened 2023-01-20
remove unused "Coming Soon" code

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing

#1045 

#### What's this PR do?
Removes unused dependency + code. The tippy stuff was all removed in #524 and #530 

1. There should be no observable change: Run a course or www and it should behave the same.
2. If you want to double check that we got them all, search for `"tippy"` or `"coming-soon"`. *There is one remaining instance of `"coming-soon"` as a default class in the old course(-v1) theme, which I decided not to touch*.

### DIFF
--- a/course-v2/layouts/partials/partial_collapse_list.html
+++ b/course-v2/layouts/partials/partial_collapse_list.html
@@ -1,5 +1,5 @@
 {{ $params := . }}
-{{ $className := .klass | default "coming-soon" }}
+{{ $className := .klass | default "" }}
 {{ $useLinks := .useLinks | default true }}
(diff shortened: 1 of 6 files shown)


course-v2/layouts/partials/partial_collapse_list.html
```diff
@@ -1,5 +1,5 @@
 {{ $params := . }}
-{{ $className := .klass | default "coming-soon" }}
+{{ $className := .klass | default "" }}
```
ChristopherChudzicki, line 2: All instances of `partial_collapse_list` in course-v2 were passing an explicit `klass` parameter, so the default was never getting used.

Issue #1043, opened 2023-01-20
usage of partialCached results in incorrect URLs with relativeUrls: true enabled


gumaerc:
### Steps to Reproduce

If you enable the `relativeUrls: true` setting in a Hugo configuration when using a theme here in `ocw-hugo-themes`, incorrect URLs are rendered in many places. A prime example is the left nav. If you turn this setting on, you will notice that the links in the left nav at the course home page will have a variable amount of `../` in front of them, or sub pages will only have `./` in front of them. Since these links are rendered with `partialCached`, the relative URL is rendered on one page and then that URL is reused on pages at different directory levels, which results in broken links.

### Expected Behavior

The relative URLs rendered anywhere in the site should correctly point to the page that they are intended to point to

### Actual Behavior

The URL will point to the page relative to whatever page was rendering when the cache was built

gumaerc:
Closing this issue, as after further testing I was unable to replicate the issue again.

PR #1041, opened 2023-01-19
[WIP] use relative urls wherever possible

gumaerc:
This PR will not be merged and just serves to show the changes being prototyped to move to rendering relative URLs wherever possible. The actual PR's for these changes will be broken out into smaller issues / PR's.

Issue #1013, opened 2022-12-16
consolidate online and offline course themes


gumaerc:
Right now we have 3 main use cases for building a course site:
1. To be viewed online as part of [ocw.mit.edu](http://ocw.mit.edu/)
2. To be viewed offline (download version)
3. To create mirror drives (1 above, but offline)

Currently we have 2 themes related to building course sites: `course-v2`, `course-offline`:

 - `course-v2` allows us to accomplish item 1 from above
 - `course-offline` allows us to accomplish 2 and 3 from above

We have all these use cases covered, however, what we have done in `course-offline` is not optimal. We've split places where Hugo is rendering URLs into a set of partials that have a default in `base-theme` that are then [overridden in `course-offline`](https://github.com/mitodl/ocw-hugo-themes/tree/main/course-offline/layouts/partials). This was done to avoid having to change the way case 1 renders URLs while still creating a solution for 2 and 3. By default, Hugo renders links starting with a `/`, which are called "schemaless relative urls of subtype path-absolute" according to the [URL standard](https://url.spec.whatwg.org/#relative-url-string), and an offline site like 2 and 3 need relative URLs to work. Hugo also renders "clean" URLs by default and does not include `/index.html` when linking to pages, so this is appended to URLs in `course-offline`. Generally, this takes the approach of rewriting the URL to be relative during templating before Hugo renders the site as HTML.

The preferred approach for making a Hugo site browsable from the local file system, [according to the docs](https://gohugo.io/content-management/urls/#relative-urls), is to turn on the `relativeUrls` setting in the Hugo config file. This setting is a *post processing* feature that, after all the HTML is rendered, will crawl the site and rewrite URLs that it finds in a [specific set of attributes](https://github.com/gohugoio/hugo/blob/1477d0ba90acd3aa464556d95edab396a518016e/transform/urlreplacers/absurlreplacer.go#L68-L76). We should be using this setting to render our URLs as relative across the board, but there are a few things preventing us from simply turning the setting on:

 - https://github.com/mitodl/ocw-studio/issues/1664 is what we believe to be a bug with the `baseUrl` setting that causes how we use it to be incompatible with turning on `relativeUrls`
 - Links to shared static assets (JS, CSS, images) would be rendered incorrectly
 - Links to other courses would be rendered incorrectly
 - Links in the sitemaps would be rendered incorrectly
 - Hard coded links in shared partials such as `/about` in the header would be rendered incorrectly

[...]

PR #1049, opened 2023-01-20
remove the usage of partialCached

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/1043

#### What's this PR do?
This PR removes the use of `partialCached` where we call partials and replaces that with simply `partial`. `partialCached` was originally introduced back in `hugo-course-publisher` (https://github.com/mitodl/hugo-course-publisher/pull/393) when we were attempting to build all OCW courses as one Hugo site. The performance gains you get from this are very minimal, when the Hugo build is already a small percentage of the time it takes a course build to run. The performance issues we were having at the time were more related to trying to build way too big of a Hugo site. With course builds isolated to individual Hugo builds, the recursive nav is not nearly as big of a problem. The build time is indeed reduced when using `partialCached`, but we're talking on the level of milliseconds. In my testing, courses that saw a ~600ms build time with `partialCached` took around 1200ms to build, so about double the time but the build time was already really low.

The reason behind removing the usage of `partialCached` has to do with the efforts in https://github.com/mitodl/ocw-hugo-themes/issues/1013 to consolidate the differences between the online and offline themes when it comes to URL rendering. We want to turn on relative URLs for all builds to make the HTML more portable. With `partialCached` being used with relative URLs, you end up with incorrect URLs in partials like the left nav. Another option here would be increasing the specificity of the cache key, but ultimately the partial would have to be re-rendered anyway when the URLs end up being different, which they will at each level of subdirectory beyond the root with `relativeUrls` turned on. The already marginal gains will be even more marginal. I think at least at first when turning on relative URLs, we should completely remove `partialCached` and maybe reintroduce it again at a later point if we can find a spot where it will give us significant gains without causing issues with incorrect URLs.

 - Clone the `cg/unified-build` branch of `ocw-hugo-projects` and point to it in your `.env`
 - Start up any course with a multi tiered nav with `yarn start course` (I used `OCW_TEST_COURSE=8.701-fall-2020`)
 - Browse to the site at http://localhost:3000 and click around the items on the nav. You will notice that the nav highlighting doesn't work currently with `relativeUrls` turned on , but clicking the nav items should at least bring you to the correct page.
 - Switch back to the `main` branch in `ocw-hugo-projects`
 - Go back to the site and all the links should still function, and nav highlighting should work again

I noticed this issue while prototyping the move to relative URLs in https://github.com/mitodl/ocw-hugo-themes/pull/1041/. I started rendering relative URLs and on sub pages the wrong amount of `../` would be used to step back to the root on subpages, almost like the nav was being cached on another page and re-used when it shouldn't be because the URLs were different. Initially, removing the usage of `partialCached` solved this problem. Upon further testing, I was unable to reproduce this issue by pointing `ocw-hugo-themes` to `main` with `relativeUrls: true` still turned on. I'm keeping this PR open in case we need it, but I need to do further testing to ensure that we aren't ditching `partialCached` for no reason. It's important to be extra sure that this issue will not show itself because if it does there will be incorrect URLs everywhere which would be very bad.

### DIFF
--- a/course-v2/layouts/_default/baseof.html
+++ b/course-v2/layouts/_default/baseof.html
@@ -10,9 +10,9 @@
     <!-- End Google Tag Manager (noscript) -->
   {{ end }}
   <div class="overflow-auto">
-    {{ partialCached "mobile_course_nav.html" . }}
-    {{ partialCached "mobile_course_info.html" . }}
-    {{ block "header" . }}{{ partialCached "header" . }}{{ end }}
+    {{ partial "mobile_course_nav.html" . }}
+    {{ partial "mobile_course_info.html" . }}
+    {{ block "header" . }}{{ partial "header" . }}{{ end }}
     {{ $isCourseHomePage := (eq .Params.layout "course_home") }}
     {{ block "subheader" . }}{{ partial "course_banner.html" . }}{{ end }}
 
@@ -33,7 +33,7 @@
       </div>
       <div class="row course-cards">
         <div class="col-2 course-home-grid large-and-above-only">  
-          {{ partialCached "desktop_nav.html" . }}
+          {{ partial "desktop_nav.html" . }}
         </div>
         <div class="col-sm-12 col-lg-10 course-home-grid">
           <div class="">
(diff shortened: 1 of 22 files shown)


gumaerc:
Closing this as after further testing, I was unable to reproduce the issues I was having and verified that they are seemingly not due to the use of `partialCached`

Issue #986, opened 2022-11-17
Remove old "course" theme.


ChristopherChudzicki:
We currently have two course themes, "course" and "course-v2". This causes some confusion, especially in regards to bug fixes.

Many of the bugs in "course-v2", e.g., #982, were inherited from "course". So when we fix the bug, do we fix it in both or just v2? And if we fix it in both, should reviewers test the change in both? Should we verify the commit on rc in both? (Probably not).  This is especially confusing for new contributors.

### Acceptance Criteria
- [ ] confirm that no courses on prod/rc are using the theme
- [ ] remove old "course" theme and associated CI actions
- [ ] remove old course config from ocw-hugo-projects

ChristopherChudzicki:
@gumaerc Am I missing anything in the above to-do list?

@pdpinch @gumaerc are we actually ok removing the old theme? I think we should. It has caused confusion in some PRs. (e.g., #984 )

gumaerc:
I think we are good to remove the original course theme at this point. We should preserve a copy of it somewhere in version control in case we ever need to refer to it in the future for historical reasons, but there shouldn't be anything actively using it anymore.

PR #1051, opened 2023-01-24
refactor: old course theme removed

MAbdurrehman12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

closes https://github.com/mitodl/ocw-hugo-themes/issues/986

#### What's this PR do?
- Removes old course theme

- Checkout this branch
- Build any course locally
- Verify that the course builds successfully
- Now test the course thoroughly and verify that everything is on point
- Repeat this for multiple courses


ChristopherChudzicki (commented):
👏 🚀 🔥

ChristopherChudzicki (approved):

ChristopherChudzicki (approved):
@MAbdurrehman12  👍 Looks good.

Do you also want to delete `content_header.html` (and keep `content_header_v2.html`), etc. ? I think that would make sense.

MAbdurrehman12:
> @MAbdurrehman12 👍 Looks good.
> 
> Do you also want to delete `content_header.html` (and keep `content_header_v2.html`), etc. ? I think that would make sense.

Ahh.. thanks for pointing this out. Will do that

www/layouts/resources/single.html
```diff
@@ -4,8 +4,8 @@
 {{ end }}
 <div class="container mt-5">
   <div class="rounded mb-2">
-    {{ partial "content_header.html" . }}
+    {{ partial "content_header_v2.html" . }}
```
ChristopherChudzicki, line 7: I was curious about this change since it's the only change that could conceivably affect the production site.

And it does. But I think that's OK:
- The change seems desirable: we should have done this when v2 is released. Now www resources look more similar to course resources

Before:
Screenshot 2023-01-25 at 7 44 28 AM

Now:
Screenshot 2023-01-25 at 7 44 21 AM

ChristopherChudzicki (approved):

www/layouts/resources/single.html
MAbdurrehman12: > The change seems desirable: we should have done this when v2 is released. Now www resources look more similar to course resources

I agree!

PR #1054, opened 2023-01-24
remove manual, redundant inclusion of 4 deps bundled with nanogallery2

ChristopherChudzicki:
#### Pre-Flight checklist

- [x] Testing

Toward #1033 

#### What's this PR do?
This PR removes 4 unnecessary dependencies from our bundles.

In short: bundling nanogallery2 with webpack is a bit of a pain
- originally, we were doing [this suggested workaround](https://github.com/nanostudio-org/nanogallery2/issues/98#issuecomment-391770383); but it results in duplicating 4 dependencies
- [this workaround](https://github.com/nanostudio-org/nanogallery2/issues/98#issuecomment-492121137) is better because it does not duplicate any dependencies.

Test image galleries with any course that supports them. For example, `2.672-spring-2009`.

1. Run `yarn start course`
2. Visit http://localhost:3000/pages/labs/ ... there will be links to many image galleries.
3. The image galleries won't actually work because the shortcode does not respect RESOURCE_BASE_URL. Temporarily edit the shortcode, `base-theme/layouts/shortcodes/image-gallery.html` so that `itemsBaseURL` points to `https://ocw.mit.edu/courses/2-672-project-laboratory-spring-2009/`:

    ```html
    <div id="{{ .Get "id" }}" class="image-gallery" data-nanogallery2='{ "itemsBaseURL": "https://ocw.mit.edu/courses/2-672-project-laboratory-spring-2009/" }'>
    {{ .Inner }}
    </div>
    ```
4. Check that the image galleries look the same as at https://ocw.mit.edu/courses/2-672-project-laboratory-spring-2009/

Some other courses with image galleries:
```
5.125a-fall-2005 ... under "Projects"
cms.608-spring-2008 under "image gallery"
```


base-theme/assets/webpack/webpack.common.ts
```diff
@@ -78,40 +82,6 @@ const config: webpack.Configuration = {
...
-          { loader: "expose-loader", options: { exposes: ["NGTweenable"] } }
-        ]
-      },
-
-      {
-        test: require.resolve("hammerjs"),
-        use:  [
-          {
-            loader:  "expose-loader",
-            options: { exposes: ["NGHammer"] }
```
ChristopherChudzicki, line 95: In our previous workaround (the code being removed here), nanogallery2 was actually assigning NGHammer to `window`, and then this re-assigns it. With some other webpack changes I'm working on, webpack complained about this **re-assignment**. Which is the immediate motivation for this PR, in addition to removing unnecessary deps.

_In contrast, Nanogallery2 was not assigning NGTweenable globally, which it is supposed to do._

www/assets/www.tsx
```diff
@@ -1,5 +1,3 @@
-import "../../node_modules/nanogallery2/src/css/nanogallery2.css"
```
ChristopherChudzicki, line 1: We include this in "course-v2" theme; it is not needed in www.

PR #807, opened 2022-08-01
fix hot reload in dev

gumaerc:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-hugo-themes/issues/806

#### What's this PR do?
In https://github.com/mitodl/ocw-hugo-themes/pull/787, we upgraded the version of webpack used in our webpack build to version 5 along with a number of dependencies.  A side effect of that is that when running the new dev server, if you make a change to a JS or CSS file, webpack will generate a "hot update" file.  This has the side effect of making the reference in the `.Site.Data.webpack` (`webpack.json`) data template an array instead of a single string.  So, when Hugo goes to render this it throws an error about expecting a string but getting an array:

```
[start:course:hugo] Change detected, rebuilding site.
[start:course:hugo] 2022-08-01 10:04:58.708 -0400
[start:course:hugo] Data changed "/home/gumaerc/Code/ocw-hugo-themes/base-theme/data/webpack.json": WRITE
[start:course:hugo] ERROR 2022/08/01 10:04:58 Rebuild failed:
[start:course:hugo] 
[start:course:hugo] ERROR 2022/08/01 10:04:58 Failed to render pages: render of "home" failed: execute of template failed: template: home.html:68:18: executing "home.html" at <partial "webpack_url.html" $theme.js>: error calling partial: "/home/gumaerc/Code/ocw-hugo-themes/base-theme/layouts/partials/webpack_url.html:3:11": execute of template failed: template: partials/webpack_url.html:3:11: executing "partials/webpack_url.html" at <hasPrefix . "/">: error calling hasPrefix: unable to cast []interface {}{"/static/js/main.js", "/static/main.3b26c1d8902715b1ddb7.hot-update.js"} of type []interface {} to string
```

This PR adds two new partials, `include_js.html` and `include_css.html`, and sets all the places including these files from a `webpack.json` entry to pass the data to these.  The partials check to see if the data is an array or not, and if it is, only include the first item in the list.

#### How should this be manually tested?
 - Read the readme if you've never run a site before locally.
 - Run `npm run start:www`
 - Visit http://localhost:3000 and verify that you see the `ocw-www` home page and there are no errors in the console
 - In `www/assets/css/home.scss`, add the following CSS rule:
[...]

PR #1059, opened 2023-01-25
Deduplicate webpack assets

ChristopherChudzicki:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/ocw-hugo-themes/issues/1033

#### What's this PR do?
This PR changes how we share common resources between different pages on OCW. The overall approach is to make it less manual and have webpack do more code splitting automatically. See file comments for more, and comments in https://github.com/mitodl/ocw-hugo-themes/issues/1033 for some background.

There should be no noticeable change in behavior, other than smaller bundles. Some things you might do:
- run `yarn start www` and make sure things look ok
- run `yarn start course` and make sure things look ok
    - check that videos work!
- run `rm -rf base-theme/dist && WEBPACK_ANALYZE=true  NODE_ENV=production yarn build:webpack` and check out the production bundles

#### More Details

The files that were loaded before/after on sites are (before gzip):
```
# BEFORE.... v1.82, commit cc68e3e
course pages:
    course_v2.css   198 kb     ⌉___ first bundle
    course_v2.js    866 kb     ⌋
    main.css        220 kb     ⌉____ second bundle
    main.js         323 kb     ⌋

www pages:
    www.css        580 kb     ⌉___ first bundle
    www.js         228 kb     ⌋
    main.css       220 kb     ⌉____ second bundle
    main.js        323 kb     ⌋

# NOW:
course pages:
    course_v2.css  260 kb        ⌉
    course_v2.js   778 kb        |--- single bundle
    common.js      250 kb        ⌋

www pages:
    www.css        229 kb       ⌉
    www.js         391 kb       |--- single bundle
    common.js      250 kb       ⌋
```
A roughly 35% decrease in www js/css asset size, and 20% in course. Neat.

### DIFF
--- a/base-theme/assets/webpack/webpack.common.ts
+++ b/base-theme/assets/webpack/webpack.common.ts
@@ -139,19 +139,15 @@ const config: webpack.Configuration = {
   ),
   optimization: {
     splitChunks:  {
-      name:        "common",
       minChunks:   2,
       cacheGroups: {
-        shared: {
-          test: /[\\/]node_modules[\\/]/,
-          name(module, chunks, cacheGroupKey) {
-            const chunkNames = chunks.map(item => item.name)
-            if (chunkNames.includes(entryNames.www)) {
-              return `${cacheGroupKey}_www`
-            }
-            return cacheGroupKey
-          },
-          chunks: 'all',
+        common: {
+          test:   /[\\/]node_modules[\\/]/,
+          name:   "common",
+          chunks: chunk => {
+            const splitChunks = [entryNames.www, entryNames.courseV2]
+            return splitChunks.includes(chunk.name)
+          }
         }
       }
     },
--- a/base-theme/assets/webpack/webpack.common.ts
+++ b/base-theme/assets/webpack/webpack.common.ts
@@ -13,16 +13,31 @@ import packageJson from "../../../package.json"
 const fromRoot = (pathFromRoot: string) =>
   path.resolve(__dirname, "../../../", pathFromRoot)
 
+const entryNames = {
+  courseV1: "course_v1",
+  courseV2: "course_v2",
+  www:       "www",
+  fields:    "fields"
+}
+
 const config: webpack.Configuration = {
   resolve: {
     // Add `.ts` and `.tsx` as a resolvable extension.
     extensions: [".ts", ".tsx", ".js"]
   },
   entry: {
-    main:                fromRoot("./base-theme/assets/index.ts"),
-    course_v2:           fromRoot("./course-v2/assets/course-v2.ts"),
-    www:                 fromRoot("./www/assets/www.tsx"),
-    fields:              fromRoot("./fields/assets/fields.js")
+    [entryNames.courseV2]: [
+      fromRoot("./course-v2/assets/course-v2.ts"),
+      fromRoot("./base-theme/assets/index.ts"),
+    ],
+    [entryNames.www]: [
+      fromRoot("./www/assets/www.tsx"),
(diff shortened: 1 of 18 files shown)


base-theme/layouts/partials/include_js.html
```diff
@@ -1,11 +1,7 @@
 {{ $context := .context }}
-{{ $url := .url }}
-{{ $type := (printf "%T" $url) }}
-{{ if eq $type "[]interface {}" }}
-{{ with index $url 0 }}
-<script src="{{ partial "webpack_url.html" (dict "context" $context "url" .) }}"></script>
-{{ end }}
```
ChristopherChudzicki, line 7: This conditional was added in https://github.com/mitodl/ocw-hugo-themes/pull/807 to deal with a local dev problem. It seems to no longer be an issue.

base-theme/assets/webpack/webpack.common.ts
```diff
@@ -122,7 +139,22 @@ const config: webpack.Configuration = {
...
+  optimization: {
+    splitChunks: {
+      minChunks:   2,
+      cacheGroups: {
+        common: {
+          test:   /[\\/]node_modules[\\/]/,
+          name:   "common",
+          chunks: chunk => {
+            const splitChunks = [entryNames.www, entryNames.courseV2]
+            return splitChunks.includes(chunk.name)
```
ChristopherChudzicki, line 152: This `optimization.splitChunks` bit tells webpack to put anything shared between `www` and `course-v2` entrypoints into a separate chunk, shared between both bundles. The chunk name will be `common.[fullhash].js`. We are responsible for making sure script/link tags are updated appropriately.

Explicitly limited to the `www` and `course-v2` entrypoints to avoid including extra, unecessary stuff. E.g., if we still had `course`(-v1) entrypoint around, then with `minChunks: 2`, VideoJS would have ended up in this chunk, which would be bad for `www`, where it is uncessary.

We could do this more dynamically (more chunks, fewer restrictions), but that complicates the script/link tags. This seems like a good start. See https://github.com/mitodl/ocw-hugo-themes/issues/1033#issuecomment-1401042248 for more details.

base-theme/layouts/partials/include_js.html
```diff
@@ -1,11 +1,7 @@
...
-{{ end }}
-{{ if eq $type "string" }}
-<script src="{{ partial "webpack_url.html" (dict "context" $context "url" $url) }}"></script>
-{{ end }}
+{{ $urls := .urls }}
+{{ range $url := $urls }}
```
ChristopherChudzicki, line 3: `include_js.html` and `include_css.html` now take multiple urls.

gumaerc (approved):
:+1: LGTM!

Some good optimizations here! I also tested this with the offline theme and everything worked just fine. I noticed that the "fields" theme still has a lot of extra crap in its bundle, but I'm hoping we just get rid of that soon. It was an experiment in case we wanted to use Hugo to handle infinite corridor pages, but it doesn't seem like that's going to happen anymore so it just sits there getting more and more outdated.

ChristopherChudzicki:
>  I noticed that the "fields" theme still has a lot of extra crap in its bundle, 
Yes, this was somewhat intentional. I was optimizing for:
- simple-ish
- good splitting between course-v2 and www
- nothing "extra" in common, either now or with future potential changes to "fields" theme.

We could do better with more shared bundles, but then we probably want to list the chunks programmatically rather than by hand.

PR #1071, opened 2023-02-06
feat: implement lazy-loading for videojs

### LINKED ISSUE #648: only load Video and Image Gallery JS when needed
We should focus on removing large, rarely-used dependencies from the main course bundle. I.e., all course pages can load the same JS, except video and image-gallery pages should load extra files. (Videojs and its dependencies, along with nanogallery, make up the vast majority of the course JS bundle. It's at least 75%, but I think closer to 90%).

So, IMO:
### Acceptance Criteria
- [x] Only load VideoJS and its dependencies on pages that use VideoJS
- [x] Only load nanogallery2 and its dependencies (there are none) on pages that use it.

### Suggested Approaches
I have two ideas for how to approach this:

1. **At build time:** Put VideoJS in a separate Webpack entrypoint. When building a page, figure out if uses VideoJS. (E.g., if a partial uses VideoJS, have it write to some context like `.Page.Scratch`.). If VideoJS is used, add the appropriate link/script tags.
2. **Dynamic imports:** Webpack is very good at splitting dynamic imports into separate chunks and loading them only as needed. So if the partials call something like `initVideo()`, a function defined in webpack assets whose body includes `import("video.js").then(() => { /* init code */ })`, then `video.js` (and other potential dependencies) would not be loaded unless `initVideo()` is called.

Option 2 seems a little simpler, but Option 1 is nice because it would be determined at built time, and that really seems like it should be the goal.

ibrahimjaved12:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

#648 

#### What's this PR do?
This PR implements lazy loading for videojs along with dependencies using localStorage to make sure they load only once and when they're needed

(Required)

(Optional)

(Optional)


course-v2/layouts/_default/baseof.html
```diff
@@ -176,6 +176,17 @@
     }
   </script>
 
+  <script>
```
ChristopherChudzicki, line 179: @ibrahimjaved12 glad you picked this up. Will be a really big improvement for OCW.

What would you think of using a dynamic `import()` for this? Something like:
```ts
// course-v2/assets/videojs-imports.ts
// exactly as you have it now

// course-v2/assets/some_other_file.ts
window.initVideoJs = import("./videojs-imports.ts")

// in base-theme/layouts/partials/youtube_player.html
initVideoJs()
```

I think there are two advantages to the dynamic import:
1. webpack will automatically split `course-v2/assets/videojs-imports.ts` into a separate chunk out for us (I'm pretty sure)
2. `import()`ed code should run after first render, so I think the page will be interactive faster

Webpack compiles `import()` statements to its own version (not ES6 imports), but they still use promises. So the "downside" would be that it wouldn't work on IE11. But apparently YouTube doesn't support IE11, and our videos are  YouTube embeds, so that's probably OK.

_Re browser support: We do not have a hard rule. The goal is to generally work on old browsers, and provide full functionality on newer browsers (say, anything beyond IE11). _

ibrahimjaved12:
Closing this as lazy-loading for videojs was already implemented here (with dynamic import approach): https://github.com/mitodl/ocw-hugo-themes/pull/1089

PR #1072, opened 2023-02-08
demo dynamic imports

ChristopherChudzicki:
This is a demo in response to a slack message from @ibrahimjaved12, in regards to https://github.com/mitodl/ocw-hugo-themes/pull/1071#discussion_r1098682971

@ibrahimjaved12 It seems like we have three possible approaches for splitting out the VideoJS stuff:
1. `.Page.Scratch` hugo manipulations
2. inserting a script tag like you do in current version of #1071 
3. dynamic imports

This demo PR is meant to help explore (3).

Question for you... do you have any opinions yet about which approach seems best?

#### What's this PR do?
This PR shows a demo of dynamic imports. 

**HOWEVER**: This proof of concept only works with production build, NOT dev build. I think that in order for dynamic imports to work with the dev server, some aspects of how we set up the dev server need to change.

*The filepaths below are for my system, where my repos live at `~/dev/ocw-hugo-themes`, `~/dev/ocw-hugo-projects`, and `~dev/ocw-content-rc`. Modify as necessary*. 
1. First, make sure you have an env file locally. E.g., `yarn with-env --dev --print-env ''  > .env`. A `.env` file is needed for running the productionb uild.
5. Ensure you have a local copy of course `9.40-spring-2018`. If you don't, you can just run `yarn start course` to clone it. (It's the default course).
6. Run the production build: `yarn build ~/dev/ocw-content-rc/9.40-spring-2018 ~/dev/ocw-hugo-projects/ocw-course-v2/config.yaml`
7. Serve the production build: `npx serve ~/dev/ocw-content-rc/9.40-spring-2018/dist`
8. Visit the given url.
9. In dev console, `initVideoJS()`.
10. In `ocw-hugo-themes`, `ls -lh  base-theme/dist/static_shared/js`. You should see a separate chunk called something like `240.b3db3.js` that contains just the `initVideoJS` code. *Of course, the code in this PR is just "hello world", but the idea would be for all the big videojs files to end up in that separate chunk*.


base-theme/assets/webpack/webpack.prod.ts
```diff
@@ -12,7 +12,7 @@ const prodOverrides: Configuration = {
 
   output: {
     path:       path.join(__dirname, "../../dist/static_shared"),
-    publicPath: "/static_shared",
+    publicPath: "/static_shared/",
```
ChristopherChudzicki, line 15: I think publicPath issues were the cause of the error you mentioned in Slack.

This fixes it for the production build. Fixing it for the development build is going to take a bit more effort, and something I want to talk a bit with Carey about in the office tomorrow. 

I think we'll **need** to address these publicPath issues if we decide to go the dynamic import route. We probably *should* address them either way.

ibrahimjaved12:
From the three options:

1. .Page.Scratch hugo manipulations
2. inserting a script tag like you do in current version of https://github.com/mitodl/ocw-hugo-themes/pull/1071
3. dynamic imports

### Comparisons
**Comparison of Option 1 & Option 2**
I think 1. `.Page.Scratch` is better as compared to 2., because we are using Hugo's way of doing it -- this would mean no `<script>` tag -- which honestly is looking so out of fashion there. As we have already a partial to do this for us if we go for Hugo's way of using Scratch:
`$videojs_url := $webpack.videojs` (using the entry point)
and then a conditional call to partial inside perhaps youtube_player.html: 
`partial "include_js.html"` (with passed dictionary and above context), this will take care of the `<script> `approach currently implemented in the other PR i put up. So this seems a much more relatively elegant way of doing it.

But we'd also need to decide between the tradeoff of using .Scratch vs using localStorage. Both would have the same functionality: if once it is loaded, add a flag variable to make sure it does not import again. We could probably compare the performance of both for that matter. And also make sure of their accessibility.

**Option 1 vs Option 2 vs Option 3**
But out of the three, I do believe that the 3rd option is preferable, there is no manual keeping track of state using dictionary or localStorage, and we also have the benefit of it being asynchronous.

### Testing process
I ran the code using the steps, `initVideoJS()` inside dev console is working. It is also working inside youtube_player.html, if I try to trigger the function using `window.initVideoJS()`

But: After I do `yarn build ...... .....`,  I had this fatal error. Just mentioning.
image

Also:

> 7. In ocw-hugo-themes, ls -lh  base-theme/dist/static_shared/js. You should see a separate chunk called something like 240.b3db3.js that contains just the initVideoJS code. Of course, the code in this PR is just "hello world", but the idea would be for all the big videojs files to end up in that separate chunk.

This is something I've been trying to find the cause of for a while now. Inside my base-theme/dist/static_shared, I do not see js or css directories at all. However I was able to see those files from the live server. Inside dev, I tried removing `CleanWebpackPlugin` but the files would still not appear. And for this case, using prod, they're not appearing either.. inside my local disk. Can you please let me know if I'm missing something.

### Using Dev instead of Production
I think you're right, it might be related to `publicPath`.
Because in dev, I tried to implement the same thing (but with the real imports):
`GET http://localhost:3000/static_sharedjs/course-v2_assets_videojs-imports_js.js net::ERR_ABORTED 404 (Not Found)
`However if we changed localhost:3000 to localhost:3001 (keeping the rest of the url as it is), the file does exist.

ChristopherChudzicki:
@ibrahimjaved12 Thanks for that write up. Overall, I agree very much with

> But out of the three, I do believe that the 3rd option is preferable, there is no manual keeping track of state using dictionary or localStorage, and we also have the benefit of it being asynchronous.

**Regarding `.Page.Scratch` approach:** Re
>  $videojs_url := $webpack.videojs (using the entry point)
and then a conditional call to partial inside perhaps youtube_player.html:
partial "include_js.html" (with passed dictionary and above context)

I think what you describe doesn't actually use `.Page.Scratch`, and this was my initial thought also: make `youtube_player.html` responsible for inserting the script tag. My concern with that is it is possible for a single OCW page to contain multiple videos. (A resource page will only have one, but a `course/whatever-course/pages/page-with-lots-of-videos` might have multiple). And if there are multiple videos, we only want a single script tag. Hence writing to `.Page.Scratch`, similarly to what you dide with localstorage.

Anyway, I think we should continue with Option 3, but it might be best to get the webpack stuff fixed in a separate PR. I believe that if we set `publPath` to `https://localhost:3001` during development, we can (hopefully) get rid of the `webpack_url.html` partials. I want to talk with @gumaerc about this a little today when he gets into the office.

Issue #1076, opened 2023-02-09
Course zip downloads are missing


ChristopherChudzicki:
### Steps to Reproduce

1. Visit https://ocw.mit.edu/courses/3-091-introduction-to-solid-state-chemistry-fall-2018/download/ 

### Expected Behavior

Should be able to download course zip file

### Actual Behavior

No zip file

PR #979, opened 2022-11-14
hide download button if zip doesn't exist v2

gumaerc:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-hugo-themes/issues/937

#### What's this PR do?
This PR revises the code in https://github.com/mitodl/ocw-hugo-themes/pull/960. Instead of manually constructing the `short_id` from info already in the course metadata, the `short_id` was explicitly added to JSON files published by `ocw-studio` (https://github.com/mitodl/ocw-studio/pull/1566) and that is being used instead. This was done because `short_id` is customizable upon site creation and the OCW team has changed their format, and may do so again in the future. This is the most foolproof way to ensure that the link to the ZIP archive of a given course is correct.

#### How should this be manually tested?
Testing this is a little bit tricky, since `hugo server` overrides any `baseUrl` setting to be prefixed with `//localhost:3000/`. We need to make a tiny edit for our test to work. Therefore, testing this particular functionality with `yarn start course` isn't going to work here. The best way to test this is by using a locally running instance of `ocw-studio`.

 - First of all, make sure you've taken care of the following prerequisites in `ocw-studio` and check the readme if you're unsure about any of them:
   - Custom Github test organization configured
   - Minio support configured
   - Concourse support configured
   - Recent database restore on your local DB
   - Set the following environment variables in your `.env`:
```
OCW_HUGO_PROJECTS_BRANCH=main
OCW_HUGO_THEMES_BRANCH=cg/download-testing
STATIC_API_BASE_URL=https://live.ocw.mit.edu
```
   - Make a small edit to `websites/views.py` at line 341 to limit the mass build to just the site we are testing (you can use any course in your database that uses `ocw-course-v2`): `sites = sites.prefetch_related("starter").order_by("name").filter(name="21l-011-the-film-experience-fall-2013")`
 - Spin up `ocw-studio` with `docker-compose up`
 - Run the following management commands, given a test course with the name `21l-011-the-film-experience-fall-2013`:
```
docker-compose exec web ./manage.py upsert_theme_assets_pipeline
docker-compose exec web ./manage.py backpopulate_pipelines --filter ocw-www
[...]

PR #1077, opened 2023-02-09
remove zip file existence check

gumaerc:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/1076

#### What's this PR do?
This PR removes some code that checks to make sure that a ZIP file exists at the expected URL before generating a link button pointing to it on the download page. This functionality was introduced in https://github.com/mitodl/ocw-hugo-themes/pull/960 and https://github.com/mitodl/ocw-hugo-themes/pull/979 as a way to prevent the link to the download button from being generated at build time if the file did not exist yet. The code is fragile, and relied on the existence of a bug in Hugo that has since been fixed - https://github.com/gohugoio/hugo/issues/10604. We no longer need this check, as since https://github.com/mitodl/ocw-studio/pull/1630 is part of the single course publishing pipeline now, the ZIP is guaranteed to be generated and put in place before the first online build of a course is run.

 - Run `yarn start course`
 - Visit http://localhost:3000 and click the download course button in the course info drawer
 - On the download page, you should see a button to download the course that links to a ZIP file
 - The link won't work in the local dev server, but it should look something like `https://localhost:3000/10.302-fall-2004.zip`


ChristopherChudzicki (approved):
👍 Re local dev url, mine is more like `https://live-qa.ocw.mit.edu//localhost:3000/9.40-spring-2018.zip` but I think that's fine for local and not a behavior change.

gumaerc:
> +1 Re local dev url, mine is more like `https://live-qa.ocw.mit.edu//localhost:3000/9.40-spring-2018.zip` but I think that's fine for local and not a behavior change.

Yea, that has to do with setting `RESOURCE_BASE_URL`, which is a local dev setting and isn't set in RC or production.

PR #1117, opened 2023-04-04
fix: tsconfig.json errors for Overwriting Input File

### LINKED ISSUE #1114: Overwriting Input File Errors in tsconfig.json
### Steps to Reproduce

Run: `npx tsc --build `
You'll see a list of all of those errors.

### Expected Behavior

There shouldn't be errors.

### Actual Behavior

There are errors reported by `tsc --build` (and also by VS Code)

### Screenshot or Screencast

These are all 11 errors:

image
image

I noticed these errors when I was working on videojs stuff (https://github.com/mitodl/ocw-hugo-themes/pull/1089) That's when they were reported by VS Code.
This is coming from the VS Code Typescript Extension.
If we don't have an output directory (`outDir`) in `compilerOptions` inside `tsconfig.json`, the compiled JavaScript files overwrites the input files. And that's where these errors are coming from.

ibrahimjaved12:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested

Closes #1114

#### What's this PR do?
This PR sets an output directory (`outDir`) in `compilerOptions` with exclude path so that the compiled Javascript files will not attempt to overwrite the input files which were previously placed right next to the input files with the same names.

1. Checkout to this branch
2. Run `npx tsc --build`
3. Observe that there are no errors reported.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -20,7 +20,8 @@
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "strict": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "outDir": "dist"
   },
   "include": [
     "./base-theme/assets/**/*.ts",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -36,7 +36,7 @@
   ],
   "exclude": [
     "./base-theme/assets/webpack",
-    "./dist/**/*"
+    "./dist/**/*.js"
   ],
   "ts-node": {
     "files": true,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -21,7 +21,7 @@
     "forceConsistentCasingInFileNames": true,
     "strict": true,
     "skipLibCheck": true,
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -34,7 +35,8 @@
     "./www/assets/**/*.js",
   ],
   "exclude": [
-    "./base-theme/assets/webpack"
+    "./base-theme/assets/webpack",
+    "./dist/**/*"
```
fakhar-ud-din, line 39: Since we only want to exclude js files, should we not add .js file. extension here?

Not sure, perhaps, you were also doing something else here. Otherwise looks good to me!

ibrahimjaved12: I think you're right. Since in the dist directory, `.js.map` files are also generated, and all of our errors are only coming from `.js` files, there's no reason to exclude other files, such as `.js.map`

fakhar-ud-din (approved):
LGTM 👍

Issue #1137, opened 2023-05-02
video expandable tabs other than related resources not rendering content


gumaerc:
### Steps to Reproduce

Build a course such as https://github.mit.edu/ocw-content-rc/15.s21-january-iap-2014 that has videos that utilize the video expandable tab for content other than "Related Resources."

### Expected Behavior

The content is rendered

### Actual Behavior

The content is not rendered

PR #1123, opened 2023-04-10
Update Video Partial to Handle Resource Link Shortcodes

pt2302:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested
  - [X] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1116.

#### What's this PR do?
Updates the video partial to handle resource link shortcodes.

#### How should this be manually tested?
To test this across all of OCW, there actually needs to be an additional data update component (https://github.com/mitodl/ocw-studio/issues/1740). The testing steps described here simulate this update, but users would not actually be editing the raw markdown to make this change normally. The main change in the resource markdown is replacing a hard-coded relative link with a resource link shortcode.

Run `yarn start course res.18-005-spring-2010` to fetch the repository locally. Stop the server, and replace the YAML portion of `res.18-005-spring-2010/content/resources/big-picture-of-calculus.md` with
```
content_type: resource
description: ""
end_time: ""
file: null
learning_resource_types: []
license: https://creativecommons.org/licenses/by-nc-sa/4.0/
ocw_type: ""
optional_tab_title: ""
optional_text: ""
parent_title: Highlights of Calculus (5 videos)
parent_type: SupplementalResourceSection
related_resources_text: >-
  Practice problems  

  {{% resource_link "49176758-ccc5-5fb8-7ef0-97fda2bca7ed" "(PDF)" %}}


  Audio summary for practice problems  

  ([MP3 - 1.6MB](http://www.archive.org/download/MITRES18.005/MITRES18_005S10_BigPictureCalculus_Summary_32K.mp3))
resource_index_text: ""
resourcetype: Video
start_time: ""
title: Big Picture of Calculus
uid: afb08409-c408-09e3-672c-8d9dd62a9b9f
video_files:
  archive_url: https://archive.org/download/MITRES18.005/MITRES18_005S10_BigPictureCalculus_300k.mp4
  video_captions_file: /courses/res-18-005-highlights-of-calculus-spring-2010/1ce9dc37be9557d5a3316caec220811e_UcWsDwg1XwM.vtt
  video_thumbnail_file: https://img.youtube.com/vi/UcWsDwg1XwM/default.jpg
  video_transcript_file: /courses/res-18-005-highlights-of-calculus-spring-2010/424a7610fccc797aaa2455d392246057_UcWsDwg1XwM.pdf
video_metadata:
  youtube_id: UcWsDwg1XwM
```

[...]

PR #1135, opened 2023-04-28
feat: keyboard navigable video buttons

ibrahimjaved12:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [X] Testing
  - [X] Code is tested
  - [X] Changes have been manually tested

#### What are the relevant tickets?
Closes #1126 

#### What's this PR do?
This PR: 
- Makes video buttons (such as transcript, related resources, download icons, etc) keyboard navigable
- Introduces playwright tests for the keyboard-navigable functionality.

#### How should this be manually tested?
**Testing functionality:**
During these steps, make sure the mobile responsiveness for the video page/video player also works as intended.
1. Checkout to this branch
2. `yarn start course res.18-009-fall-2015`
3. For embed videos:
    - Go to this link: http://localhost:3000/pages/video-series-overview/
    - Navigate by using `'tab'` key and go to video page by using tabs and then enter at `"VIEW VIDEO PAGE"` tab bar. 
4. Stop the local server and use this course now: `yarn start course res.18-009-fall-2015`
5. Go to any video page: http://localhost:3000/resources/part-1-introduction-and-overview-of-business-plans/
6. Using `tabs` in combination of `enter` or `spacebar`, make sure you can:
    - Collapse/expand the tab bars (for Transcript / Lecture Notes)
    - For Lecture Notes, make sure you can navigate using keyboard and open the link inside text: `Introduction to Business Plans (PDF)`
    - Make sure you can navigate to the download video icon and the links at "Download video" "Download transcript" are keyboard navigable and clickable.
 
**Running playwright tests:**

1. Checkout to this branch.
2. Run each playwright test about 200 times to ensure there is no flakiness:
`yarn test:e2e video-tabs.spec.ts --repeat-each 200
`

### Reviewing Playwright tests:
For reviewing playwright tests of this PR, this discussion might be helpful as it also relates to this PR:
https://github.com/mitodl/hq/discussions/1391

PR #1138, opened 2023-05-02
pass page context to video_expandable_tab wherever it's used

gumaerc:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/1137

#### What's this PR do?
In https://github.com/mitodl/ocw-hugo-themes/pull/1123, we updated the `video_expandable_tab.html` partial to use `.RenderString` as opposed to `markdownify`. Since `.RenderString` is a property of the `Page` context, the page context has to be passed into the partial for it to work. This PR passes that context on the remaining references to `video_expandable_tab.html` that were not handeld in https://github.com/mitodl/ocw-hugo-themes/pull/1123. A new playwright test has also been added to ensure that the expected tab titles and contents are rendered where expected. Playwright was upgraded to a new version as part of this to take advantage of the `Locator.all()` function.

 - Download https://github.mit.edu/ocw-content-rc/15.s21-january-iap-2014 and set your `COURSE_CONTENT_PATH` and `OCW_TEST_COURSE` env vars to point to it
 - Start the course with `yarn start course`
 - Ensure that at http://localhost:3000/resources/part-2-legal-issues/, a tab with the title "Lecture Notes" is rendered, and inside of it is a link to a PDF resource. This link will not work locally, because it's a hard coded markdown link with the `/courses` prefix, but it will work in production. Converting links in the `optional_text` field to resource link shortcodes in `ocw-studio` is a separate issue.


ChristopherChudzicki (changes requested):
**Requested Change:** See comment on test. IMO, it's worthwhile to do it with `getByRole`.

On the other hand, there are other issues with how we set up tabs. (E.g., we specify `aria-controls` on the tab, but we give it a classname. The value of `aria-controls` is supposed to be the ID of the tabpanel.) And addressing them all is probably out-of-scope. But I think adding the right roles wouldn't hurt. 

BTW: Re tab / tabpanel roles, a nice example of this is Material UI's Tab demo: https://mui.com/material-ui/react-tabs/ Obviously we aren't using react in ocw (at least outside of search). But I like to look at their HTML output in dev console sometimes to see how the roles (and aria stuff) is set up. 

@ibrahimjaved12 Heads up on this PR, since it's potentially related to #1135

test-sites/ocw-ci-test-course/content/resources/ocw_test_course_mit8_01f16_l01v01_360p.md
```diff
@@ -11,8 +11,8 @@ image_metadata:
 learning_resource_types: []
 license: https://creativecommons.org/licenses/by-nc-sa/4.0/
 ocw_type: ""
-optional_tab_title: ""
-optional_text: ""
+optional_tab_title: "Optional Tab"
+optional_text: "[OCW](https://ocw.mit.edu)"
 parent_title: Test Course Video
```
ChristopherChudzicki, line 16: @gumaerc Should this video actually be showing a transcript tab?

The transcript is `""`. I thought hugo treated that like `null` ?

gumaerc: This is a good point. On line 26 of `video.html`, we have:

`{{ if $captionsLocation }}`

First of all, this is only checking the captions location. I assume this is because the "Transcript" tab actually uses the captions file with VideoJS to render the interactive transcript, then the PDF is offered with a download button. Furthermore, both of the raw values in `.Params` were being passed through `resource_url.html` before checking them, which has the effect of prefixing them with `RESOURCE_BASE_URL` when testing locally, which makes them not a blank string. My updated logic checks the raw params to form a boolean about whether or not the expandable tab should be rendered and adds a test to assert that it works the way it should.

gumaerc:
@ChristopherChudzicki All set for another look

tests-e2e/util/VideoPage.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export class VideoPage {
+  readonly page: Page
+  readonly coursePage: CoursePage
+  readonly tabs: Locator
+  readonly tabPanels: Locator
+
+  constructor(page: Page) {
+    this.page = page
+    this.coursePage = new CoursePage(page, "course")
```
ChristopherChudzicki, line 12: **Suggestion** IMO, there's enough overlap between a "Video Page" and a "Video Embed", that it makes sense to have a single model like "Video Section" _which I realize isn't a great name._ Then in a test you'd use both:
```ts
  const course = new CoursePage(page, "course")
  await course.goto("/resources/ocw_test_course_mit8_01f16_l01v01_360p")
  const video = new VideoSection(page)
```

**Requested Change:** I do think it is worthwhile to...
1. Scope the tab selectors to the video area. *scoping is good, and It's not out of the question that we add "tab"/"tabpanel" to the course info tab.*
2. Make `tabs` and `tabpanels` methods that optionally accept options. *I also think singular naming is more in-line with how playwright does things

My recommendation would be:
```ts 

type ByRoleOptions = Parameters<Page["getByRole"]>[1]
class VideoSection {
  /**
   * Scope locators within the video area.
   */
  readonly within: Locator

  constructor(page: Page) {
    this.within = page.locator(".video-page")
  }


  tab(opts: ByRoleOptions): Locator {
    return this.within.getByRole("tab", opts)
  }

  tabpanel(opts: ByRoleOptions): Locator {
    return this.within.getByRole("tabpanel", opts)
  }
}
```

tests-e2e/util/WwwPage.ts
```diff
@@ -28,13 +28,16 @@ class WwwPage {
...
-  getCourseCard(title: string, { expectedCount = 1 } = {}): Locator {
+  async getCourseCard(
+    title: string,
+    { expectedCount = 1 } = {}
+  ): Promise<Locator> {
     const cardXPath = `div[${xPath.predicates.hasClass("course-card")}]`
-    const courseCard = this.page
+    const courseCard = await this.page
       .getByRole("link", { name: title })
       .locator(`${closest(cardXPath)} >> visible=true`)
```
ChristopherChudzicki, line 38: I'm fairly certain that this `await` does nothing: locators are evaluated lazily...there's nothing to really wait for until you use the locator for something.

The await below makes a lot of sense.

tests-e2e/ocw-ci-test-course/video-tabs.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    /\s*$/, // first tab is just download button with no title.
+    /Related Resources/,
+    /Optional Tab/
+  ])
+
+  const panels = await videoPage.tabPanels.all()
+  const html = await Promise.all(panels.map(panel => panel.innerHTML()))
+
+  expect(html).toHaveLength(3)
+
```
ChristopherChudzicki, line 20: Think it's worth adding `const _emptyPanel = html[0]` ? or a comment.

ChristopherChudzicki (changes requested):

ChristopherChudzicki:
@gumaerc Functionality looks good I did suggest a change to VideoPage that I think will make it easier for @ibrahimjaved12 and @fakhar-ud-din .

> - Download https://github.mit.edu/ocw-content-rc/15.s21-january-iap-2014 and set your COURSE_CONTENT_PATH and OCW_TEST_COURSE env vars to point to it
> - Start the course with yarn start course

pppppshaw, `yarn start course 15.s21-january-iap-2014` is what you mean 😄

gumaerc:
@ChristopherChudzicki Thanks, ready for another look.

ChristopherChudzicki (approved):

ChristopherChudzicki (approved):

Issue #1133, opened 2023-04-21
offline video playback


gumaerc:
As a user of an OCW mirror drive or otherwise offline site, I'd like to be able to play back locally stored archive videos while browsing said offline site

#### Acceptance Criteria:

- [x] Embedded videos and video pages load the offline version of the video in offline themes
- [x] If the offline version of the video isn't available, try and load the video from YouTube instead

#### Out of Scope:
- Transcript and captions are also loaded from local resources in the offline themes

gumaerc:
Progress update:

  - I have videos playing back locally in VideoJS locally
  - Since the mirror drive and the offline ZIP downloads both use the same theme, `course-offline`, if this change is made there then the sites in the ZIP downloads are going to expect the videos to be included. With this in mind and also the fact that there are videos missing from the original OCW mirror, I'm exploring the possibility of checking for the existence of the archive video during the offline build. If the video is not found, the YouTube player would be rendered instead
  - Regardless, the archive videos will need to be pulled into the Concourse worker at build time and I'm worried that this will significantly increase publish time because of the size of videos. I'm working on testing locally to get an idea about how much impact copying in the videos might have on publishing time.

PR #1160, opened 2023-05-23
offline video playback

gumaerc:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/ocw-hugo-themes/issues/1133

#### What's this PR do?
This PR changes the behavior video pages and embeds in the offline themes. A new partial, `video_player.html` was added. Places that called `youtube_player.html` directly have been altered to call this partial instead. In the online theme, `youtube_player.html` is simply called without any other action. In the offline theme, a check is made against the `static_resources` leaf node to see if a copy of the archive video is present at build time. If the video exists, `local_video_player.html` is called, which renders a VideoJS player pointing at the locally stored archive video. If the video is not present, `youtube_player.html` is called instead to load the video from YouTube.

 - Download the course content for [18-06-fall-2010](https://github.mit.edu/ocw-content-rc/18.06-spring-2010)
 - Make sure you have [`ocw-hugo-projects`](https://github.com/mitodl/ocw-hugo-projects) cloned locally
 - Source `ocw-studio` production AWS credentials and set them up in a terminal
 - In the same terminal change directory to the course content repo you downloaded above
 - Run:
```
mkdir -p content/static_resources
aws s3 sync s3://ol-ocw-studio-app-production/courses/18-06-linear-algebra-spring-2010/ ./content/static_resources/
```
 - Once the static resources have been downloaded, run the following in the `ocw-hugo-themes` directory, being sure to use *absolute paths*: `yarn build /path/to/18.06-spring-2010/ /path/to/ocw-hugo-projects/ocw-course-v2/config-offline.yaml`
 - In your file browser, go to your course content repo and enter the `dist` folder, then double click on `index.html` to open the offline site in a web browser
 - Go to the Instructor Insights page and click on the embedded video there and ensure it plays back and that the VideoJS controls work properly and that the video is sized properly in the layout
 - Go to the Video Lectures section and ensure the same thing for any of the video lectures contained within
 - Go back to the course content repo, enter the `content/static_resources` folder and delete one of the archive videos, noting which one you deleted. Find the page that this video is referenced on so you know the URL to browse to when testing by searching the repo for the video filename.
 - Re-run the offline build
 - Visit the page for the video that is now missing and verify that the YouTube player is loaded instead.

This is part of the [mirror drive epic](https://github.com/mitodl/hq/issues/785) and enables playback of archive videos if they are included with a course at build time. This PR should be reviewed, but not merged until a separate situation is addressed. In `ocw-studio` where MIT's OCW content is authored and subsequently Concourse where the builds are actually run, the offline build is run and is both used to generate the individual ZIP downloads of courses as well as the content for the mirror drive, stored in separate S3 buckets. In our Concourse pipelines, mp4 files are currently filtered out of the ZIP downloads to reduce the file size of the downloads. Before this is merged we should alter the logic here to run a Hugo build with *and* without the archive videos. The version without the archive videos should have the videos filtered out before the Hugo build even runs, so that the theme does not assume they exist and renders the YouTube player instead on pages in that offline site that have video players. If this PR is merged without addressing this issue first, the offline sites in ZIP downloads will have offline video players rendered that will attempt to load local mp4 videos that don't exist.

This should not be merged until https://github.com/mitodl/ocw-studio/pull/1836 is released, as the way mp4 files are filtered out of builds in Concourse after Hugo runs means that local video players will be rendered in offline sites, and the videos will not be present in the ZIP file.

image


base-theme/layouts/partials/local_video_player.html
```diff
@@ -0,0 +1,28 @@
...
+    data-setup='{
+    "fluid": true,
+    "inactivityTimeout": 0'
+  >
+      {{ if .archiveUrl }}
+        <source src="{{ partial "resource_url.html" (dict "context" .context "url" .archiveUrl) }}" type="video/mp4" />
+      {{ end }}
+      {{ if .captionsLocation }}
+        <track kind="captions" src="{{ .captionsLocation }}" mode="disabled" srclang="en" label="English">
+      {{ end }}
```
ChristopherChudzicki, line 16: As far as I can tell, offline captions (w/o youtube) don't work at all right now, either in the interactive transcript or in the player itself.

Maybe better to just leave these tracks out? That way there's no (non-functional) "CC" button. 

Screenshot 2023-05-25 at 4 34 56 PM

ChristopherChudzicki (changes requested):
Looks good, just the curly brace request.

gumaerc:
@ChristopherChudzicki should be ready for another look

base-theme/layouts/partials/local_video_player.html
```diff
@@ -0,0 +1,34 @@
...
+  >
+      {{ if .archiveUrl }}
+        <source src="{{ partial "resource_url.html" (dict "context" .context "url" .archiveUrl) }}" type="video/mp4" />
+      {{ end }}
+  </video>
+</div>
+
+{{ if and .startTime .endTime }}
+<script>
+  window.onload = function() {
```
ChristopherChudzicki, line 20: The current implementation here doesn't work with the `or` suggestion above. Also, the player is a bit awkward if you press play AFTER `endTime`. (it plays for a second, then pauses). Maybe that's reasonable behavior, though.


1. use `const`
2. conditional on start time / end time
3. **MAYBE, MAYBE NOT**  remove "stopper" after its first application. *The youtube player handles this very nice: The video stops, the play button goes away. You can force the youtube player to continue by scrubbing if you really want.*

Re (3) ... The current behavior, 

```js
  window.onload = function() {
    if (window.videojs) {
      const player = window.videojs("{{ $playerId }}")
      const startTime = {{ .startTime }}
      const endTime = {{ .endTime }}
      if (startTime) {
        player.currentTime(startTime)
      }
      const stopper = () => {
        if (player.currentTime() >= endTime) {
          player.pause()
          player.off("timeupdate", stopper) // Maybe???
        }
      }
      if (endTime) {
        player.on("timeupdate", stopper)
      }
    }
  }
```

ChristopherChudzicki (changes requested):
One requested change around startTime / endTime stuff. Right now video doesn't work if startTime exists but not endTime (which we allow for youtube player).

Also, reminder from our slack conversation:
- With current publish setup, HTML files will be produced as if videos existed, but then the videos will be filtered out of the zip => broken videos.

base-offline/layouts/partials/video_player.html
```diff
@@ -0,0 +1,14 @@
+{{ $archiveUrl := .archiveUrl }}
+{{ $archiveVideoFound := false }}
+{{ $staticResourcesBundle := site.GetPage "/static_resources" }}
+{{ $archiveVideos := $staticResourcesBundle.Resources.ByType "video" }}
+{{ range $archiveVideos }}
+  {{ if strings.Contains $archiveUrl .Name }}
+    {{ $archiveVideoFound = true }}
+  {{ end }}
+{{ end }}
```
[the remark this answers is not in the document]
gumaerc: Good catch, I didn't see this in the docs. I did get this to work but had to modify it a little. The code as you suggested it does not work, because `.Match` is looking at the `.Name` property of each resource, so what you pass in needs to be what you are matching against `.Name`. Since the `.Name` property of resources is the filename, I had to parse the url and then the path:

```
{{ $archiveVideoFound := false }}
{{ $archiveUrl := .archiveUrl }}
{{ $parsedArchiveUrl := urls.Parse $archiveUrl }}
{{ $parsedArchiveUrlPath := path.Split $parsedArchiveUrl.Path }}
{{ $archiveFile := $parsedArchiveUrlPath.File }}
{{ $staticResourcesBundle := site.GetPage "/static_resources" }}
{{ $archiveVideos := $staticResourcesBundle.Resources.ByType "video" }}
{{ $archiveVideoSearchResuts := $staticResourcesBundle.Resources.Match $archiveFile }}
{{ $archiveVideoFound = gt (len $archiveVideoSearchResuts) 0 }}
```

base-theme/layouts/partials/resource_url.html
```diff
@@ -9,8 +9,10 @@
 */}}
 {{- $prefix := getenv "RESOURCE_BASE_URL" | default "" -}}
 {{- $url := printf "%s%s" (strings.TrimSuffix "/" $prefix) (urls.Parse $url).Path -}}
-{{- if hasPrefix $url "http" -}}
-  {{- return $url -}}
-{{- else -}}
+{{- if not (hasPrefix $url "http") -}}
   {{- partial "site_root_url.html" $url -}}
```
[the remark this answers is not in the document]
gumaerc: You are correct, I'm not sure what I was thinking here. To be honest, I'm not sure we need to reference `site_root_url` here. This partial was created as a way to bypass the `baseURL` setting and reference the root of the site directly, for links in the header that need to do that such as `/about`, etc. If you look at that partial, you'll see it uses a bunch of `printf` statements to print URLs directly, and this is because it was written near the start of the project and Hugo did not support calling `return` in partials back then. We should revisit how this partial works, or if it is even needed but that is work for another PR.

gumaerc:
@ChristopherChudzicki all set for another look

base-offline/layouts/partials/video.html
```diff
@@ -0,0 +1,26 @@
+{{ $youtubeKey := trim .Params.video_metadata.youtube_id " "}}
```
ChristopherChudzicki, line 1: **Suggestion / Reminder:** `{{- -}}` to cut down on whitespace.

I don't really think it's worth changing or worrying about, since we don't have a way to enforce it and it is only occasionally annoying.

ChristopherChudzicki (approved):
LGTM, looking at the studio PR now.

PR #1149, opened 2023-05-15
fix: non-pdf files show pdf thumbnail

HussainTaj-arbisoft:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [x] Mobile width screenshots
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/1338

#### What's this PR do?

In Studio, all files (txt, c, py, tex, pptx, docx, etc) are `Document`s. To display PDF icon for only PDFs, we need to check the document's file type.

#### How should this be manually tested?
1. Download the course [ht-test-site](https://github.mit.edu/ocw-content-rc/ht-test-site) and place it in your local ocw-content folder. (You might not need to do it manually if you have MIT GitHub and RC content configured)
2. Navigate to your ocw-hugo-theme clone.
3. Run `yarn start ht-site-site`. (This name should match the folder name you have in your content folder)
4. Open [http://localhost:3000/lists/a-resource-list/](http://localhost:3000/lists/a-resource-list/)
5. **Expected Behavior**: (Refer to the screenshots)
    - PDF files have a PDF icon
    - Videos have a video icon
    - All other files have a generic file icon
6. (Optional) If you're interested in comparing this output to what we had previously, you may explore [this page](https://draft-qa.ocw.mit.edu/courses/1-ht-test-site-spring-2023/lists/a-resource-list/).

#### Screenshots (if appropriate)

Screenshot 2023-05-15 at 5 39 40 PM
Screenshot 2023-05-15 at 5 39 19 PM

PR #1163, opened 2023-05-26
fix video download popup for multiple instances of videos

ibrahimjaved12:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/hq/issues/1518

#### What's this PR do?
This video fixes video download popups for multiple instances of videos (sometimes we have multiple embed videos in a single page). Previously the download button was working only for first video if there are multiple in a page.

The issue was reported for the courses:
3.091-fall-2018
cms.611j-fall-2014
20.219-january-iap-2015

For `3.091-fall-2018`:
- Instructor Insights, Goodie Bag Tutorials and Problems

For `cms.611j-fall-2014`:
- on all pages under Student Games Projects

For `20.219-january-iap-2015`:
- on all pages starting with "Day X" and the Student projects

1. Checkout to this branch
2. `yarn start course X` (where X is the above course IDs)
3. Browse over to the mentioned pages above, and where there are multiple embed videos, make sure all of them have working download buttons. Although they might be too many (and there maybe other courses also), just check for few of them on each course and that should be good to go.
4. Make sure it's also working for single video pages.

After it, run playwright tests:
use `test.only` for the test`test("Verify that the Download Button works for multiple embed videos in a page")` or simply run this command to run all video-tab tests once:
`yarn test:e2e video-tabs.spec.ts`

### DIFF
--- a/course-v2/assets/js/video_download_popup.js
+++ b/course-v2/assets/js/video_download_popup.js
@@ -1,16 +1,22 @@
 export const initVideoDownloadPopup = () => {
-  const downloadIcon = document.querySelector(".video-download-icons")
-  const popup = document.querySelector(".video-tab-download-popup")
-  if (downloadIcon) {
-    downloadIcon.addEventListener("click", event => {
-      event.stopPropagation()
-      if (popup) popup.classList.toggle("hidden")
+  const downloadIcons = document.querySelectorAll(".video-download-icons")
+  const popups = document.querySelectorAll(".video-tab-download-popup")
+
+  if (downloadIcons.length > 0) {
+    downloadIcons.forEach((downloadIcon, index) => {
+      const popup = popups[index]
+      downloadIcon.addEventListener("click", event => {
+        event.stopPropagation()
+        if (popup) popup.classList.toggle("hidden")
+      })
     })
   }
 
   document.addEventListener("click", () => {
-    if (popup) {
-      popup.classList.add("hidden")
-    }
+    popups.forEach(popup => {
+      if (popup) {
+        popup.classList.add("hidden")
+      }
+    })
   })
 }
--- a/course-v2/assets/js/video_download_popup.ts
+++ b/course-v2/assets/js/video_download_popup.ts
@@ -8,21 +8,19 @@ export const initVideoDownloadPopup = () => {
       const popup = popups[index] as HTMLElement
       downloadIcon.addEventListener("click", event => {
         event.stopPropagation()
-        if (popup) {
-          // Clicked on the same download button, toggle the popup
-          if (popup === activePopup) {
-            popup.classList.toggle("hidden")
-            activePopup = null
+        // Clicked on the same download button, toggle the popup
+        if (popup === activePopup) {
+          popup.classList.toggle("hidden")
+          activePopup = null
+        }
+        // Clicked on a different download button, close previous popup (if any) and toggle open new popup
+        else {
+          if (activePopup) {
+            activePopup.classList.add("hidden")
           }
-          // Clicked on a different download button, close previous popup (if any) and toggle open new popup
-          else {
-            if (activePopup) {
-              activePopup.classList.add("hidden")
-            }
-            // Show the clicked popup
-            popup.classList.remove("hidden")
-            activePopup = popup
-          }
+          // Show the clicked popup
+          popup.classList.remove("hidden")
+          activePopup = popup
         }
       })
     })
--- a/course-v2/assets/js/video_download_popup.ts
+++ b/course-v2/assets/js/video_download_popup.ts
@@ -3,27 +3,25 @@ export const initVideoDownloadPopup = () => {
   const popups = document.querySelectorAll(".video-tab-download-popup")
   let activePopup: HTMLElement | null = null
 
-  if (downloadIcons.length > 0) {
-    downloadIcons.forEach((downloadIcon, index) => {
-      const popup = popups[index] as HTMLElement
-      downloadIcon.addEventListener("click", event => {
-        event.stopPropagation()
-        if (popup === activePopup) {
-          // Clicked on the same download button, toggle the popup
-          popup.classList.toggle("hidden")
-          activePopup = null
-        } else {
-          // Clicked on a different download button, close previous popup (if any) and toggle open new popup
-          if (activePopup) {
-            activePopup.classList.add("hidden")
-          }
-          // Show the clicked popup
-          popup.classList.remove("hidden")
-          activePopup = popup
+  downloadIcons.forEach((downloadIcon, index) => {
+    const popup = popups[index] as HTMLElement
+    downloadIcon.addEventListener("click", event => {
+      event.stopPropagation()
+      if (popup === activePopup) {
+        // Clicked on the same download button, toggle the popup
+        popup.classList.toggle("hidden")
+        activePopup = null
+      } else {
+        // Clicked on a different download button, close previous popup (if any) and toggle open new popup
+        if (activePopup) {
+          activePopup.classList.add("hidden")
         }
-      })
+        // Show the clicked popup
+        popup.classList.remove("hidden")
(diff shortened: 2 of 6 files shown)


tests-e2e/util/VideoElement.ts
```diff
@@ -5,7 +5,7 @@ export class VideoElement {
   readonly container: Locator
 
   constructor(page: Page) {
-    this.container = page.locator(".video-page")
+    this.container = page.locator(".video-embed, .video-page")
   }
 
   tab(opts: ByRoleOptions): Locator {
```
ChristopherChudzicki, line 11: The reason I suggested adding an `nth` method on VideoElement is that I thought it might be confusing to call `nth()` on the "final" locators (I.e., `tab().nth()`, `downloadButton().nth()`, ...).

E.g., imagine multiple video elements on a page, each with multiple tabs. Then `videoElement.tab()` would return something like 
```
video1_tab1, video1_tab2, video1_tab3, video2_tab1, video2_tab2, ...
```
So scoping the `nth` to the container seemed like a good idea. I.e., "Select tabs for the nth video" not "select the nth tab for all the videos".

**In practice this may not be a problem** since we only get multiple videos with embedded videos, and they only have a single tab anyway.

course-v2/assets/js/video_download_popup.js
```diff
@@ -1,16 +1,22 @@
...
+      })
     })
   }
 
   document.addEventListener("click", () => {
-    if (popup) {
-      popup.classList.add("hidden")
-    }
+    popups.forEach(popup => {
+      if (popup) {
```
ChristopherChudzicki, line 17: It's safe to skip the `if(popup)` here: We needed it before because `document.querySelector` can return null. But `querySelectorAll` will always return a list.

course-v2/assets/js/video_download_popup.js
```diff
@@ -1,16 +1,22 @@
...
-      if (popup) popup.classList.toggle("hidden")
+  const downloadIcons = document.querySelectorAll(".video-download-icons")
+  const popups = document.querySelectorAll(".video-tab-download-popup")
+
+  if (downloadIcons.length > 0) {
+    downloadIcons.forEach((downloadIcon, index) => {
+      const popup = popups[index]
+      downloadIcon.addEventListener("click", event => {
+        event.stopPropagation()
+        if (popup) popup.classList.toggle("hidden")
```
ChristopherChudzicki, line 10: In contrast to my comment below, you do need the `if (popup)` here.

**BTW:** could you convert this file to typescript?

course-v2/assets/js/video_download_popup.ts
```diff
@@ -0,0 +1,34 @@
+export const initVideoDownloadPopup = () => {
+  const downloadIcons = document.querySelectorAll(".video-download-icons")
+  const popups = document.querySelectorAll(".video-tab-download-popup")
+  let activePopup: HTMLElement | null = null
+
+  if (downloadIcons.length > 0) {
```
ChristopherChudzicki, line 6: **Requested Change:** I think you should remove this check: There's no need to check a list's length is nonzero when iterating. Iterating over a zero-length list is totally fine.

_Checking that the list is actually defined, and not `null` or `undefined` might be necessary sometimes. But querySelectorAll **always** returns a list so there's no need for that here_

course-v2/assets/js/video_download_popup.ts
```diff
@@ -0,0 +1,34 @@
...
+    downloadIcons.forEach((downloadIcon, index) => {
+      const popup = popups[index] as HTMLElement
+      downloadIcon.addEventListener("click", event => {
+        event.stopPropagation()
```
ChristopherChudzicki, line 10: **Suggestion:** You can avoid keeping track of `activeElement` by restructuring this if you want.

Currently:
- (A) every download button has its own event listener
- (B) the overall document has an event listener

usually, (B) is responsible for closing the popup on "clicks outside" of the open popup. Except when the "click outside" ahppens to be another download button, then (A) is responsible.

A common way to deal with "click outside" behavior is to dynamically attach a closing listener on the opening click:
```ts
  downloadIcons.forEach((downloadIcon, index) => {
    const popup = popups[index]
    downloadIcon.addEventListener("click", event => {
      popup.classList.remove("hidden")
      const remove = (e: Event) => {
        if (event === e) return         // do not close popup when the original event bubbles up
        popup.classList.add("hidden")
        document.removeEventListener("click", remove) // cleanup after closing
      }
      document.addEventListener("click", remove)
    })
  })
```
Note that in this scenario you need to remove the "closing listener" once it has done its job.

Up to you.

tests-e2e/ocw-ci-test-course/video-tabs.spec.ts
```diff
@@ -2,6 +2,26 @@ import { test, expect } from "@playwright/test"
 import { CoursePage } from "../util"
 import { VideoElement } from "../util/VideoElement"
 
+test("Verify that the Download Button works for multiple embed videos in a page", async ({
```
ChristopherChudzicki, line 5:
```suggestion
test("that the Download Button works for multiple embed videos in a page", async ({
```
"verify" just means "test" anyway.

tests-e2e/ocw-ci-test-course/video-tabs.spec.ts
```diff
@@ -2,6 +2,26 @@ import { test, expect } from "@playwright/test"
...
+  page
+}) => {
+  const coursePage = new CoursePage(page, "course")
+  await coursePage.goto("pages/multiple-videos-series-overview/")
+  const videoElementsCount = await new VideoElement(page).count()
+  for (let i = 0; i < videoElementsCount; i++) {
```
ChristopherChudzicki, line 11: **Requested change:** Either add an assertion here:

```
expect(videoElementsCount).toBe(3)
```

or loop over a list with a definite length.

See my comment here: https://github.com/mitodl/ocw-hugo-themes/pull/1149#discussion_r1202469710

In general, I like to avoid iterating over dynamic lists in tests. If you need to, make an assertion to ensure that the list has the length you expect (and in particular that it does not have zero length).

ChristopherChudzicki (changes requested):
Looking good. Two small requestsand also a suggestion.

tests-e2e/ocw-ci-test-course/video-tabs.spec.ts
ibrahimjaved12: Done

tests-e2e/ocw-ci-test-course/video-tabs.spec.ts
ibrahimjaved12: Done

course-v2/assets/js/video_download_popup.ts
ibrahimjaved12: Done

course-v2/assets/js/video_download_popup.ts
ibrahimjaved12: Sticking to the previous (current) code. Also, for sake of completeness, going to mention the complete modified code here:
```
export const initVideoDownloadPopup = () => {
  const downloadIcons = document.querySelectorAll(".video-download-icons")
  const popups = document.querySelectorAll(".video-tab-download-popup")

  downloadIcons.forEach((downloadIcon, index) => {
    const popup = popups[index]
    downloadIcon.addEventListener("click", event => {
      if (!popup.classList.contains("hidden")) return // Abort if popup is already open
      popup.classList.remove("hidden")
      const remove = (e: Event) => {
        if (event === e) return         // do not close popup when the original event bubbles up
        popup.classList.add("hidden")
        document.removeEventListener("click", remove) // cleanup after closing
      }
      document.addEventListener("click", remove)
    })
  })
}
```

ChristopherChudzicki (approved):

Issue #1174, opened 2023-06-21
offline search for mirror drives


gumaerc:
As noted in https://github.com/mitodl/hq/discussions/375, OCW search cannot work when the user is offline, since it depends on the search API provided by open.mit.edu 

We need to replace the online search with an offline search that provides minimal functionality:
- [ ] Use the data provided by `ocw-www`'s content added in https://github.com/mitodl/ocw-studio/issues/1698 to iterate all websites known to the root website (`ocw-www` in this case)
- [ ] Feed this data into an offline search library such as Fusejs or Lunr
- [ ] Provide a basic interface for searching this data and providing search results in the compact view
- [ ] Link to the search page from the home page

Issue #1075, opened 2023-02-09
JS/TS partials do not transpile properly


ChristopherChudzicki:
### Steps to Reproduce

1. Visit https://ocw.mit.edu/courses/3-091-introduction-to-solid-state-chemistry-fall-2018/ or any course in IE11, e.g., via https://live.browserstack.com/dashboard (Or any browser that does not support ES6 arrow functions).
2. View consoles

### Expected Behavior
No syntax errors

### Actual Behavior
Syntax errors

ChristopherChudzicki:
We include TS source code as JS in HTML via Hugo's [`js.Build` pipe](https://gohugo.io/hugo-pipes/js/). But it appears that pipe is not respecting our TSConfig. It would be nice to figure that out.

At the very least, we could specify ES5 as the target environment (see https://gohugo.io/hugo-pipes/js/#options).

PR #1175, opened 2023-06-21
offline search for mirror drives

gumaerc:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1174

This PR adds the [Fuse.js](https://fusejs.io/) library to the project and utilizes it to provide offline search functionality in the `www-offline` theme, used as the home page of mirror drives. The search icon is now displayed in the header on pages that aren't the search page, linking to a new `search.html` page at the root. In https://github.com/mitodl/ocw-studio/pull/1705, we set up `ocw-studio` to publish a file under `content/websites` for each website that has been published live at least once inside the root website, `ocw-www`. This content is currently being used to generate a paginated list of sites at the root when using the `www-offline` theme in `index.html`. In `search.html`, we put that same list into an array in Javascript and feed it into Fuse.js as a data source, indexing on `primary_course_number` and `title`. The search page mostly replicates a basic version of the online search without any facets or resource search, providing links to sites relative to the root of the mirror drive in the same way the paginated index does. Everything is done in raw Javascript right in a script tag on the page to avoid any issues with CORS when viewing the page without a web server offline.

image
image

# How can this be tested?
 - Clone `ocw-hugo-themes` on this branch as well as https://github.mit.edu/ocw-content-rc/ocw-www and `ocw-hugo-projects` on the `cg/www-offline-enable-section` branch
 - In `ocw-hugo-themes`, run `yarn install` to make sure you install the latest dependencies
 - In `ocw-hugo-themes`, run `yarn build /path/to/ocw-content-rc/ocw-www/ /path/to/ocw-hugo-projects/ocw-www/config-offline.yaml`
 - Browse to your `ocw-content-rc/ocw-www/dist` folder and double click on `index.html`
 - Click the search icon in the header
 - Type any query you want here and search for courses
 - Click on any of the courses and verify that the path in your browser looks something like `file://path/to/ocw-content-rc/ocw-www/dist/courses/course-name`
 - These courses will not be present because this isn't a full mirror drive build, but if you want to test browsing to an offline site this way you can download the course ZIP from the live site, create a `courses` folder under `dist` and extract it there to a folder matching the course`s `name` property

# Release process
https://github.com/mitodl/ocw-hugo-projects/pull/255 should be merged and released before this PR

### DIFF
--- a/www-offline/layouts/search/section.html
+++ b/www-offline/layouts/search/section.html
@@ -46,116 +46,120 @@
   </div>
 </div>
 <script>
-  const searchResultTemplate = document.createElement("template")
-  searchResultTemplate.innerHTML = `
-<article>
-  <div class="card learning-resource-card compact-view">
-    <div class="card-contents">
-      <div class="lr-info search-result">
-        <div class="col-2 course-num">
-        </div>
-        <div class="course-title">
-          <a class="course-link"></a>
-        </div>
-        <div class="col-2 resource-level">
+window.onload = function() {
+  if (window.Fuse) {
+    // define the search-result web component and add it for use
+    const searchResultTemplate = document.createElement("template")
+    searchResultTemplate.innerHTML = `
+    <article>
+      <div class="card learning-resource-card compact-view">
+        <div class="card-contents">
+          <div class="lr-info search-result">
+            <div class="col-2 course-num">
+            </div>
+            <div class="course-title">
+              <a class="course-link"></a>
+            </div>
+            <div class="col-2 resource-level">
+            </div>
+          </div>
         </div>
       </div>
-    </div>
-  </div>
-</article>
-`
+    </article>
+    `
 
-  class SearchResult extends HTMLElement {
-    constructor() {
-      super()
+    class SearchResult extends HTMLElement {
+      constructor() {
+        super()
+      }
+
+      connectedCallback() {
+        this.innerHTML = searchResultTemplate.innerHTML
+        this.courseNumberDiv = this.querySelector(".course-num")
+        this.courseLink = this.querySelector(".course-link")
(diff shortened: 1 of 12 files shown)


www-offline/layouts/search/section.html
```diff
@@ -0,0 +1,165 @@
...
+                        <div class="w-100">
+                          <form role="search" class="search-box d-flex flex-direction-row">
+                            <input
+                              id="search-input"
+                              class="w-100 pl-2"
+                              type="text"
+                              aria-label="Search"
+                              placeholder="Search OpenCourseWare"
+                              />
+                            <button id="search-button" type="submit" class="py-2 px-3">Search</button>
```
ChristopherChudzicki, line 30:
```suggestion
                            <button id="search-button" type="button" class="py-2 px-3">Search</button>
```

Then you won't need to prevent default down below.

ChristopherChudzicki (commented):
This makes sense, works as described, and is super cool 👍 

Some things I noticed... I think that at least the first two are worth addressing:

1. I think we should change `minMatchCharLength: 5` to something significantly lower. Currently, OCW has no math courses (joking...):

Screenshot 2023-06-23 at 2 39 36 PM

2. When there are no matches, the page is very blank. I think we should say "No results found". *Or perhaps always show the count.*

3.  (Fixed) ~Sometimes the titles contain invalid characters on the search page. For example, `3.034-fall-2005` has an ampersand in its title:~


And, as mentioned in slack: We may want to move some/all of the JS to a file that gets processed by webpack. JS in template files do not. The arrow functions will make IE mad. See also #1075

gumaerc:
@ChristopherChudzicki all set for another look

ChristopherChudzicki (approved):

PR #1179, opened 2023-07-04
course-v2: replace window.onload with window.addEventListener for nanogallery2 and videojs

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/1749

It fixes the problem of video player not working when image gallery (nanogallery2) is in the same page for course-v2 theme.
The solution consists of using `window.addEventListener` instead of using `window.onload` function.

# How can this be tested?
See that nanogallery2 and videojs are working when together but also independently. See that embed videos/regular videos are both working.

Note: Before proceeding with each video/image gallery, make sure to clear cache.

To see that it works if we have image gallery and video both in same page, use course 16.885j-fall-2005
1. Checkout to this branch
2. `yarn start course 16.885j-fall-2005`
3. Go to related resources from left panel.
4. Check that video (embed video) is working (including the download button) and that the image gallery is also working.
5. Click "View video page" and check that video resource is working (regular video)

When you check these videos also ensure that regular videos are working (including the video download button) by clicking on "View video page" underneath the embed videos.
Another example of embed videos: `18.06sc-fall-2011` (instructor insights from left panel)
And another example of embed videos: `18.02sc-fall-2010` ([Click me for localhost video link](http://localhost:3000/pages/2.-partial-derivatives/part-a-functions-of-two-variables-tangent-approximation-and-optimization/session-27-approximation-formula/))
Another example of nanogallery2: `2.672-spring-2009` ([Click me for localhost image gallery link](http://localhost:3000/pages/labs/#nanogallery/fd614745-fc1f-6cb1-a422-ffc6befbbdd1_nanogallery2/0/1))

# Additional Context 
There is a critical thing. We cannot have another function `window.onload` in our codebase at least for a functionality/page that may also include either nanogallery2 or videojs dynamic imports because that will cause conflicts. We will need to continue using `event listener` instead of `window.onload` for this to work.


ChristopherChudzicki (changes requested):
This makes sense: use `addEventListener("load", ...)` so as to allow more than one listener.

However, I think this makes sense to do everywhere, and MDN seems to recommend it:
> Note: The addEventListener() method is the recommended way to register an event listener. The benefits are as follows:

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

So, my requested change is: replace all the instances of `window.onload`.

ibrahimjaved12:
FYI: Closed this PR because branch name was not precise as the scope of the PR had changed due to new discoveries of the original issue. And reopened it as mentioned in above link: https://github.com/mitodl/ocw-hugo-themes/pull/1185

Issue #1182, opened 2023-07-05
unexpected results for offline search 


pdpinch:
### Steps to Reproduce

1. Visit the offline search, for example, at http://ocw-content-offline-live-qa.s3.amazonaws.com/search.html
2. Search on "energy" 

### Expected Behavior

The search results should include any courses with "energy" in the title.

### Actual Behavior

The search results include courses like:
- 7.05 General Biochemistry
- 21H.306 The Emergence of Europe

PR #1183, opened 2023-07-06
adjust fuse threshold in offline search to better filter out irrelevant results

gumaerc:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1182

This PR simply adjusts the `threshold` value in the FuseJS configuration for offline search in `www-offline` to 0.2 from 0.4. After some testing, this seems to be a good place to set it in order to filter out irrelevant results.

image

# How can this be tested?
Follow the same instructions as https://github.com/mitodl/ocw-hugo-themes/pull/1175, but pay extra attention to the results and verify that irrelevant results have been minimized

### DIFF
--- a/www-offline/layouts/search/section.html
+++ b/www-offline/layouts/search/section.html
@@ -73,12 +73,8 @@ window.addEventListener('load', function() {
     const Fuse = window.Fuse
     const searchOptions = {
       shouldSort: true,
-      threshold: 0.2,
-      location: 0,
-      distance: 100,
-      matchAllTokens: true,
       includeScore: true,
-      maxPatternLength: 32,
+      threshold: 0.2,
       minMatchCharLength: 3,
       keys: ["title", "primary_course_number"]
     }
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki (changes requested):
In addition to the comment I left, one other requested change: Move `const searchResultsSection = document.getElementById("search-results")` above `performSearch`.

Currently, the `else` block throws an error since `searchResultsSection` is not defined. *Gosh, I forgot how easy these errors are in un-linted files.*

ChristopherChudzicki (changes requested):
@gumaerc Did you want to add `ignoreLocation: true`? Currently, fuse is using the default values for distance/loc, so "Geographic Information Systems" still shows limited results.

**Requested change:**  As mentioned:  Move const searchResultsSection = document.getElementById("search-results") above performSearch.

Currently, the `else` block is throwing when there are no results.

gumaerc:
@ChristopherChudzicki Whoops, I missed that part but that makes sense. It should be all set now.

ChristopherChudzicki (approved):

PR #1184, opened 2023-07-07
Change OCW News mobile(xs) display to horizontal

ibrahimjaved12:
https://github.com/mitodl/hq/issues/1388

This is a follow up PR to the original issue for some design modification.
Such that currently for small screens(xs), the news images stay a fixed height and tend to go vertical. But we want them to follow the same fashion for small screens as it was originally, i.e, horizontal. This is to help stay consistent with the image sizes and cropping work.

# Screenshots (In progress):
image
image
image
image
image

# How can this be tested?
The overall functionality/layout/design structure has already been tested. For this PR we just want to make sure for smaller screens the view looks correct. Make sure that the changes applied (changed related to numerical values) are actually visible on the page (that they are not overridden by some other properties) and nothing breaks.

For mobile interface (on all sizes where we have one news on display) make sure that no kind of images break the design. I tried testing with 100x100 images and it was ok, you're free to test anything you have in mind too.
You can use the netlify link/local host server and inspect element and add any image link from google in one of the already existing `img` tags to do this.

### DIFF
--- a/base-theme/assets/css/variables.scss
+++ b/base-theme/assets/css/variables.scss
@@ -46,10 +46,13 @@ $panel-course-info-text-color: #464646;
 // OCW news in homepage
 $featured-news-heading-font-lg: 1.875rem;
 $featured-news-line-height-lg: 2.325rem;
-$featured-news-heading-font-sm: 1.25rem;
-$featured-news-line-height-sm: 1.56rem;
 $featured-news-heading-font-md: 1.4rem;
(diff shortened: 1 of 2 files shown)


pdpinch:
This is definitely better. I discussed some further improvements with @mbilalmughal but this is good enough to merge.

HussainTaj-arbisoft (approved):
LGTM 👍

Issue #1189, opened 2023-07-10
Twitter Card Not Showing Course Images


pt2302:
## Description

In https://github.com/mitodl/ocw-hugo-themes/pull/1177, the `og:image` was updated to use a fully-qualified URL. However, this still does not result in working Twitter cards: the course image does not display properly when a link to a course is shared via Twitter. Twitter-specific tags should be added to make the Twitter cards work.

PR #1190, opened 2023-07-10
Adding Twitter-specific tags

pt2302:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1189.

Adds the tags `twitter:site` and `twitter:image:src` to all courses, allowing them to be shared via Twitter with the course image displaying correctly.

# How can this be tested?
To test this locally, ensure that `RESOURCE_BASE_URL=` is set to any empty string in the `.env` file. Run `yarn start <course-id>` for any course. Inspect the page source, and verify that `twitter:site` points to `@mitocw` and `twitter:image:src` has a fully-qualified URL pointing to the course image.

### DIFF
--- a/tests-e2e/ocw-ci-test-course/homepage.spec.ts
+++ b/tests-e2e/ocw-ci-test-course/homepage.spec.ts
@@ -63,36 +63,24 @@ test.describe("Course info", () => {
   })
 })
 
-test("og:image tag has fully-qualified URL", async ({ page }) => {
+test("Has expected meta tags in <head>", async ({ page }) => {
   const course = new CoursePage(page, "course")
   await course.goto()
   const metaShareImage = page.locator('meta[property="og:image"]')
+  const metaTwitterSite = page.locator('meta[name="twitter:site"]')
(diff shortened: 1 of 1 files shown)


tests-e2e/ocw-ci-test-course/homepage.spec.ts
```diff
@@ -72,3 +72,27 @@ test("og:image tag has fully-qualified URL", async ({ page }) => {
     "https://live-qa.ocw.mit.edu/courses/123-ocw-ci-test-course-fall-2022/example_jpg.jpg"
   )
 })
+
+test("twitter:site tag points to correct Twitter account", async ({ page }) => {
```
ChristopherChudzicki, line 76: **Comment:** I would be inclined to have a single test, `test("Has expected meta tags in <head>", () => {...})` or something like that.

Playwright is pretty fast, but each of these tests does take ~0.5s on my machine, which adds up eventually. (It's certainly not a problem, but I don't really see the value in separate tests here. If one of them fails, the failing assertion should clearly identify which test failed.)

ChristopherChudzicki (approved):

PR #1185, opened 2023-07-07
replace all instances of window.onload with addEventListener

ibrahimjaved12:
# What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/1749

# Description (What does it do?)
It replaces all instances of `window.onload` with `window.addEventListener` to allow multiple listeners for `load`.

This fixes the problem of video player not working when image gallery (nanogallery2) is in the same page for course-v2 theme.

# Context of this PR:
At the time of the original issue, the branch name reflected that issue but after the [review comment](https://github.com/mitodl/ocw-hugo-themes/pull/1179#pullrequestreview-1514561398) the scope of the PR changed and so I wanted to change the branch name to accurately reflect that. Hence this is a new PR for same issue.

# How can this be tested?
For the course-v2 online theme, follow same steps mentioned here: https://github.com/mitodl/ocw-hugo-themes/pull/1179

To test for offline theme:
First, checkout to this branch. Then:

### A. For nanogallery2 and videojs:
Follow the same steps mentioned in this PR: https://github.com/mitodl/ocw-hugo-themes/pull/1160
But use this course instead: `16.885j-fall-2005`. This is because we want to use a page that has both a) video b) image gallery in it. This is roughly around 4gb in its size when you download it, you can use that course, or any other course where this requirement is met of having both kind of resources in single page. 
Go to "related resources" from the left panel. And you will find a video on page with image gallery underneath it.
The video's name is: `ocwnasalaunch-220k.mp4`

So first you will be testing this with internal videojs youtube player, and then later you will be deleting it from the `content/static_resources` directory and then you will see if the videojs-youtube player appears on it and it works.

Similarly you will check for image-gallery underneath it that the it is working as expected.

### B. For fusejs:
Follow the same steps mentioned in this PR: https://github.com/mitodl/ocw-hugo-themes/pull/1175
Except that all the branches mentioned in the PR are merged with main now. 
So first make sure that you have fresh copy of ocw-www in your local ocw-content-rc, you can do that by deleting it from your local directory and then running `yarn start www` in `ocw-hugo-themes `and approving to download new copy of `ocw-www`.
Similarly, run `git pull origin main` in your local directory of `ocw-hugo-projects`. and `ocw-hugo-themes`. 

After that we can start following the given PR from 2nd step onwards.

PR #1191, opened 2023-07-11
Optimize imports using global flag variable to ensure videojs and nanogallery are loaded once only

ibrahimjaved12:
https://github.com/mitodl/hq/issues/1749

When we replaced `window.onload` with `window.addEventListener` to allow multiple listeners for `load`, we faced a new problem that a single kind of imports (eg. videojs imports) are called multiple times if there are multiple videos in a page. Not only that was not optimized, but this also caused some problems with the typescript functionality of download button popup (opening and closing) on videojs (online) videos.

This PR fixes this problem by using a global flag variable to keep track of videojs/nanogallery2 dynamic imports and to ensure they execute only once.

# How can this be tested?
1. Make sure the videos play + the video download button / related resources tab etc works as expected.

The functionality of the download button is as:
Click on video's download button, a popup should appear. Click on it again, it should close. Or, click outside the area, and it should close. Or click on another video's download button within the same page, the first video's popup should close and the other one's should open. (If it has Download Video or Download Transcript present)

The edge cases that caused problems:
1. If there were multiple dynamic imports present in a single page using `window.onload` only the last one would execute, hence we switched to using `AddEventHandler`. The original approach caused issues in the course where we had both image gallery+video present in a single page. (16.885j-fall-2005 --> related resources)
2. With the AddEventHandler, we had problem in multiple videos in a single page, because multiple import calls were made. (18.02sc-fall-2010 -- [57th session](https://live-qa.ocw.mit.edu/courses/18-02sc-multivariable-calculus-fall-2010/pages/3.-double-integrals-and-line-integrals-in-the-plane/part-b-vector-fields-and-line-integrals/session-57-work-and-line-integrals/)) Here we had problem we download button popup not working as expected, which is the reason for this PR.

And then of course we will want to handle these and every case for offline theme as well. 
**The steps to test this would be same as this: https://github.com/mitodl/ocw-hugo-themes/pull/1185 but a more thorough test with regards to videojs/nanogallery2 functionalities such as combinations of such different edge cases.**

Now as I understand, for offline theme, we do not have the video download button working as part of functionality because it's offline theme. And if we have other video tabs, like related resources, or "view video page", we would show the download icon on the right but it still shouldn't work. And if we don't have those additional tabs, then we wouldn't have the tab with download button icon at all. Please confirm this @ChristopherChudzicki.

### DIFF
--- a/course-offline/assets/course-offline.ts
+++ b/course-offline/assets/course-offline.ts
@@ -1,4 +1,5 @@
 import "video.js/dist/video-js.css"
+
 import "offcanvas-bootstrap/dist/js/bootstrap.offcanvas.js"
 import "promise-polyfill/src/polyfill.js"
 import "../../course-v2/assets/css/course-v2.scss"
@@ -17,7 +18,6 @@ import "nanogallery2/src/jquery.nanogallery2.core.js"
 import "nanogallery2/src/css/nanogallery2.css"
 import "videojs-youtube"
 import videojs from "video.js"
-import _ from "lodash"
 
 export interface OCWWindow extends Window {
   $: JQueryStatic
@@ -27,8 +27,6 @@ export interface OCWWindow extends Window {
 
 declare let window: OCWWindow
 
-const initVideoJS = _.once(() => (window.videojs = videojs))
-
 $(function() {
   initCourseDescriptionExpander(document)
   initCourseInfoExpander(document)
@@ -37,5 +35,5 @@ $(function() {
   checkAnswer()
   showSolution()
   initCourseDrawersClosingViaSwiping()
-  initVideoJS()
+  window.videojs = videojs
 })
--- a/course-offline/assets/course-offline.ts
+++ b/course-offline/assets/course-offline.ts
@@ -1,5 +1,4 @@
 import "video.js/dist/video-js.css"
-
 import "offcanvas-bootstrap/dist/js/bootstrap.offcanvas.js"
 import "promise-polyfill/src/polyfill.js"
 import "../../course-v2/assets/css/course-v2.scss"
@@ -18,6 +17,7 @@ import "nanogallery2/src/jquery.nanogallery2.core.js"
 import "nanogallery2/src/css/nanogallery2.css"
 import "videojs-youtube"
 import videojs from "video.js"
+import _ from "lodash"
 
 export interface OCWWindow extends Window {
   $: JQueryStatic
@@ -27,6 +27,8 @@ export interface OCWWindow extends Window {
 
 declare let window: OCWWindow
 
+const initVideoJS = _.once(() => (window.videojs = videojs))
+
 $(function() {
   initCourseDescriptionExpander(document)
   initCourseInfoExpander(document)
@@ -35,5 +37,5 @@ $(function() {
   checkAnswer()
   showSolution()
   initCourseDrawersClosingViaSwiping()
-  window.videojs = videojs
+  initVideoJS()
 })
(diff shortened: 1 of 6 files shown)


course-offline/assets/course-offline.ts
```diff
@@ -18,6 +17,7 @@ import "nanogallery2/src/jquery.nanogallery2.core.js"
 import "nanogallery2/src/css/nanogallery2.css"
 import "videojs-youtube"
 import videojs from "video.js"
+import _ from "lodash"
```
ChristopherChudzicki, line 20: @ibrahimjaved12 My apologies for suggesting a lodash approach. I had been under the impression we were already using lodash in the course bundle, but I guess we aren't. Given that, I think it's better to not use `_.once` and just avoid the double initialization with a simple flag.

ChristopherChudzicki (changes requested):

course-offline/assets/course-offline.ts
```diff
@@ -27,6 +27,8 @@ export interface OCWWindow extends Window {
 
 declare let window: OCWWindow
 
+const initVideoJS = _.once(() => (window.videojs = videojs))
+
```
ChristopherChudzicki, line 31: Is there a reason to make this change? I don't think the `window.videojs = videojs` assignment was problematic.

course-v2/assets/course-v2.ts
```diff
@@ -26,13 +34,14 @@ $(function() {
   initCourseDrawersClosingViaSwiping()
 })
 
-// @ts-expect-error for window.initVideoJS()
+window.videoJSLoaded = false
+window.nanogallery2Loaded = false
```
ChristopherChudzicki, line 38: I don't think there's any reason to put these on the global `window` object. We could scope them:

```ts
let videoJSLoaded = false
window.initVideoJS = () => {
  if (videoJSLoaded) return
  initVideoDownloadPopup()
  import("./videojs-imports").then(module => {
    module.initVideoJS()
  })
}
```

Or we could create a helper version of `once` without lodash:
```ts
const once = (callback: () => void) => {
  let hasRun = false
  if (!hasRun) {
    callback()
  }
}
```

Either way, I think it's worth putting fewer things on the global object.

ChristopherChudzicki (changes requested):

ChristopherChudzicki (approved):

course-v2/assets/course-v2.ts
```diff
@@ -26,13 +32,20 @@ $(function() {
...
     module.initVideoJS()
   })
+  videoJSLoaded = true
 }
 
-// @ts-expect-error for window.initNanogallery2()
-window.initNanogallery2 = () => import("./nanogallery2-imports")
+window.initNanogallery2 = () => {
+  if (nanogallery2Loaded) return
+  import("./nanogallery2-imports")
```
ChristopherChudzicki, line 49: @ibrahimjaved12 By the way: I expect that this import was not actually problematic, but it doesn't hurt to make it happen only once.

Suppose:
```ts
//somefile.js
console.log("Hello from somefile.js")

// main file:

import("./somefile").then(() => {
    console.log('Somefile has loaded, A')
})

import("./somefile").then(() => {
    console.log('Somefile has loaded, B')
})
```
The logs should be:
```

```
Hello from somefile.js
Somefile has loaded, A // The import promise has resolved, so the `then` runes
Somefile has loaded, B // The import promise has resolved (but does not re-run the imported code)
```

When you import a file, its contents only get run once.

PR #1194, opened 2023-07-12
Allow hiding course download button

pt2302:
Closes https://github.com/mitodl/hq/issues/1751.

If the `Hide Course Download` setting is set to `true` in the course metadata, this hides the link to the download page and the button to actually download the course.

# How can this be tested?

Run `yarn start course <course-id>` for any course. Navigate to the course's local markdown, and modify the `data/course.json` file to have a field with the value `"hide_download": true`. Verify that there is no course download button at `localhost:3000/download` and that the link to the download page does not appear. Change `"hide_download": false`, and verify that the course download button appears. Now, delete the field entirely, and verify that the button appears and that everything else works as expected.


HussainTaj-arbisoft (approved):
LGTM 👍

Issue #1203, opened 2023-07-16
Hide all resources when Hide Download is True


pt2302:
## Description

The course download button is disabled when `Hide Course Download` is set to `true` in OCW Studio (this feature was added in https://github.com/mitodl/ocw-hugo-projects/pull/257 and https://github.com/mitodl/ocw-hugo-themes/pull/1194). However, for courses that do have downloadable resources, the list of resources is still visible at `<course_url>/download`.

## Expected Behavior

Even if the course contains resources, the list of resources should not be displayed if `Hide Course Download` is set to `true`.

## Current Behavior

If a course contains resources, the list of resources is still visible even when `Hide Course Download` is set to `true`.

PR #1204, opened 2023-07-16
Hide all downloadable resources when hide course download is set

pt2302:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1203.

Following up on https://github.com/mitodl/ocw-hugo-themes/pull/1194, this PR now hides all resources (if any) on the course download page when `Hide Course Download` is set to `true`, and replaces them with the text `No Resources Found.`

# How can this be tested?
Run `yarn start course <course-id>` for any course. Navigate to the course's local markdown, and modify the `data/course.json` file to have a field with the value `"hide_download": true`. Verify that there is no course download button at `localhost:3000/download` and that the link to the download page does not appear on any course pages. Change `"hide_download": false`, and verify that the course download button appears (along with the list of resources, if any).

This should be tested with courses that have no downloadable resources, such as `res.18-005-spring-2010`, and also with courses that do have downloadable resources, such as `sts.462-spring-2006`.

### DIFF
--- a/tests-e2e/ocw-ci-test-course/download.spec.ts
+++ b/tests-e2e/ocw-ci-test-course/download.spec.ts
@@ -7,3 +7,13 @@ test("Download button exists for test course", async ({ page }) => {
   const downloadButton = page.getByRole("link", { name: "Download course" })
   await expect(downloadButton).toBeVisible()
 })
+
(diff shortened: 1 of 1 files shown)


pdpinch:
I thought we decided not do this

ChristopherChudzicki (approved):

PR #1207, opened 2023-07-18
Course theme accessibility fixes (aria-hidden and aria-expanded)

ibrahimjaved12:
#### Pre-Flight checklist

- [x] Testing
  - [x] Changes have been manually tested for online + offline theme
  - [x] Changes have been tested using VoiceOver (Mac's screen reader)

Closes https://github.com/mitodl/hq/issues/1972
Closes https://github.com/mitodl/hq/issues/1973

This PR fixes accessibility related issues on course pages. It focuses on `aria-hidden` and `aria-expanded`.
Apart from the above mentioned issues (sub-tasks), it also fixes the mentioned 3 problems: https://github.com/mitodl/hq/issues/1918#issuecomment-1645290474

# How can this be tested?
We will be using this course to test changes: `16.885j-fall-2005`
This is not the same course as mentioned in the issues (https://ocw.mit.edu/courses/res-7-005-biology-teaching-assistant-ta-training-fall-2021/download/). We need this course to check for other changes. Also, [the course](https://github.mit.edu/mitocwcontent/RES.7-005-Summer-2020) in` ocw-content-rc` it had some missing content which was needed for testing. (I tried changing `RESOURCE_BASE_URL` but the resources in `./download` still would not appear.) If anyone knows this fix, please let me know.

But we don't have to test for that course since these are theme related changes. So we will be using `16.885j-fall-2005 `instead. 
Checkout to this branch,
Run `yarn start course 16.885j-fall-2005 `.
Go to `http://localhost:3000/download/`

And then, check for the issues mentioned in https://github.com/mitodl/hq/issues/1972. And then check for https://github.com/mitodl/hq/issues/1973. And then check for the 3 fixes I mentioned here https://github.com/mitodl/hq/issues/1918#issuecomment-1645290474. 

All of these changes are present in `http://localhost:3000/download/`.
Except one change, for resource file download. Click on any resource and observe the `Download File` button.
You can use this link: `http://localhost:3000/resources/cohen_bio/`

If you're on Mac, you can also use Mac's VoiceOver to confirm how screen reader now reads those fixes, as compared to before: https://ocw.mit.edu/courses/16-885j-aircraft-systems-engineering-fall-2004/download/

I also tested the changes for offline theme (because it uses the same files) just to be safe. But testing for just online (course) theme should be sufficient.

### DIFF
--- a/course-v2/layouts/partials/resource_list_item.html
+++ b/course-v2/layouts/partials/resource_list_item.html
@@ -14,10 +14,11 @@
         {{ $downloadableLink := partial "get_resource_download_link.html" .params }}
           <span class="pr-2">
             {{ if $downloadableLink }}
-              <a href="{{- $downloadableLink -}}" target="_blank" download>
+              <a href="{{- $downloadableLink -}}" aria-label="Download file" target="_blank" download>
                 <img
                   class="hide-offline resource-download"
                   src="/static_shared/images/download.svg"
+                  aria-hidden="true"
                 />
               </a>
             {{ end }}
--- a/base-theme/layouts/partials/resource_body_v2.html
+++ b/base-theme/layouts/partials/resource_body_v2.html
@@ -15,7 +15,7 @@
       {{- if .Params.file -}}
       <div class="row">
           <div class="col-12 d-flex flex-direction-row align-items-center">
-              <a class="download-file" href="{{ partial "resource_url.html" (dict "context" . "url" .Params.file) }}"><span class="material-icons">file_download</span> Download File</a>
+              <a class="download-file" href="{{ partial "resource_url.html" (dict "context" . "url" .Params.file) }}"><span aria-hidden="true" class="material-icons">file_download</span> Download File</a>
           </div>
       </div>
       {{- end -}}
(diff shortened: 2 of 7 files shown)


HussainTaj-arbisoft (commented):
~Can we add `aria-hidden` to this icon as well?~ Now that I think about it, this button downloads the file. So hiding it won't be a good idea.

Screenshot 2023-07-25 at 1 11 29 PM

course-v2/layouts/partials/resource_list_item.html
```diff
@@ -14,10 +14,11 @@
         {{ $downloadableLink := partial "get_resource_download_link.html" .params }}
           <span class="pr-2">
             {{ if $downloadableLink }}
-              <a href="{{- $downloadableLink -}}" target="_blank" download>
+              <a href="{{- $downloadableLink -}}" aria-label="Download file" target="_blank" download>
                 <img
                   class="hide-offline resource-download"
                   src="/static_shared/images/download.svg"
+                  aria-hidden="true"
```
HussainTaj-arbisoft, line 21: Nice 👍

HussainTaj-arbisoft (approved):
LGTM 👍

PR #1211, opened 2023-07-19
no roboto font

ChristopherChudzicki:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1314

Removes roboto font.

The screenshot is WITH the change. The pasted in bit is the old version.

Screenshot 2024-02-09 at 9 41 34 AM

# How can this be tested?
View a video page and confirm it looks almost exactly the same as it does on the live site.


ibrahimjaved12 (approved):
LGTM 👍

PR #1216, opened 2023-08-03
feat: add mobile device warning to download course page

HussainTaj-arbisoft:
closes https://github.com/mitodl/hq/issues/2065

Adds a message to the download course page.

The message only appears on touch devices. I've used the `pointer` property for this purpose. [MSDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer#:~:text=a%20pointing%20device.-,coarse,-The%20primary%20input). This might not be perfect but works in most cases from what I've seen.

demo4

# How can this be tested?

## Prerequsites

* You'll need a markdown of a course. Any course would work. You can pick one from here https://github.mit.edu/ocw-content-rc/. Place this in your local ocw-content folder.

## Steps

1. Navigate to your local ocw-hugo-themes setup. Checkout branch `hussaintaj/2065`.
2. Run
    ```
    yarn start course course-id
    ```
3. Navigate to [localhost:3000/download](http://localhost:3000/download)
4. Open devtools, toggle the device toolbar and select a mobile device to view the page. Alternatively, open the link on your phone `<your-local-ip>:3000/download`.
5. **Expected Behavior**: You see the new message clearly. Please take a look at the screenshots for reference. The message should only appear on mobile (touch) devices.

### DIFF
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -167,6 +167,14 @@ pre {
   }
 }
 
+.touch-device-only {
+  display: none;
+
+  @media (pointer: coarse) {
+    display: block;
+  }
+}
+
 // TODO: overriding some css of header and footer in this new theme
 // when we roll out this theme completely then we can move this css to thier respective files in base-theme
 
--- a/course-v2/assets/css/course-v2.scss
+++ b/course-v2/assets/css/course-v2.scss
@@ -167,10 +167,10 @@ pre {
   }
 }
 
-.touch-device-only {
+.limited-pointer-only {
   display: none;
 
-  @media (pointer: coarse) {
+  @media (pointer: coarse), (pointer: none) {
     display: block;
   }
 }
--- a/course-v2/layouts/partials/resources_header.html
+++ b/course-v2/layouts/partials/resources_header.html
@@ -8,11 +8,6 @@
 {{- else -}}
 <div class="mb-2">
   <h2>Download</h2>
-  <div class="alert-warning p-2 touch-device-only">
-    <strong>Note: </strong>
-    The downloaded course may not work on mobile devices.
-    We recommend using a desktop computer.
-  </div>
   <div class="hide-offline download-course-container background-color-light-grey p-3">
     <div class="row align-items-center">
       <div class="d-flex justify-content-center col-12 col-md-3 px-3 pb-2 pb-md-0">
@@ -26,6 +21,11 @@
         For help downloading and using course materials, read our {{ partial "link.html" (dict "href" $zenDeskUrl "text" "FAQs" "target" "_blank") }}.
       </div>
     </div>
+  </div>
+  <div class="alert-warning p-2 touch-device-only">
+    <strong>Note: </strong>
+    The downloaded course may not work on mobile devices.
+    We recommend using a desktop computer.
   </div>
 </div>
 {{- if $noResourcesFound -}}
--- a/course-v2/layouts/partials/resources_header.html
+++ b/course-v2/layouts/partials/resources_header.html
@@ -9,7 +9,7 @@
 <div class="mb-2">
   <h2>Download</h2>
   <div class="hide-offline download-course-container background-color-light-grey p-3">
-    <div class="row">
+    <div class="row align-items-center">
       <div class="d-flex justify-content-center col-12 col-md-3 px-3 pb-2 pb-md-0">
         <a class="download-course-button p-2" href="{{ $downloadCourseUrl }}">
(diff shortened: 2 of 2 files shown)


course-v2/assets/css/course-v2.scss
```diff
@@ -167,6 +167,14 @@ pre {
   }
 }
 
+.touch-device-only {
+  display: none;
+
+  @media (pointer: coarse) {
+    display: block;
+  }
```
HussainTaj-arbisoft, line 175: Even though some devices have both, touch and mouse, this rule only checks the primary device. On PCs, this is a mouse. On tablets, this is touch. I tested this rule on a touch laptop and this stands true.

pdpinch: Ok, the logic makes sense to me. I don't know if your PR reviewer will have a different opinion.

course-v2/layouts/partials/resources_header.html
```diff
@@ -8,8 +8,13 @@
 {{- else -}}
 <div class="mb-2">
   <h2>Download</h2>
+  <div class="alert-warning p-2 touch-device-only">
+    <strong>Warning</strong>
```
pdpinch, line 12: I think "warning" is too strong a word for this situation. 

```suggestion
    <strong>Note: </strong>
```

That said, the class name `"alert-warning"` is fine.

pdpinch:
Please move the "note" to the bottom, below the "This package..." message.

course-v2/layouts/partials/resources_header.html
```diff
@@ -22,6 +22,11 @@ <h2>Download</h2>
       </div>
     </div>
   </div>
+  <div class="alert-warning p-2 touch-device-only">
+    <strong>Note: </strong>
+    The downloaded course may not work on mobile devices.
+    We recommend using a desktop computer.
```
pdpinch, line 28:
```suggestion
    We recommend using a computer.
```

Someone noted that it works fine on a laptop, so we can remove "desktop"

pdpinch, line 28:
```suggestion
    We recommend using a computer with the downloaded course package.
```

small tweak

course-v2/assets/css/course-v2.scss
ibrahimjaved12, line 175: I think we will also want to cover devices that have `pointer: none`.
We could use something like:
```
@media (pointer: coarse), (pointer: none) {
    .touch-device-warning {
        display: block;
    }
}
```
or maybe `not` condition with `pointer:fine` --> `display: block`?
We are looking for any devices which is not a desktop or a laptop (or a touch laptop) and thus will likely not be able to run the .html files as expected, so I think it's alright to use this condition.

I did check on a smart TV, and the warning didn't appear there with the current code (because it has `pointer: none`).

ibrahimjaved12 (changes requested):

course-v2/layouts/partials/resources_header.html
```diff
@@ -22,6 +22,11 @@ <h2>Download</h2>
       </div>
     </div>
   </div>
+  <div class="alert-warning p-2 touch-device-only">
```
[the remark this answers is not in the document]
HussainTaj-arbisoft: alert-warning is an existing class and style that we have in our code. I think it is used in one other place related to videos.

It's a generic class for our warning messages.

course-v2/assets/css/course-v2.scss
HussainTaj-arbisoft: Thanks! You're right. I'll add that.

course-v2/layouts/partials/resources_header.html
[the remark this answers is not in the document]
ibrahimjaved12: I meant the`.touch-device-only` class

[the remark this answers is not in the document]
HussainTaj-arbisoft: `display: none` is the default rule. The relevant code is under the media query which has `display: block`.

I understand your perspective. I look at this from an engineer's perspective. When you see a code like `<div class="touch-device-only">` it gives the idea that this element will only be rendered for touch devices. Which is the behavior of this class.

And since this class can be used in other places and is not limited to warning messages, I thought a more generic name would be more suitable.

Let me know if you still find the name ambiguous.

PS: I switched it to `limited-pointer-only` because we now use it in places of limited pointer availability.

HussainTaj-arbisoft:
@pdpinch I just need a confirmation. This is good to go, right?

pdpinch (approved):
OCW team is happy with the text now. Let's ship it!

PR #1225, opened 2023-08-15
feat: add file size in resource list and resource pages

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/1681

Adds file size in resource list pages and individual resource pages. Refer to figma designs in https://github.com/mitodl/hq/issues/1681#issuecomment-1746874289

In this PR, apart from the SCSS changes, we:
* Remove the `get_resource_thumbnail_attributes.html` partial
* Use new partial `get_resource_category.html` instead (we no longer have imgs for thumbnails)
* Introduced a `file_size_formatter.html` partial for calculating file size in respective units (converted from bytes)
* Also changed the styling of resource body page in general (apart from the file sizing fixes). Previously we used `container` class (this is still used in a lot of places in OCW) but removed it for this page as per @mbilalmughal recommendation. It was less aesthetically pleasing. Instead, for resource body page we have `resource-page-container`. 
* There's only reformatting changes in `resource_list_collapsible.html`

# How can this be tested?
Test rendering for (to ensure nothing has broken for moving resource partials from course-v2 to base-theme):
www online + www offline
course online + course offline.

Steps:
First checkout to this branch.
`yarn build /to/ocw-content-rc/ocw-www or a course/ /to/ocw-hugo-projects/ocw-www or ocw-course-v2/config-offline.yaml`  For offline themes.

`yarn start www` For online www
Make sure the builds succeed ^
from dist directory do check that the theme is alright too from any index pages.

For course, we will want to:
1. `yarn start course 6.0002-fall-2016` (using this course because this has file size content)
2. Go to: http://localhost:3000/download/
3. Check the UI.
4. Click on a resource (other than video lectures)
5. Check the UI. Also you might want to use bigger resource names (like I have done in the screenshots) to ensure the UI does not break.
6. Also try different file sizes, too small, too big. Make sure the unit is okay.
7. Also check for cases where a resource might not have file size (just clear the `file_size` from the `.md` file of a resource)

For offline course theme, build a course and go to dist/index.html and then "Browse Resources". Steps will be same as above but not as thorough testing just make sure the changes have applied.

# References
Use these figma links for reference.
https://github.com/mitodl/hq/issues/1681#issuecomment-1746874289

- [x] Desktop screenshots

https://github.com/mitodl/ocw-hugo-themes/assets/109785089/94d91928-921e-4f2a-b4e3-b6c3a7d7a8cf

image
image
image
image
image
image
image
image
image
image
image
image


mbilalmughal:
@ibrahimjaved12 The UI looks good to me. 
One request, If you could decrease the space between the description and the text, that would be better.

However, as we have discussed, we won't be implementing any additional fixes at this time. For instance, the right-hand space on the mobile container is not equal to the right-hand side.

mbilalmughal:
@Ferdi Could you please take a look at this? I've reduced the file size label's font size and explained it in this [thread](https://mitodl.slack.com/archives/GQ89D0Z5M/p1692727016952879)

pdpinch:
@mbilalmughal can you explain what you and @ibrahimjaved12 have changed in response to Ferdi's comment? I don't recall the original design, so I'm not sure what's new.

mbilalmughal:
@pdpinch, we've changed the file size to a minimum of 10px. Please see the attached image. @ibrahimjaved12 has not pushed the code yet, we are waiting for a response.

Originally, the file size font and the title font were the same, but now the file size font has been reduced to 10px.

image

mbilalmughal:
Before, also attached with this PR

image

pdpinch:
@ibrahimjaved12 we've decided to make another pass on the design for this. I expect this will be on hold for at least another week.

base-theme/layouts/partials/file_size_formatter.html
```diff
@@ -0,0 +1,52 @@
+{{ $fileSize := int . }}
+{{/*  Cap at 999 TB  */}}
+{{ if gt $fileSize 1098412116148224 }}{{ $fileSize = 1098412116148224 }}{{ end }}
+{{ $modifiedFileSize := 0 }}
+{{ $units := slice "B" "KB" "MB" "GB" "TB" }}
```
ChristopherChudzicki, line 5:
```suggestion
{{ $units := slice "B" "kB" "MB" "GB" "TB" }}
```

I can look more later if no one else gets to it; just gave a quick skim.

course-v2/layouts/partials/resource_list_collapsible.html
```diff
@@ -1,39 +1,39 @@
...
+          {{ partial "see_all.html" (dict "permalink" .Permalink) }}
+          </div>
+        </div>
+      {{ end }}
     </div>
-  </div>
-  {{ end }}
+    {{ end }}
   </div>
 {{ end }}
```
ibrahimjaved12, line 39: No technical changes here just reformatted.

base-theme/layouts/partials/file_size_formatter.html
ibrahimjaved12: Thanks @ChristopherChudzicki

course-v2/layouts/partials/resource_list_item.html
```diff
@@ -1,31 +1,35 @@
...
+        {{ $resourceCategory := partial "get_resource_category.html" (dict "resourceType" .params.resourcetype "fileType" .params.file_type) }}
+        <div class="resource-type-thumbnail {{ $resourceCategory }}">
+          <div class="d-flex align-self-center">{{- $resourceCategory -}}</div>
+        </div>  
+        <div class="download-container{{if not .params.file_size}} pb-1{{end}}">
+          {{ if not .hideDownloadIcon }}
+            {{ $downloadableLink := partial "get_resource_download_link.html" .params }}
+            <span class="download-link">
+              {{ if $downloadableLink }}
+                <a href="{{- $downloadableLink -}}" target="_blank" download>
```
gumaerc, line 14: Putting the anchor tag here results in only the download icon, not the entire button, being clickable. The resource type above and the file size below show an I-beam cursor when you hover over them. The areas to the left and right of the tiny download icon are not even clickable, just the icon itself. Do we want to move this anchor tag to the outside, so that the button as a whole is clickable?

gumaerc (changes requested):
Everything looks great here and there doesn't seem to be any adverse affects from moving partials around. I generated file sizes for a number of courses and tested them, videos in the hundreds of MB look fine, and smaller files also look fine as well. I don't see any other problems with the code, but I do have one concern that I think should be addressed:

course-v2/layouts/partials/resource_list_item.html
ibrahimjaved12: I do think that it might not be very user friendly to only have such a small clickable space (download icon). 
But let's get a second opinion from @mbilalmughal on this.

@mbilalmughal do you think we should make the whole icon clickable for download, not just the "small download icon" inside of it?
image

mbilalmughal: @ibrahimjaved12 @gumaerc To ensure good UX practice, the "Hit Area" of any CTA (call to action) should be at least the size of the thumb, which is typically around 30 to 50 pixels, making it easy for the user to tap or click.

@ibrahimjaved12 I would say Carey is suggesting the right thing; the entire icon should be clickable, and it would be so nice of you if you could add a little [bounce effect](https://codepen.io/halink0803/pen/qdVXqm) to the download icon when hovering.

ibrahimjaved12: Thanks @mbilalmughal .
Sure, let's do this.

ibrahimjaved12: Done. Added the requested changes. Also did a little bit of refactoring. I noticed that `resource_url.html` was used to get the downloadable URL of resource, but we already had it as `$downloadableLink`.

I'm still adding a video demonstration of the animation in the PR but it is ready to be re-reviewed

ibrahimjaved12:
Animated hover video:

https://github.com/mitodl/ocw-hugo-themes/assets/109785089/94d91928-921e-4f2a-b4e3-b6c3a7d7a8cf

gumaerc (approved):
:+1:

PR #1227, opened 2023-08-21
Stop layout shift on videos

### LINKED ISSUE #1213: Stop layout shift on videos
Currently, videos shift layout when loading:

https://github.com/mitodl/ocw-hugo-themes/assets/9010790/ad47ee66-6e20-4a3b-bcbc-1e9d2a0b1d62

I believe this began happening with #1071. We should still lazy-load VideoJS, but we should style the containing div to reduce layout shift.

*Search Console has flagged [Cumulative Layout Shift](https://web.dev/cls/) as an issue on a few pages, and I suspect videos like this are the cause.*

ibrahimjaved12:
Closes #1213 

Stops the layout shift on video player when `videojs` has not yet dynamically loaded by adding height and width styling.

# Testing

We will test for:
1. Regular videos
2. Embed videos
3. Offline videos (youtube player)
4. Offline videos (local video player)

**How to test:**
I tested using `low end mobile` throttling setting in inspect element. And refreshing the page several times, and stopping the loading right after the page opens. Make sure there is no shift in layout (as demonstrated in the video in the issue).

For embed videos and regular videos, let's use this course: `18.06-spring-2010`
1. `yarn start course 18.06-spring-2010`
2.  This is an example of embed video: http://localhost:3000/pages/instructor-insights/
3.  This is an example of regular video: http://localhost:3000/resources/an-interview-with-gilbert-strang-on-teaching-linear-algebra/
4.  Test the layouts in both cases^

For offline theme, use any course whose content (static resources -- a video) you already have downloaded to ensure local video player runs as expected after this change. If you don't have that, use the testing instructions in this PR to download course contents: https://github.com/mitodl/ocw-hugo-themes/pull/1160

Afterwards, if you haven't made a build, do so with:
`yarn build /path/to/18.06-spring-2010/ /path/to/ocw-hugo-projects/ocw-course-v2/config-offline.yaml`

Go to your local course `18.06-spring-2010/dist` directory, open up `index.html` and go to video lectures. Go to lecture 2. Test the local video player layout. 

Now, go to `18.06-spring-2010/content/static_resources` and delete `02.mp4`
Run again: `yarn build /path/to/18.06-spring-2010/ /path/to/ocw-hugo-projects/ocw-course-v2/config-offline.yaml`
Go to lecture 2 again now from the above steps (dist directory). Test the youtube video player layout.

### DIFF
--- a/base-theme/layouts/partials/video_expandable_tab.html
+++ b/base-theme/layouts/partials/video_expandable_tab.html
@@ -5,17 +5,18 @@
   data-target=".{{- .tabClass -}}"
   data-toggle="collapse"
   aria-controls="{{- .tabClass -}}"
+  aria-expanded=""
   role="tab"
   >
   <div class="video-tab-header">
     {{ if .tabTitle }}
-      <span class="tab-title-section" aria-label="{{ .tabTitle }}">
+      <span class="tab-title-section">
         {{ if and .resourceUrl $isEmbedVideo }}
-          <a href="{{ .resourceUrl }}">
+          <a href="{{ .resourceUrl }}" aria-label="{{ .tabTitle }}">
           {{ partial "video_tab_title.html" (dict "tabTitle" .tabTitle) }}
           </a>
         {{ else }}
-          <button>
+          <button aria-label="{{ .tabTitle }}">
             {{ partial "video_tab_title.html" (dict "tabTitle" .tabTitle) }}
           </button>
         {{ end }}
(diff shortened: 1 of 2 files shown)


ibrahimjaved12:
> Looks good 👍 Could you add a comment explaining the 56.25, though?

We could also have used:
`
.youtube-container > video {
position: absolute;
    top: 0;
    left: 0;
    width: 100%;
    height: 100%;
}`
But that represents the local videojs player theme I think, not youtube-videojs player that we use for web:
image

But to explain 56.25:
I observed how div sizing is done after VJS has finished loading, and this was the code:
image
It does come from `vjs-fluid` class, I think the 9/16 translation is done to 56.25% there. We were also already using these dimensions in our code almost like a video-height sort of thing:
```
.video-page {
  max-width: $video-section-max-width;
  margin-bottom: 15px;

  .video-container {
    height: 0;
    padding-bottom: 56% !important;
    position: relative;
    max-width: 100%;
    margin-bottom: 1px;
  }
}
```

I tested the dimensions for different videos, for different screen sizes (mobile/ipad/desktop), all had 56.25% `padding-top`. Therefore I figured we could use these dimensions.

ChristopherChudzicki (approved):

PR #1233, opened 2023-08-25
fix: accessibility of video tabs

### LINKED ISSUE #1219: Accessibility of video tabs
## Accessibility of:
1.  `base-theme/layouts/partials/video_tab_title.html`
2.  `base-theme/layouts/partials/video_expandable_tab.html`
Code:
```
<i class="material-icons toggle md-18"></i>
<span class="tab-title">
  {{- .tabTitle -}}
</span>

```
Go to link: https://ocw.mit.edu/courses/3-091-introduction-to-solid-state-chemistry-fall-2018/resources/lecture-2/
And use a screen reader on "Transcript" and the video download icon.

They seem to be redundant. Like, each word is read multiple times for both of the buttons. I think this might be linked to `base-theme/layouts/partials/video_expandable_tab.html` as we have the relevant `aria-label="Download Video and Transcript"` there. Investigate both of these files. Also check for other tab titles: "Related Resources", "Lecture Notes", etc how a screen reader reads them.

## Screenshot
image

ibrahimjaved12:
Closes #1219 

Fixes accessibility related issues of video tabs by restructuring the code. This includes the toggleable video tabs and the download button.

# How can this be tested?
You may use a screenreader for testing or check the accessibility using inspect element. Make sure the screen reader would correctly read the text now (and no redundancy/repeats).
In addition to this, we also want to make sure that the UI does not break because there were styling changes. (Check for different screen sizes)

Here are some courses names:
1. 3.091-fall-2018 (Example of multiple video instances in a single page -- Goodie Bag Tutorials from left menu). Contains embed videos + click on 'View video page' for regular video pages.
2. 18.06sc-fall-2011 (Another example of embed videos -- Instructor Insights from left menu)
3. 16.885j-fall-2005 (Example of video + nanogallery in same page -- Related Resources from left menu)

Make sure to check for offline theme as well. Steps same as mentioned here for offline course theme only:
https://github.com/mitodl/ocw-hugo-themes/pull/1227

Checkout to this branch, `yarn start course <name>` and then do the above mentioned instructions.

To run playwright tests:
`yarn test:e2e video-tabs.spec.ts `

### DIFF
--- a/base-theme/assets/css/video.scss
+++ b/base-theme/assets/css/video.scss
@@ -1,5 +1,5 @@
 .video-tab-header {
-  padding: 5px 15px 5px;
+  padding: 5px 15px;
   border-top: 1px solid #3d3c3d;
   background-color: black;
   color: white;
@@ -14,6 +14,8 @@
     left: 2px;
   }
   .tab-title-section {
+    width: 100%;
+
     > * {
       padding-right: 10px;
       background: none;
@@ -59,7 +61,12 @@
   }
 }
 
-.video-tab-toggle-section[aria-expanded="true"] {
+.tab-toggle-button {
+  text-align: left;
+  width: 100%;
+}
+
+.tab-toggle-button[aria-expanded="true"] {
   i.toggle:after {
     content: "keyboard_arrow_down";
   }
@@ -80,7 +87,7 @@
   img.video-download-icon {
     height: 22px;
     position: relative;
-    left: 9px;
+    left: 5.25px;
     vertical-align: baseline;
   }
 }
--- a/base-theme/assets/css/video.scss
+++ b/base-theme/assets/css/video.scss
@@ -14,7 +14,8 @@
     left: 2px;
   }
   .tab-title-section {
-    width: 100%;
+    display: flex;
+    flex: 1;
 
     > * {
       padding-right: 10px;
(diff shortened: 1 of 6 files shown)


HussainTaj-arbisoft (changes requested):
When there is no transcript, the download button aligns itself to the left. I think it should be on the right, right? The rest seems good to me.

Screenshot 2023-08-31 at 4 18 04 PM

Another thing, I don't know if this is in the scope of the ticket but I can't seem to navigate inside the on-screen transcript. The screen reader assistant doesn't read the transcript when I click it, nor can I navigate into it using tabs. This might be intentional, I just wanted to bring it up in case it isn't.

ibrahimjaved12:
>When there is no transcript, the download button aligns itself to the left. I think it should be on the right, right? The rest seems good to me.
That's a good find. Thanks.

>Another thing, I don't know if this is in the scope of the ticket but I can't seem to navigate inside the on-screen transcript. The screen reader assistant doesn't read the transcript when I click it, nor can I navigate into it using tabs. This might be intentional, I just wanted to bring it up in case it isn't.

Yes, it has been designed this way it appears. We should probably raise that discussion somewhere to confirm this and open a new ticket if required.

HussainTaj-arbisoft (approved):
LGTM 👍 

PS: I could hear the transcription using the _read from here_ option in voice-over (by default control+option A).

PR #1242, opened 2023-09-11
Make notifications accessible

### LINKED ISSUE #1218: Accessibility of notification.html
## Accessibility of `i` in `base-theme/layouts/partials/notification.html`
Code:

```
<div id="notification_{{ .Params.uid }}" class="notification d-none">
  <div class="notification-message">
    {{ .Content }}
  </div>
  <div class="notification-close">
    <i class="material-icons">close</i>
  </div>
</div>

```
From code this almost seems okay without `aria-label,` but we should double check with a screen reader on the page just to be sure.

ibrahimjaved12:
Closes #1218 

Makes notifications accessible by adding `aria-label` and `role`.

# How can this be tested?
1. Checkout to this branch
3. Go to your local ocw-content-rc directory, and open up `ocw-www/content/2023-24hc.md`. Change `draft` to `false`. This should enable the notification.
4. `yarn start www`
5. Go to `localhost:3000` and you should see a notification bar on top of the page. You can also access it at: http://localhost:3000/notifications/2023-24hc/ (ignore the duplicated text at bottom)
6. The changes in this PR are with respect to the Close button on the right. It should now say "Close Notification, Button" if you run it by a screen reader.
7.  Make sure the close notification button works.


base-theme/layouts/partials/notification.html
ChristopherChudzicki, line 1: Is this partial actually used anywhere?

Searching the codebase for `notification.html`, I do not see any usages.  But I'm not 100% sure whether that's a surefire indication the file is unusued.

ChristopherChudzicki (changes requested):
A div should, as far as I know,  never have `role="button"`. 

-  `<button />` Automatically has role button, **and** click handlers are automatically fired by spacebar and enter key. 
- `<div role="button" ... >`: has role button, but does NOT behave like a button... 
    - it won't be tab-focusable (iirc)
    -  and click handlers will not be fired by space/enter

base-theme/layouts/partials/notification.html
ChristopherChudzicki: **Answer:** Yes, search for `partial "notification"`. 

The extension `.html` is not necessary, but is probably recommended since you can have markdown partials.

ibrahimjaved12: Yup, it's used in `baseof.html`

I'll add .`html` along with the button modifiication

ibrahimjaved12:
@ChristopherChudzicki 
I looked at the notifications from `www-offline` theme.. since we had the same notifications partial also called in the offline theme.

The notifications were not appearing -- that's because we have `d-none` in the class. In the online `www` theme, we have a JS function that removes `d-none` and also works on close button functionality. The JS file (`notification.js`) is for online `www` theme so these functionalities do not work in the offline theme.

I suggest that we talk about this in `product-ocw`, so we can decide whether we want to have notifications in the offline theme or not. If we do, we can open a separate issue for this and work on it. But I suggest for the scope of this PR we work on accessibility related fixes only.

PR #1245, opened 2023-09-19
fix: instructor(s) and topic(s) typo

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/2409

Adds conditional rendering to `Topic` and `Instructor` for their plural terms in `ocw-www` homepage theme.

# How can this be tested?
1. Checkout to this branch
2. yarn start www
3. Go to http://localhost:3000/
4. Under "New Courses" you should be able to see some courses that do not have a) Instructors text b) Topics text
5. Under "New Courses" you will also see some courses that have singular Instructor with text `Instructor:`, and the text `Instructors:` if there are more than one.
6. I'm not really sure how you would test `Topics` in the same manner without editing a new test course or making a new course for it to appear in the homepage. But we could test it using code:

`base-theme/layouts/partials/course_carousel_card.html:`
We have this line of code:
`{{ $firstThreeCourseTopics := partial "course_cards_topics_list.html" $courseTopics | default slice }}
`
Which is responsible for the topics shown. Normally for all of these cases, it has at least 2 topics so we are not able to see the singular heading. We could just force it to have single topic and see how it effects the layout.
```
{{ $singleTopic := slice }}
{{ $singleTopic := $singleTopic | append (index $firstThreeCourseTopics 0) }}
```
And then:
```
<span class="card-label">Topic{{ if gt (len $singleTopic) 1 }}s{{ end }}:</span> 
{{ partial "course_cards_topics.html" $singleTopic }}
```
Using this, we can confirm the text `Topic:` is shown with only one topic displaying.


HussainTaj-arbisoft (approved):
LGTM 👍

Issue #1252, opened 2023-09-27
Quiz questions: "Check" and "Show answer" should be <button> not <div> (keyboard interactivity)


ChristopherChudzicki:
## Expected Behavior
The "Check" and "Show answer" buttons should
- be focusable
- fire click event handlers when pressing spacebar/Enter

I.e., the should behave like normal browser buttons.

## Current Behavior
The "Check" and "Show/Hide answer" buttons on quiz questions are normal divs; they cannot be focused by keyboard users or interacted with.

## Steps to Reproduce
See https://ocw.mit.edu/courses/15-071-the-analytics-edge-spring-2017/pages/logistic-regression/the-framingham-heart-study-evaluating-risk-factors-to-save-lives/quick-question-225/

## Possible Solution
Use normal `<button>` elements.

PR #1253, opened 2023-09-28
Make multiple choice quiz buttons keyboard-interactive

pt2302:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1252.

This PR makes the `Show Solution`/`Hide Solution` and `Check` buttons in multiple-choice quizzes keyboard-interactive, which is needed for improved accessibility.

# How can this be tested?
1. Spin up a course with a multiple-choice quiz, for example `yarn start course 15.071-spring-2017`.
2. Navigate to a page with a multiple-choice quiz, such as `http://localhost:3000/pages/logistic-regression/the-framingham-heart-study-evaluating-risk-factors-to-save-lives/quick-question-225/`.
3. Verify that the `Check` and `Show Solution` buttons are keyboard-interactive. For the `Check` button, this means selecting one of the quiz options and then using the keyboard to push the `Check` button. For the `Show Solution` button, this does not require one of the options to be selected first.

### DIFF
--- a/course-v2/layouts/partials/quiz_multiple_choice_solution.html
+++ b/course-v2/layouts/partials/quiz_multiple_choice_solution.html
@@ -4,7 +4,7 @@
   </button>
   {{if .solution }}
   	&emsp;
-  	<button class="btn btn-outline-primary multiple-choice-show-button ">
+  	<button class="btn btn-outline-primary multiple-choice-show-button">
   		Show Solution
     </button>
   {{end}}


ChristopherChudzicki (changes requested):
The buttons are now keyboard interactable, but they have mostly default button styling, which is probably not what we want.

Screenshot 2023-09-28 at 12 06 25 PM

You can probably fix it with a `btn btn-link`, though I'd check what styles (if any) the class is applying now.

ChristopherChudzicki (approved):

Issue #1250, opened 2023-09-27
Quiz question: Inconsistent states


ChristopherChudzicki:
## Expected Behavior
On quiz questions, if the "Explanation" text is shown, then a green check identifying the correct answer is also shown.

## Current Behavior
"Explanation" can be shown without a green check.

## Steps to Reproduce
1. Click "Show answer", then click any other quiz option, as seen in video.
https://github.com/mitodl/ocw-hugo-themes/assets/9010790/bb2d6b29-225f-409b-9364-40faf56d79ad

## Possible Solution
Quiz questions are stateful. Two possible states are "Solution revealed" and "Solution not revealed".
1. Clicking another option should NOT clear the check mark if the question is in "Solution revealed" state, as indicated by text of show/hide answer button.
2. Clicking another option should fully reset the "Solution revealed" state

PR #1249, opened 2023-09-26
Toggle show solution for multiple-choice quizzes

pt2302:
# What are the relevant tickets?
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1248.

# Description (What does it do?)
Updates the text of the `Show Solution` button in multiple-choice quizzes to `Hide Solution` once the solution is shown.

# How can this be tested?
1. Spin up a course with a multiple-choice quiz, for example `yarn start course 15.071-spring-2017`.
2. Navigate to a page with a multiple-choice quiz, such as `http://localhost:3000/pages/logistic-regression/the-framingham-heart-study-evaluating-risk-factors-to-save-lives/quick-question-225/`.
3. Verify that clicking the `Show Solution` button displays the solution and the button text has been changed to `Hide Solution`. Click the `Hide Solution` button and verify that the solution is now hidden and the button text has been changed to `Show Solution`.

PR #1258, opened 2023-10-04
Make multiple-choice quiz state consistent

pt2302:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1250.

If the solution is shown on a multiple-choice quiz, this PR makes the change that selecting an option resets the state of the quiz to the "solution not shown" state. Previously, selecting an option would remove the checkmarks indicating correctness but would not hide the solution text.

# How can this be tested?
1. Spin up a course with a multiple-choice quiz, for example `yarn start course 15.071-spring-2017`.
2. Navigate to a page with a multiple-choice quiz, such as `http://localhost:3000/pages/logistic-regression/the-framingham-heart-study-evaluating-risk-factors-to-save-lives/quick-question-225/`.
3. (Optional) Select one of the quiz choices.
4. Click the `Show Solution` button. The solution should appear, and the button text should now read `Hide Solution` (feature added here: https://github.com/mitodl/ocw-hugo-themes/pull/1249)
5. Click on any quiz choice. Verify that the checkmarks disappear (this was the behavior prior to this PR) and now that the solution is hidden and the button now has the text `Show Solution` again.

### DIFF
--- a/course-v2/assets/js/quiz_multiple_choice.js
+++ b/course-v2/assets/js/quiz_multiple_choice.js
@@ -9,6 +9,13 @@ export const clearSolution = () => {
       ).forEach(solution => {
         solution.classList.remove("toggle-visible")
       })
+      const question = radio.closest(".multiple-choice-question")
+      question
+        .getElementsByClassName("multiple-choice-solution")[0]
+        .classList.remove("toggle-visible")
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki (approved):
👍

PR #1264, opened 2023-10-17
fix long single word resource title going out of div

ibrahimjaved12:
Related to https://github.com/mitodl/hq/issues/1681

I noticed that if we have a have a single worded long title, it breaks the `div`.

You can notice this behavior here: https://ocw.mit.edu/courses/3-091-introduction-to-solid-state-chemistry-fall-2018/download/
If you go under readings, and inspect element to something like 300 width you will see it goes out of the div. So basically if the text is longer than that, then it will be apparent for bigger screens too.

Problem:
image

Fix:
image

# How to test
1. checkout to this branch
2. yarn start course 6.0002-fall-2016
3. go to downloads and inspect element an existing title to add a long, one word title to a resource. It should not break the `div` box.

### DIFF
--- a/course-v2/assets/css/course-resources.scss
+++ b/course-v2/assets/css/course-resources.scss
@@ -155,16 +155,14 @@ h4 {
 
   .resource-list-title {
     text-decoration: none;
+    word-break: break-all;
   }
   .resource-thumbnail {
     text-decoration: none;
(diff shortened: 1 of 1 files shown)


gumaerc (approved):
:+1:

PR #1269, opened 2023-10-19
feat: add off-site navigation confirmation dialog

HussainTaj-arbisoft:
Closes https://github.com/mitodl/hq/issues/1812 

This PR:

- Adds an icon next to external sidebar links to distinguish them.
- Adds a confirmation dialog that opens before the page navigates to the external link.

## Choice of approach

A single modal's markdown is placed inside the `body` tag. The confirm action link is dynamically updated when a user clicks an external link.

Bootstrap's modals are sensitive to positioned parents. This is the reason why the modal needs to be placed inside the `body` tag.

## Design: The off-bits

You might observe that the output looks a bit different from the [designs](https://www.figma.com/proto/7TIcfINO8O8SPVyALe632H/File-Size?page-id=0%3A1&type=design&node-id=165-2&viewport=122%2C-1406%2C0.62&t=Z28hyDE4MOM5HpBE-1&scaling=min-zoom&starting-point-node-id=165%3A2&mode=design) (which might not be obvious). The reasons for this are:

* The design uses `Arial` while we use `Helvetica` font.
* The side nav link icons are aligned to the baseline of the text, which is in accordance with the design. But the design has no [descenders](https://mockitt.wondershare.com/dictionary/what-is-baseline.html).
* I've reused our existing font measurements, instead of creating new ones, to avoid tech debt later on. There is a slight, but almost negligible, difference in those font sizes.

The following image shows a stress test of the icon placement design.
Screenshot 2023-10-23 at 1 28 08 PM

Screenshot 2023-10-23 at 2 53 02 PM

Screenshot 2023-10-23 at 2 53 14 PM

https://github.com/mitodl/ocw-hugo-themes/assets/71316217/bfc87c2c-2ea7-4ab6-ab0f-9141c11be6a2

https://github.com/mitodl/ocw-hugo-themes/assets/71316217/82ec1ab8-dd69-4a01-9aa4-c676152f5535

# How can this be tested?

1. Grab content for a course from [this list](https://github.com/mitodl/hq/issues/1812#issuecomment-1745509175).
2. Place the content in your local content directory.
3. Navigate to your local setup of ocw-hugo-themes.
4. Checkout branch `hussaintaj/1812-external-link-modal`.
5. Run
    ```
    yarn start course <course-id>
    ```
6. Observe the external links in the sidebar.
    * Expect external links to be marked with an icon.
    * Expect a dialog to pop-up when you click on an external link.
7. Observe the confirmation dialog.
    * Expect that clicking "Stay Here", the close button, or outside the modal closes the dialog.
    * Expect that clicking on "Continue" takes you to the external link.


course-v2/assets/js/mobile_course_drawers.ts
```diff
@@ -1,17 +1,20 @@
+export const MOBILE_COURSE_NAV_DRAWER_ID = "mobile-course-nav"
+export const MOBILE_COURSE_INFO_DRAWER_ID = "course-info-drawer"
```
gumaerc, line 2: Making these into constants certainly isn't a bad thing, but I don't see them referenced anywhere else. Is it necessary? Why would just the drawer ID be a constant and not the other ID's?

gumaerc (approved):
This works great, tested with a bunch of different sites in all the browsers I have access to. I just noticed one minor thing that doesn't really matter all that much:

course-v2/assets/js/mobile_course_drawers.ts
HussainTaj-arbisoft: We use the`MOBILE_COURSE_NAV_DRAWER_ID` in the external link's click handler.

https://github.com/mitodl/ocw-hugo-themes/blob/f46f5d7d6fdadf303d737162c2f93d95bf34db83/course-v2/assets/js/external_link_modal.ts#L10

> Is it necessary?

Yes, we use it to close the drawer. Otherwise, the drawer captures external clicks and prevents navigation.

> Why would just the drawer ID be a constant and not the other ID's?

That's a valid question. Right now, only the drawer IDs have the potential for use outside of the original JS file. That's why I exported `MOBILE_COURSE_INFO_DRAWER_ID`, even though it's not in use at the moment. It might seem a bit odd to export just one drawer's ID.

The drawer IDs can be used for opening and closing actions. I didn't think the other IDs would have similar reuse possibilities, so I didn't export them.

pdpinch:
Thanks for the screen recording. I shared it with the OCW team. The design is great, BUT the video revealed an edge case that we didn't consider. 

One of those links goes to https://www.ocw-openmatters.org, which the OCW team considers to be an internal site. Is there a reasonable way to add it to the list of sites considered "internal" ?

HussainTaj-arbisoft:
> Is there a reasonable way to add it to the list of sites considered "internal"?

We can create a whitelist of sorts and treat those links as internal links. The current code can easily be extended to include this functionality.

We will need the complete list of domains and patterns to exclude. It will help us decide what string comparison approach to use.

I'll share the complete list of external URLs currently being used in production.

HussainTaj-arbisoft:
A complete list of external URLs is shared here: https://github.com/mitodl/hq/issues/1812#issuecomment-1778996149

pdpinch:
@HussainTaj-arbisoft what remains to be done here?

HussainTaj-arbisoft:
> @HussainTaj-arbisoft what remains to be done here?

@pdpinch, the exclusions remain to be implemented. I had a suggestion about it and I'm waiting for feedback on https://github.com/mitodl/hq/issues/1812#issuecomment-1788850100

HussainTaj-arbisoft:
@gumaerc, I've updated this PR according to our new requirements. It is ready to be looked at again.

There are now three types of behaviors links will show.
- Internal links: Links to internal content.
- External Link with warning: Shows a confirmation dialog before redirecting to the external link.
- External Link without warning: Redirects to the external link without a confirmation.

Menu items can now have a parameter called `includeLicenseWarning`. You may view an example [here](https://github.com/hussaintaj-mitodl/res.7-003-fall-2020/blob/main/config/_default/menus.yaml).

Testing instructions are the same as before. For reference and testing, you may take [this course content](https://github.com/hussaintaj-mitodl/res.7-003-fall-2020/). Alternatively, this could also be tested as a part of https://github.com/mitodl/ocw-studio/pull/2034.

HussainTaj-arbisoft:
@gumaerc, this PR is also on hold for now. Since it incorporates changes of https://github.com/mitodl/ocw-studio/pull/2034, it won't require a review unless we move ahead with that change.

HussainTaj-arbisoft:
Superseded by https://github.com/mitodl/ocw-hugo-themes/pull/1364

Issue #1279, opened 2023-11-14
playwright tests time out on video tests


gumaerc:
## Expected Behavior
All end to end tests should pass

## Current Behavior
The test titled "Transcripts start time matches video start time" times out and fails after 30 seconds when running `const activeCaption = await videoSection.transcript.activeLine().getAttribute("data-begin")`, after running `await videoSection.playButton().click()`

## Steps to Reproduce
1. Run `yarn test:e2e`

## Possible Solution
The problem seems to be related to the Chromium browser that Playwright downloads not being able to start playing back a video. If a solution cannot be found to get Chromium to play the video within the Playwright testing context, the test should be removed for now until a solution can be found in the future.

## Additional Details
This is the full stack trace of the error as seen by Github Actions:

```
  1) [chromium] › ocw-ci-test-course/video.spec.ts:40:5 › Transcripts start time matches video start time 

    Test timeout of 30000ms exceeded.

    Error: locator.getAttribute: Target closed
    =========================== logs ===========================
    waiting for locator('.video-page, .video-embed').locator('.transcript-line.is-active')
    ============================================================

      54 |   const activeCaption = await videoSection.transcript
      55 |     .activeLine()
    > 56 |     .getAttribute("data-begin")
         |      ^
      57 |   const nextCaption = await videoSection.transcript
      58 |     .nextLine()
      59 |     .getAttribute("data-begin")

        at /home/runner/work/ocw-hugo-themes/ocw-hugo-themes/tests-e2e/ocw-ci-test-course/video.spec.ts:56:6

    Pending operations:
      - locator.getAttribute at tests-e2e/ocw-ci-test-course/video.spec.ts:56:6

```

ChristopherChudzicki:
To see the issue more explicitly, I would run `yarn test:e2e video.spec --debug`. Navigate to http://localhost:3010/courses/ocw-ci-test-course/pages/video-series-overview/ ...

- In Chromium (launched by playwright), video will not play
- in other browsers, it will.

This seems like it must have been caused by an update to Chromium. Playwright's browsers are not pinned, I believe.

PR #1281, opened 2023-11-15
swap out chromium for google chrome and firefox in playwright tests

gumaerc:
Closes https://github.com/mitodl/ocw-hugo-themes/issues/1279

The "Transcripts start time matches video start time" test in `video.spec.ts` loads a video that started refusing to play in the Chromium browser, which Playwright uses on the backend to run its tests. This PR switches out the Chromium browser for Google Chrome and Firefox, which seem to run the test fine.

# How can this be tested?
 - Check out this branch locally
 - Make sure you don't have any values in your `.env` that would cause the end to end tests to fail. It may be best to back up your .env and run the tests with an empty one so it uses the defaults.
 - Run `yarn test:e2e` and verify that the tests pass

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -24,10 +24,14 @@ const config: PlaywrightTestConfig = {
   },
   projects: [
     {
-      name: "chromium",
+      name: "firefox",
       use:  {
-        ...devices["Desktop Chrome"]
+        ...devices["Desktop Firefox"]
       }
+    },
+    {
+      name: "Google Chrome",
+      use:  { ...devices["Desktop Chrome"], channel: "chrome" }
     }
   ],
   globalSetup: path.resolve(__dirname, "./tests-e2e/global-setup.ts")
--- a/test-sites/ocw-ci-test-course/content/resources/ocw_test_course_mit8_01f16_l01v03_360p.md
+++ b/test-sites/ocw-ci-test-course/content/resources/ocw_test_course_mit8_01f16_l01v03_360p.md
@@ -13,10 +13,12 @@ license: https://creativecommons.org/licenses/by-nc-sa/4.0/
 resourcetype: Video
 title: ocw_test_course_MIT8_01F16_L01v03_360p.mp4
 uid: d6502dcc-fdf4-462c-ae92-8855748843a8
+start_time: 13
+end_time: 50
 video_files:
-  video_captions_file: ''
+  video_captions_file: "/courses/123-ocw-ci-test-course-fall-2022/917263bef37857bd94ef67692405bcc9_erlp_sbca1s.vtt"
   video_thumbnail_file: https://img.youtube.com/vi/YDGGF2VWkAM/default.jpg
-  video_transcript_file: ''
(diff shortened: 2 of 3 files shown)


ibrahimjaved12 (commented):
It appears the test is flaky now - failing about 5%-10% of the times. Reporting the same error as earlier "target is closed".
Replace `test()` with `test.only()` on the 5th test
And then run:
`yarn test:e2e video.spec --repeat-each=50`

For this, I don't think we would be able to explore what's happening using `--debug`. I think this could be one of those cases where playwright tries to do stuff really fast before the page stuff has fully loaded? We have had cases like these before. With exactly the same percentage of flakiness.

We need to investigate / modify the test case

gumaerc:
@ibrahimjaved12 I reverted this PR to use the original video from the test and swapped out the browsers being used for testing, since Chromium is being problematic with videos.

ibrahimjaved12 (changes requested):
It looks like now we have 2 other tests failing, particularly in just `firefox`.
1. Verify that the 'Download video' and 'Download transcript' links are keyboard navigable and have the correct download URLs
```
Error: expect(received).toBe(expected) // Object.is equality

Expected: "https://live-qa.ocw.mit.edu/courses/123-ocw-ci-test-course-fall-2022/ocw_test_course_mit8_01f16_l01v01_360p_360p_16_9.mp4"
Received: null

  50 |     await page.keyboard.press("Tab")
  51 |     const hrefAttribute = await page.locator(":focus").getAttribute("href")
> 52 |     expect(hrefAttribute).toBe(downloadLinks[i])
     |                           ^
  53 |   }
  54 | })
  55 |

    at /Users/ibrahim.javed/MIT/ocw-hugo-themes/tests-e2e/ocw-ci-test-course/video-tabs.spec.ts:52:27
```
2. Video tabs content (links) are keyoard navigable
```
Test timeout of 30000ms exceeded.
Error: page.waitForURL: Navigation failed because page was closed!
=========================== logs ===========================
waiting for navigation to "**/courses/ocw-ci-test-course/resources/example_pdf/" until "load"
============================================================

  93 |     page.keyboard.press("Tab")
  94 |     page.keyboard.press("Enter")
> 95 |     await coursePage.page.waitForURL(`**/${tab.url}`)
     |                           ^
  96 |     expect(coursePage.page.url()).toContain(tab.url)
  97 |   }
  98 | })

    at /Users/ibrahim.javed/MIT/ocw-hugo-themes/tests-e2e/ocw-ci-test-course/video-tabs.spec.ts:95:27
```

Also, it appears the tests on Chrome run almost twice as fast as Firefox.
Despite that, I think it's useful that we are using both the browsers to run tests because firefox has actually pointed out good issues. (?)

Both the failed tests are pertaining to a lack of accessibility in those areas. I tested on my local firefox browser and I can confirm the behavior is same there.
Using tab on the `Download Video` button shifts focus to the next video tab -- Related resources, instead of "Download Transcript"
image

The other test expects (PDF link) to be clickable, but it seems that in Mac OS, using firefox leads to this problem -- that anchor tags are not keyboard clickable by default. Maybe the other problem is also related to that. There are fixes to this online -- that you can change your Mac settings / (or maybe firefox settings) but I don't see how that could be implemented for pw browser.

Do these problems occur in your linux firefox browser?

ibrahimjaved12 (approved):

PR #1298, opened 2024-01-18
feat: update shortcodes and partials for external-resource content

HussainTaj-arbisoft:
Closes https://github.com/mitodl/hq/issues/3405

This PR:
- Adds an `external_resource_link` shortcode.
- Updates `resource_link` shortcode to uniquely handle `external-resource` type content.
- Updates left nav to uniquely render `external-resource` content (`google.com`) instead of rendering a link to the content page (i.e. `/links/google`).

We will keep the `external_resource_link` shortcode or `resource_link` changes. It depends on what we decide in https://github.com/mitodl/hq/issues/3416. 

If a new UI element (CKEditor plugin) is to be created, we'll go with the `external_resource_link` shortcode, making things less complicated.

https://github.com/mitodl/ocw-hugo-themes/assets/71316217/0bc23776-6cb3-4d4a-a14e-32cffa55ec3e

#### Prerequisites

- Please go through https://github.com/mitodl/ocw-hugo-projects/pull/271 for context.

#### Steps

1. Grab [this content](https://github.com/hussaintaj-mitodl/test90).
2. Place the content inside your `ocw-content` directory.
3. Navigate to your `ocw-hugo-themes` directory.
4. Switch to branch `hussaintaj/3358-link-shortcode`. 
5. Run
    ```
    yarn start course test90
    ```
6. Open [HTTP://localhost:3000/pages/page](localhost:3000/pages/page).
7. Expect to see content similar to what is shown in the screen recording above.
8. Expect all links to work as expected, except the one named "broken".
9. Expect broken links to redirect you to the backup URL (see content md file for actual and backup URLs).

### Additional Context
https://github.com/mitodl/hq/issues/3358

### Checklist

Before merging this change:

- [x] Add e2e tests


base-theme/layouts/partials/external-link.html
```diff
@@ -0,0 +1,13 @@
+{{- $text := default .title .text | htmlEscape | .context.RenderString -}}
+{{- $isImage := default false .is_image -}}
+{{- $href := default "" .external_url -}}
+
+{{- if and .is_broken .backup_url -}}
```
gumaerc, line 5: Can we assign local variables for these and have defaults, like the other values above? Something like:

```
{{- $isBroken = default false .is_broken -}}
{{- $backupUrl default "" .backup_url -}}
```

base-theme/layouts/shortcodes/resource_link.html
```diff
@@ -2,5 +2,16 @@
 {{- $title := index .Params 1 | safeHTML -}}
 {{- $anchor_id := index .Params 2 | default ""}}
 {{- range where $.Site.Pages "Params.uid" $uuid -}}
+{{- if eq .Params.content_type "link" -}}
+    {{- partial "external-link"
```
gumaerc, line 6: This is a bit of a nitpick, but can we use underscores instead of hyphens in the file names for partials and shortcodes? I realize there are some partials that use hyphens, but we should be consistent and decide on one. I believe that we should use underscores.

I think hyphens are technically okay in partials because you call them using a string with the filename, but there are places in Hugo where hyphens are problematic. For example it's technically legal to use hyphens in YAML property names in front matter, but then you can call them from template code. Hugo will throw an error if you try and do something like `{{ .Params.hyphenated-property }}`. Underscores work fine though. I don't think there are any places (at least currently) where we would access a partial or shortcode as a direct property of an object like you do with front matter params, but it's possible something like that might exist in the future. Plus, I think it's easier to just have one formatting style when it comes to these things, to maintain consistency. I realize this is an opinion though, so feel free to ignore it or raise any counterpoints.

base-theme/layouts/partials/external-link.html
```diff
@@ -0,0 +1,13 @@
...
+  {{- $href = .backup_url -}}
+{{- end -}}
+
+{{- if $isImage -}}
+  <img src="{{- $href -}}" alt="{{- $text -}}" />
```
gumaerc, line 10: So, for resources we have `resource.html` and `resource_link.html`. The `resource.html` shortcode should really probably be called `resource_embed.html`, as its purpose is to take `resource` type content and embed it directly inline, whether that be a video, image or PDF. This is the difference between the "Add Link" and "Embed Resource" buttons in `ocw-studio`. The former will use the `resource_link.html` shortcode, and the latter will use `resource.html`. For images at least, I feel like we should differentiate between the shortcodes used to embed vs generate a link. Users might want to embed an external image, but they also might want to simply post a link to it. Embedding an external image also seems kind of risky, as the content of that image could change out from under us and either break or be a different image than the author intended. Whether or not to allow embedding of external images is a separate discussion, but I think we should at least create separate shortcodes for embedding external links vs simply creating `a` tags for them.

gumaerc (commented):
I tried this out with your sample content and everything seems to work as intended, but I have some comments and suggestions:

base-theme/layouts/shortcodes/resource_link.html
HussainTaj-arbisoft: This makes perfect sense to me and I agree with keeping it consistent. Thank you for sharing the rationale behind choosing underscores.

base-theme/layouts/partials/external-link.html
HussainTaj-arbisoft: Thank you for explaining the difference. It makes sense.

External image embed was added for legacy resources. In [this recent review](https://github.com/mitodl/ocw-hugo-projects/pull/271#discussion_r1466755861) we've decided to not support image embeds at all.

HussainTaj-arbisoft:
@gumaerc this PR is ready for another look.

The most significant changes since you last saw this PR are:

- Renamed `link` to `external-resource`.
- Removed `is_image` property.
- Added e2e tests and associated data.
- Used `menuItem.pageRef` instead of `menuItem.url`. You may see the relevant docs [here](https://gohugo.io/content-management/menus/#properties-site-configuration).

Right now we're going with the default UI. There is a potential for a new UI in https://github.com/mitodl/hq/issues/3416 but no solid decision has been made yet. I haven't removed the new `external_resource_link` shortcode from this PR in anticipation of a new UI component, but let me know if you think we should remove it until it's needed.

I've added some new data for the tests. https://github.com/mitodl/ocw-studio/pull/2116 updates the fixtures. It would be great if you could look at that too.

gumaerc (approved):
👍 This looks good to me. Good call using `pageRef`. In the past `pageRef` wasn't even documented functionality, and I'm pretty sure that this issue I made in the Hugo community forums pushed them to do so: https://discourse.gohugo.io/t/interaction-between-the-baseurl-and-relativeurls-settings/42770. Leaving the extra shortcode in shouldn't be a problem as long as there isn't anything using it. We should avoid leaving things in that will never get used, but it seems you are reasonably confident that it will be.

Overnight I'll be running a mass build locally against this branch to double check that every site builds properly.

PR #1301, opened 2024-01-25
fix non-credit filter

abeglova:
None.

Selecting 'Non-Credit' at  https://ocw.mit.edu/search/ does not work. This fixes the issue

We have parameter validation for the ocw search. Unfortunately, the parameter validation needs to be updated when we update the results from the api. 

### How can this be tested?
Set 
SEARCH_API_URL=https://open.mit.edu/api/v0/search/
Run `yarn start www`
Turn CORS off in your browser
Go to http://localhost:3000/search/

Verify that you can select non-credit

### Additional Context
Needing to manually update the possible parameters every time they change isn't great. 

Once ocw is migrated to the new search api and https://github.com/mitodl/mit-open/issues/269 is implemented we can get the possible parameters from the client that will be created by https://github.com/mitodl/mit-open/issues/269

### DIFF
--- a/www/assets/js/lib/constants.ts
+++ b/www/assets/js/lib/constants.ts
@@ -468,7 +468,7 @@ export const FACET_OPTIONS: Facets = {
   department_name: Object.values(departments).map(
     department => department.title
   ),
-  level:               ["Undergraduate", "Graduate", "Non Credit", "High School"],
+  level:               ["Undergraduate", "Graduate", "Non-Credit", "High School"],
   course_feature_tags: RESOURCE_TYPES,
   resource_type:       RESOURCE_TYPES
 }


gumaerc (approved):
:+1: LGTM

Although, it's worth noting that I had to set some other env variables locally to get this to work. `STATIC_API_BASE_URL`, `RESOURCE_BASE_URL` and `SITEMAP_DOMAIN` need to be set to prod (`https://ocw.mit.edu`) if you are hitting prod search. Without this, the page would simply refresh upon selecting `Non-Credit` and the facet would not be selected. I'm not sure which of the vars made the difference, likely either `STATIC_API_BASE_URL` or `RESOURCE_BASE_URL`.

ibrahimjaved12:
Thanks for fixing this! @abeglova
I had changed `Non Credit` to `Non-Credit` in `ocw-hugo-projects`, but I wasn't aware that there was a coupling of this in `ocw-hugo-themes`.

PR #1309, opened 2024-02-06
Update OCW logos and improve design alignments

ibrahimjaved12:
https://github.com/mitodl/hq/issues/3385

Updates OCW Logos and aligns the designs consistently.
Eg.
Header, footer, and search bar on homepage are now aligned.
Although not everything is aligned the same across both themes, but things (mostly header, footer, their `max-width`s) are now more consistent within the themes.

### How can this be tested?
1. Follow the discussions on the ticket / Look at figma designs from there
2. Look at different kinds of pages, different themes, apart from logos. To ensure everything is working correctly (because this isn't only a logo PR)
3. Make sure to check for different screen sizes.

For www online theme:
`yarn start www`

For www offline theme:
`yarn build ~/path/ocw-content-rc/ocw-www/ ~/path/ocw-hugo-projects/ocw-www/config-offline.yaml`
(then go to ocw-www content directory, and`dist/index.html`)

For course theme:
`yarn start course`

For offline course theme:
`yarn build ~/path/ocw-content-rc/9.40-spring-2018/ ~/path/ocw-hugo-projects/ocw-course-v2/config-offline.yaml`
(Then go to course content directory, and `dist/index.html`
Although do test on a different course than this one

Note:
You will see:
```
      <img
        src="/static_shared/images/mit_ol.svg"
        width="140px"
        alt="MIT Open Learning"
      />
```
in `Footer.tsx`
Webpack currently creates hashed version of every static image asset. This means, `mit_ol.svg` will not exist. And here, in a `.tsx` file, we cannot get our mapping too. The only reason why our [Search page](https://ocw.mit.edu/search/?q=) footer logo is currently working is because the image already existed from a pre-hash build.

Any way, as a fix to this logo ticket, what I have done is uploaded the `mit_ol.svg` logo to our AWS buckets (all OCW content buckets for RC and Prod). We can work on this problem in another ticket (probably would need a webpack rule I assume for its fix).
So we have the logo available here:
https://ocw.mit.edu/static_shared/images/mit_ol.svg
https://live-qa.ocw.mit.edu/static_shared/images/mit_ol.svg
So it should work in other environments.

To test locally, just place the `mit_ol.svg` in `static_shared` directory with the same name, it should work.

### WWW homepage:
image
image

### Search page
image

### Course page
image
image


base-theme/assets/css/main.scss
```diff
@@ -54,27 +54,6 @@ h1 {
...
-    font-family: Helvetica, sans-serif;
-    font-weight: 300;
-
-    span {
-      color: $orange;
-      font-weight: 600;
-    }
-  }
-}
-
```
ibrahimjaved12, line 77: We're not using` class="h2"` anywhere in the code base now. The reason why I was compelled to remove this particular chunk of code was the use of `color: $orange`, since that was the old theme. We're not supposed to be using` $orange` as color anymore.

The other scss properties for other heading tags may also be due a removal/refactor since we're supposed to use semantic heading tags -- and I think we did refactored those classes a long time ago to use semantic headings. But the other scss properties (for removal or refactoring) is relatively more code which would need more testing. Leaving this as it is for now

www/layouts/home.html
```diff
@@ -62,7 +62,7 @@
         </div>
       </div>
     </div>
-    <div class="unlock d-flex flex-column py-4 px-4 px-sm-7 m-2">
+    <div class="unlock d-flex flex-column py-4 px-4 px-sm-7 m-2 mr-xl-7">
```
ibrahimjaved12, line 65: "7" is 4 rem

www/layouts/home.html
```diff
@@ -15,7 +15,7 @@
   {{ end }}
 </div>
 <div class="d-flex flex-column align-items-center home-banner" style="background-image: url(/images/homepage_hero.jpg);">
-  <div class="banner-contents-container mx-auto d-flex justify-content-start justify-content-md-around w-100 align-items-center h-100">
+  <div class="banner-contents-container d-flex justify-content-between w-100 align-items-center h-100">
```
ibrahimjaved12, line 18: `mx-auto` was redundant here doing nothing.

www/layouts/home.html
```diff
@@ -15,7 +15,7 @@
   {{ end }}
 </div>
 <div class="d-flex flex-column align-items-center home-banner" style="background-image: url(/images/homepage_hero.jpg);">
-  <div class="banner-contents-container mx-auto d-flex justify-content-start justify-content-md-around w-100 align-items-center h-100">
```
ibrahimjaved12, line 18: `mx-auto` was redundant here achieving nothing.

base-theme/assets/css/main.scss
```diff
@@ -63,16 +63,6 @@ h2 {
...
-  &.branded {
-    font-family: Helvetica, sans-serif;
-    font-weight: 300;
-
-    span {
-      color: $orange;
-      font-weight: 600;
-    }
-  }
```
ibrahimjaved12, line 75: Not supposed to use orange theme anymore

base-theme/layouts/partials/footer.html
```diff
@@ -1,112 +1,114 @@
 <footer id="footer-container">
-  <div class="row pb-4 mx-0 justify-content-between">
-    <div>
-      <a
-        id="open-learning-logo"
-        href="https://openlearning.mit.edu/"
-        target="_blank"
+  <div id="home-footer">
```
ibrahimjaved12, line 2: There was a lot of confusing inconsistency in formatting here so I ran `prettier`

base-theme/layouts/partials/footer.html
```diff
@@ -1,112 +1,114 @@
 <footer id="footer-container">
-  <div class="row pb-4 mx-0 justify-content-between">
-    <div>
-      <a
-        id="open-learning-logo"
-        href="https://openlearning.mit.edu/"
-        target="_blank"
+  <div id="home-footer" class="w-100 mx-auto">
```
ibrahimjaved12, line 2: Ran `prettier` here because there was a lot of confusing inconsistent formatting.

Most of the code is same apart from a few differences such as image names, classes, etc.

pdpinch:
We need to update the favicon as well: https://brand.mit.edu/downloads?downloads_keywords=&field_asset_type%5B0%5D=5&field_asset_type%5B1%5D=5

gumaerc (commented):
As @pdpinch mentioned, if we are doing all of these changes in one PR we should include the new favicon from the branding package. Since there are a lot of styling changes in here, not all related to the new logo, I wanted to have @mbilalmughal look over a deployed version of the site before we merge this to be sure everything is looking correct. I looked over what was noted in the original issue with alignments and everything makes sense there for me, but I did notice one issue with the home page. The Chrome inspector's mobile device simulator only includes fairly modern devices. If you include an older device with a smaller screen, such as the iPhone 4, an issue reveals itself:

This branch:
image

Production:
image

It seems that the style breaks at device widths below 370 pixels. The iPhone 4 is obviously a very old device and it is unlikely that many people will be using a device with a resolution so small, but OCW is a bit of a special case in this regard. It is desired for OCW to have as much compatibility as possible with not only older browsers, but older mobile devices as well. It looks like the body and all the content we are putting on the page are properly respecting the 320 pixels wide resolution, and it's the Appzi feedback drawer that is breaking the layout now for some reason:

image

In production on mobile, the Appzi Feedback drawer actually seems to be flowing off the page and is not even accessible. You can see in the screenshot above that it is not there.

I started a full deploy of the OCW site using the mass build pipeline's `prefix` setting against this branch with the prefix `update-logos`. As of the time of writing, the deploy is not yet finished by hopefully by the time @ibrahimjaved12 and @mbilalmughal read this it should be ready. You can find the pages here:

Home page: https://live-qa.ocw.mit.edu/update-logos/
Search: https://live-qa.ocw.mit.edu/update-logos/search/
Example course: https://live-qa.ocw.mit.edu/update-logos/courses/18-05-introduction-to-probability-and-statistics-spring-2022/

The whole site should be deployed, but links currently do not contain the prefix. You can view any course page you want by browsing to the course or searching for it, then editing `/update-logos/` into the beginning of the URL path.

ibrahimjaved12:
Thanks for bringing this up @gumaerc. The white space on the updated homepage on the right-end side (coming from the drawer I think?), I do think it's in production as well currently? 

**Screen 1: Where chrome hides the rest of the screen in the far end side-- you only know when you try to drag it to left.**
image
image

**Screen 2: iphone 6/7/8**
image
image

**Screen 3: This is a bit tricky to get sometimes, but I think this is the closest behavior Chrome has to offer for iphone 4?**

image
image

I know from experience that Chrome's screens are not super accurate when it comes to iPhone layouts -- but here I don't know if it's that, because we would get a similar display on Chrome on any resolution similar to this size. It's kind of difficult to test this. As a general rule, whenever you change the layout, you should always refresh the page. But I was still getting mixed up results (mostly screen 1, but I think screen 3 is closest imitation). Maybe @mbilalmughal could shed some light on the extra space.

In past I have always used BrowserStack for iphone devices because of similar problems (btw, do DM me the MIT's account for this). But they only have until iphone 7.

I think for now we can ignore the white space as it's not coming out of this PR. but we should definitely confirm the other stuff.

ibrahimjaved12:
@gumaerc I have updated the PR with the following changes. It's ready to be reviewed

- Reduced blue box margin-right from 4rem to 3rem (on homepage)
- Updated favicon
- OCW Stories text instead of OpenCourseWare Stories (fixed iphone4 screen issue)
  - @pdpinch Currently we have `OCW News` text, and `OpenCourseWare Stories` text on homepage. The OCW text (abbreviation / full form) is inconsistent and the latter one is also causing problems for iphone 4 and smaller screens.
  - Which form should we keep? We can also implement the full term for bigger screens, and keep the OCW abbreviation only for smaller screens.
  - Currently I have adjusted it to `OCW Stories` (consistent with `OCW News`). Should I change it?
  - Can be seen here https://ocw-hugo-themes-www-pr-1309--ocw-next.netlify.app/ (may have to refresh cache for changes)
- Footer left-right padding from 3rem to 2rem for medium and smaller screens. For big screens still 3rem
- Paddings finished of search box for mobile, like before
- Slight padding adjustment of search bar in search page 
- Updated course footer (the padding is still 3rem, but there were additional margins and paddings coming from `row` and `col` classes in many places in addition - disabled those using `px-0` and `mx-0`

www/assets/css/search.scss
```diff
@@ -91,7 +91,13 @@
         padding: 0 3rem;
         @include breakpoint(tablet) {
           padding: 0;
+          margin-left: -15px !important;
```
ibrahimjaved12, line 94: Same as before; previously this was being applied through `row` class but since we want to use `mx-auto` for after `tablet`, so applying manually here

www/assets/css/search.scss
```diff
@@ -84,6 +84,17 @@
   .search-page {
     .container {
       max-width: $max-content-width;
+      @include breakpoint(phone) {
+        padding: 0 15px;
+      }
+      .row.search-filter-drawer {
+        padding: 0 3rem;
```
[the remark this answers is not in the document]
ibrahimjaved12: Good catch. Thanks for creating the pipelines off of my branch. I wasn't aware that my mass build would retrigger itself on every PR update? that's cool.

The intended reason for adding those margins was that the whole page would look self-contained, and that header/footer/search would be on the same alignments. But yes, for a mobile screen, where the header is even centered (not starting from left), we should readjust this.

I'm going to do the same behavior as before, for smaller screens. But for large screens, the paddings remain as it is.

ibrahimjaved12:
I did some tweaks on the search page to address that. In addition, I pushed the search bar a little right to match with the Courses container (like before).

Also, tweaked the filters page (On mobile view) for similar adjustments. The last change (search bar a little to right using `ml-lg-3`) is not up in the live-qa link yet but can be seen from the PR's deployment.

PR up for re-review

gumaerc (approved):
This looks good to me now. Here is a Browserstack Percy visual comparison of the changes here versus `main`: https://percy.io/b28d3eeb/ocw/builds/32682783. As long as @mbilalmughal and @pdpinch approve, I say we can move forward.

pdpinch:
I took a quick look at https://live-qa.ocw.mit.edu/update-logos/ and I'm wondering why the "Give Now" button in the page (not the nav) is wrapping:

image
 
mobile seems fine:

image

ibrahimjaved12:
> I took a quick look at https://live-qa.ocw.mit.edu/update-logos/ and I'm wondering why the "Give Now" button in the page (not the nav) is wrapping:
> 
> image
> mobile seems fine:
> 
> image

@pdpinch Please try refreshing your cache and/or opening from incognito. If the wrapping still occurs, please share your screen dimensions of the point where this behavior occurs (from inspect element). Also, please share which browser you used.

This was the earlier behavior of the button, and I had specifically added a `nowrap` rule to prevent wrapping.

gumaerc:
For what it's worth, this is what I see:

image

pdpinch:
after inspecting the element, it seems like I'm not getting the `nowrap` styling in Firefox or Safari, but I get it in Chrome. 

Chrome: 
image

Firefox: 
image

Not sure why, but Firefox says `nowrap` isn't valid. 

image

https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap?utm_source=devtools&utm_medium=inspector-css-compatibility&utm_campaign=default

ibrahimjaved12:
@pdpinch 

You're correct, it seems the support of `text-wrap: nowrap` (also `text-wrap` in general seems to be somewhat limited: https://caniuse.com/?search=text-wrap)

An alternative is `whitespace: nowrap`. It has all browsers support, except support for android (<= 4.3), so for older tablets (not smartphones) it would create the wrap. Using this with `flex: none` should cover all the support we need.
https://caniuse.com/?search=white-space
https://caniuse.com/?search=flex

Tested on: Safari, Chrome, Firefox, and Samsung A32.

@gumaerc 
I think testing mobile devices from Google Chrome's inspect element (on laptop) is essentially just setting particular height/width, the rest of the properties are used of the OS of the (laptop device), and version of the browser currently used.
So it's helpful but not very accurate in things like this, because some CSS properties are very browser-version or OS dependent. So in cases like this, `live.browserstack.com` is the way to go (also referenced by `caniuse.com`).

gumaerc:
@ibrahimjaved12 I've confirmed that the text wrapping is now working. I'm aware that browser based inspection tools only set width / pixel density and don't test rendering with the actual browser engine on any given device. My main goal here in nitpicking about the width is to get into a spot where given modern browser support, any screen size will look correct. I agree that the actual issues with older browsers are out of scope of this PR and we should open separate issues to investigate and address those. @pdpinch can clarify, but I think the strategy for OCW is to support as far back as possible without compromising maintainability or security. Anyway, I think we're good to merge this.

PR #1310, opened 2024-02-06
use git package; remove hardcoded baseurl

ChristopherChudzicki:
This PR makes a few small changes to `cg/add-mit-open-login-button`:

- Removes hard-coded MIT Open baseurl
- Installs `mit-open-login-button` from git rather than local.
    - Local might still be easier and useful for development, but git is definitely easier for sharing with others.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "lodash.uppercase": "^4.3.0",
     "mathjax": "^3.2.2",
     "mini-css-extract-plugin": "^2.6.1",
-    "mit-open-login-button": "/home/gumaerc/Code/mit-open-login-button",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -113,7 +113,7 @@
     "lodash.uppercase": "^4.3.0",
     "mathjax": "^3.2.2",
     "mini-css-extract-plugin": "^2.6.1",
-    "mit-open-login-button": "/home/gumaerc/Code/mit-open-login-button",
+    "mit-open-login-button": "https://github.com/mitodl/mit-open-login-button.git",
```
ChristopherChudzicki, line 116: Can also append `#branch-name-or-tag`

PR #1313, opened 2024-02-09
update search

### LINKED ISSUE #1294: Update the ocw  search page to use the new search apis
We changed the search apis on open to accept filter parameters instead of an open search query. This will be the api we will be using going forward. For now OCW is using the legacy search api that is separately running on open-discussions. We want to  update ocw to use the new apis so we no longer need to support the legacy search.

The bulk of the updates will happen in course-search-utils. Currently course-search-utils

1) Keeps track of search facet state
2) Keeps the parameters in the url of the search page in sync with the facet state
3) Generates the open search query

Now that we are moving to a new api we can get rid of (3) but (1) and (2) is still code that we want to share among all uis that use the search.

The work that needs to be done is

In course-search-utils:
1) Get rid of the large amount of code that generates the open search query and replace it with a small function that formats the active facets into get parameters
2) Rename the available facets to match the new parameters used by the api. For example the facet formerly known as `type` is now `resource_type` and the facet formerly called 'resource_type` is `content_feature_type`

In ocw-hugo-themes:
1) Update the code that displays facets to use use the new facet names
2) Update the code that converts search results to a LearningResource  object that can be used by course cards to use the new search result format
3) Update the code that actually makes the api call to make a get to the new endpoint instead of a post to the old endpoint

To make sure that everything works together the course-search-utils and ocw-hugo-themes work should happen silmultaneously

abeglova:
closes https://github.com/mitodl/ocw-hugo-themes/issues/1294

This pr updates the search to use the new open apis.

It uses the changes to course-search utils from https://github.com/mitodl/course-search-utils/pull/64

### How can this be tested?
Run mit-open

Set
COURSE_SEARCH_API_URL=http://localhost:8063/api/v1/learning_resources_search/
CONTENT_FILE_SEARCH_API_URL=http://localhost:8063/api/v1/content_file_search/

Run `yarn start www`

Go to http://localhost:3000/search/ and click around. Everything should work as expected. The urls will be a little different than the existing ui


www/assets/js/components/Facet.tsx
```diff
@@ -71,11 +81,4 @@ function SearchFacet(props: Props) {
   )
 }
 
-const propsAreEqual = (_prevProps: Props, nextProps: Props) => {
```
abeglova, line 74: We can't use this anymore since the aggregation is no longer nested in the bucket key - but the facets don't appear and disappear so I think the issue has been fixed elsewhere

www/assets/js/lib/constants.ts
```diff
@@ -464,11 +457,9 @@ export const FACET_OPTIONS: Facets = {
...
   ],
-  type:            ["resourcefile", "course"],
-  department_name: Object.values(departments).map(
-    department => department.title
-  ),
-  level:               ["Undergraduate", "Graduate", "Non-Credit", "High School"],
-  course_feature_tags: RESOURCE_TYPES,
-  resource_type:       RESOURCE_TYPES
+  resource_type:        ["course"],
+  department:           Object.keys(departments),
```
abeglova, line 461: We should have a followup pr to get all the options from the generated api documentation

www/assets/js/lib/search.ts
```diff
@@ -103,33 +108,60 @@ export const resourceJSONToLearningResource = (
   }
 }
 
-export const searchResultToLearningResource = (
-  result: LearningResourceResult
+export const courseSearchResultToLearningResource = (
```
abeglova, line 111: We should have a follow up pr to get rid of the LearningResource type defined in ocw-hugo-themes update the cards in `SearchResult` to expect the objects returned from the search api directly and update courseJSONToLearningResource and resourceJSONToLearningResource to convert ocw json to CourseResource and ContentFile from the open api.

ChristopherChudzicki (changes requested):
Overall, this is working very well. I've tested:
- filtering
- textual searches
- resource files & courses
- browser navigation (forward/back)

I still need to look at the actual course-search-utils changes.

### Changes

1. **`levelsMap`**: Right now `levelsMap` is constructed from `LearningResourcesListLevelEnum`. This products bad values for `high_school` and `noncredit`, namely "HighSchool" and "NonCredit" (no space / hyphen).
    - Maybe just use an explicit mapping like we do for departments?
2. **Course links:** In the course theme, there are links to the search page using old "Department" and "Level" values 
    - department appears here: https://ocw.mit.edu/courses/18-102-introduction-to-functional-analysis-spring-2021
    - level appears in the course-info box: https://ocw.mit.edu/courses/18-102-introduction-to-functional-analysis-spring-2021/video_galleries/lecture-videos/

    In slack we discussed making the search page with both the old and new department/level values, at least temporarily. I think that's reasonable to ease the eventual deployment.

### Further checking
Additionally, I think we should:
1. Set `CORS_ALLOWED_ORIGIN_REGEXES="ocw-hugo-themes-www-pr-\d+--ocw-next\.netlify"` on MIT Open Prod (if we agree that's OK)
2. Update the netlify workflow to use MIT Open Prod

That way we can share a version of this that uses the prod search for easier comparison with live OCW.

www/assets/js/components/SearchPage.tsx
```diff
@@ -16,74 +16,72 @@ import SearchFilterDrawer from "./SearchFilterDrawer"
...
   SEARCH_COMPACT_UI,
   DEFAULT_UI_PAGE_SIZE,
   COMPACT_UI_PAGE_SIZE,
-  OCW_PLATFORM
+  OCW_OFFEROR
 } from "../lib/constants"
 import { emptyOrNil, isDoubleQuoted } from "../lib/util"
-import { FacetManifest, LearningResourceResult } from "../LearningResources"
+import { FacetManifest } from "../LearningResources"
+import { CourseResource, ContentFile } from "../open_api_generated/api"
```
ChristopherChudzicki, line 32: Here and elsewhere that you use types only, I would do
```suggestion
import type { CourseResource, ContentFile } from "../open_api_generated/api"
```
which makes intent clear, and also 100% guarantees there is no impact on runtime.

ChristopherChudzicki:
Also: With the `levelsMap` change, I think you're only using types from the generated code. We could move that generated code to course-search-utils, and just export the relevant types from there. (With the goal that the generated could in course-search-utils should be replaced via `mitodl/open-api-cleints` soon.)

www/assets/js/components/SearchPage.tsx
```diff
@@ -199,14 +202,18 @@ export default function SearchPage(props: SearchPageProps) {
...
       })
+
+      if (nextResourceFilterState === CONTENT_FILE_ENDPOINT) {
+        updateSort(null)
+      }
       toggleFacets(toggledFacets)
+      updateEndpoint(nextResourceFilterState)
     },
-    [toggleFacets, activeFacets]
+    [toggleFacets, activeFacets, endpoint, updateEndpoint, sort, updateSort]
```
ChristopherChudzicki, line 212: I would remove the sort dependency like linter says.

ChristopherChudzicki (commented):
URLS like http://localhost:3000?d=Mechanical%20Engineering&d=Mathematics&l=High%20School are working great 👍 :

pdpinch:
@ChristopherChudzicki @abeglova what should we do with this PR? I'm not really clear on what it does.

pdpinch:
Is it worth keeping this PR open, or is it hopelessly out of date?

Issue #865, opened 2022-09-23
set a max width for the video player in course-v2 theme


pdpinch:
As a user, I need to be able to see the video controls when playing back an embedded video. It would be nice if I could see the full height of the transcript widget as well. 

#### Screencast demonstrating the issue

https://user-images.githubusercontent.com/430126/192046687-a784f244-e712-47e3-90d7-995e5f844a15.mov

#### Acceptance Criteria:

- [ ] Video should have a max width, whether or not the course info drawer is showing
- [ ] For a typical 16x9 video, the width should be limited so that the height allows for the controller and video transcript viewer on, for example, https://live-qa.ocw.mit.edu/course-v2/courses/18-02sc-multivariable-calculus-fall-2010/resources/changing-the-order-of-integration/

#### Out of Scope

- Not going to do

PR #898, opened 2022-10-07
style: max width for video player section

MAbdurrehman12:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [ ] Desktop screenshots
  - [ ] Mobile width screenshots
- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

#### What are the relevant tickets?
closes https://github.com/mitodl/ocw-hugo-themes/issues/865

#### What's this PR do?
- Sets max-width for video player (for both, embedded videos and video page)

#### How should this be manually tested?
- Checkout this branch or view netlify deployed version
- Build any course locally which has videos
- Verify that the video player section is not too big and fitted in your screen size with the transcript widget being visible.
- Verify that the video player section width remains same/does not change when you toggle (open/close) course info drawer.
- Verify these points for both, embedded videos and video pages.

#### Screenshots (if appropriate)

Uploading video player section max width.mov…

PR #1369, opened 2024-03-08
Video pages layout width-fix

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/2921

Removes `max-width: 100vh` from video pages styling to allow the content to spread over all of the `div` body.
This means, not only the text, but videos will now take up entire `div` width too.

### Screenshots
Refer to https://github.com/mitodl/ocw-hugo-themes/pull/1369#issuecomment-1993747002

### How can this be tested?
Checkout to this branch, check any course that has video lectures (the default course also does -- `yarn start course`). Check embed video pages as well. An example of embed video is in course `18.06sc-fall-2011` in the Instructor Insights page.

For offline course theme, run:
`yarn build ~/path/ocw-content-rc/18.06sc-fall-2011/ ~/path/ocw-hugo-projects/ocw-course-v2/config-offline.yaml`
And then open `index.html` from `dist` directory of the course, and navigate to videos. Replace course-id with course of your choice.

### DIFF
--- a/base-offline/layouts/partials/video.html
+++ b/base-offline/layouts/partials/video.html
@@ -15,13 +15,15 @@
   	<div class="description">
     	{{ .Content }}
   	</div>
-  	{{ partial "video_player.html" (dict "context" $context "youtubeKey" $youtubeKey "archiveUrl" $archiveUrl "downloadLink" $downloadLink "startTime" $startTime "endTime" $endTime) }}
+	<div class="video-player-wrapper">
+		{{ partial "video_player.html" (dict "context" $context "youtubeKey" $youtubeKey "archiveUrl" $archiveUrl "downloadLink" $downloadLink "startTime" $startTime "endTime" $endTime) }}
 
-  	{{ if $relatedResourses }}
-	  {{ partial "video_expandable_tab.html" (dict "context" $context "tabClass" "related-resource" "tabTitle" "Related Resources" "tabContent" $relatedResourses) }}
-  	{{ end }}
+		{{ if $relatedResourses }}
+		{{ partial "video_expandable_tab.html" (dict "context" $context "tabClass" "related-resource" "tabTitle" "Related Resources" "tabContent" $relatedResourses) }}
+		{{ end }}
 
-  	{{ if and $optionalTabTitle $optionalTabContent }}
-		{{ partial "video_expandable_tab.html" (dict "context" $context "tabClass" "optional-tab" "tabTitle" $optionalTabTitle "tabContent" $optionalTabContent) }}
-  	{{end}}
+		{{ if and $optionalTabTitle $optionalTabContent }}
+			{{ partial "video_expandable_tab.html" (dict "context" $context "tabClass" "optional-tab" "tabTitle" $optionalTabTitle "tabContent" $optionalTabContent) }}
+		{{end}}
+	</div>
 </div>
--- a/base-theme/assets/css/main.scss
+++ b/base-theme/assets/css/main.scss
@@ -342,8 +342,6 @@ article.content {
 }
 
 .course-section-title-container {
-  padding-right: 1.5rem;
-
   #course-title {
     border-bottom: 1px solid $medium-gray;
   }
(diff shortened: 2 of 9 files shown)


pdpinch:
Does the div have a maximum width? 

I know that on desktop views, we don't like to extend the text (and the video) too much. At really wide widths, it becomes hard to read (See, for example, the layouts of medium.com or nytimes.com)

ibrahimjaved12:
> Does the div have a maximum width?
> 
> I know that on desktop views, we don't like to extend the text (and the video) too much. At really wide widths, it becomes hard to read (See, for example, the layouts of medium.com or nytimes.com)

@pdpinch The webpage has dynamic max-width (fixed max widths, but different for different screens)
For desktop view, we have a `max-width` of 1447px for the entire webpage (which is centered)

From this, the video area `div` has `max-width` of 75% of the total width
image
Even if we close the right course info bar, then the video area takes its space, but only its space, not more than that

So our design is very similar to those websites
image
image


Previously we had this exact problem with the header and footers - that they would just spread around on left/right corners, for larger screens. But we fixed that recently.

ChristopherChudzicki:
@ibrahimjaved12: The `max-width: 100vh` was set in #898 to fix #865 ... Basically, the video playwer was getting too large and people couldn't see the controls.

From your screenshots, it looks like this change does NOT re-introduce #865 (good). But IMO worth double checking the reproduction steps in #865. **Note:** When #865 was reported, we were using a different video player, and subsequently switched to the YouTube player. So, assuming #865 has no resurfaced, that's one possible explanation for why.

ibrahimjaved12:
Kind of restrained by video player's entire transcript to be seen requirement, this is what I thought of.
Video player stays same as before, but we allow the textual content to take screen's allocated `max-width` (max-width of `xs`, `s`, `md`, `lg`, etc)

image

Basically you will see in the first pic in this screenshot, previously the text was same width as video player, but now its taking all the space it needs.

ibrahimjaved12:
This PR is again ready for review

pdpinch:
Ok, I think this is the best compromise we can get. It allows for the video and the transcript to both be in view. 

If the user really wants a full-wdith video, they can tap the full screen button.

gumaerc (approved):
👍 This looks fine to me. I find myself wondering if we could take a different approach to fullscreening the video though. According to https://developers.google.com/youtube/player_parameters, we can pass `fs=0` to disable the fullscreen button within the YouTube player. Then, in theory, we could provide our own fullscreen button. We could then use the `:fullscreen` pseudo class to define our own fullscreen layout. This layout could place the transcript on the right side of the video in landscape, and at the bottom of the video in portrait. In theory, anyway. I think that this could maybe be worth it, but it depends on how many people watch OCW videos in this manner. As @pdpinch mentioned, they can fullscreen the video and turn on closed captioning in YouTube which will provide the transcript. The only functionality this doesn't offer is scrolling through the transcript and clicking a line to seek there in the video. If we provided a custom fullscreen layout, then this would be possible.

ibrahimjaved12:
> 👍 This looks fine to me. I find myself wondering if we could take a different approach to fullscreening the video though. According to https://developers.google.com/youtube/player_parameters, we can pass `fs=0` to disable the fullscreen button within the YouTube player. Then, in theory, we could provide our own fullscreen button. We could then use the `:fullscreen` pseudo class to define our own fullscreen layout. This layout could place the transcript on the right side of the video in landscape, and at the bottom of the video in portrait. In theory, anyway. I think that this could maybe be worth it, but it depends on how many people watch OCW videos in this manner. As @pdpinch mentioned, they can fullscreen the video and turn on closed captioning in YouTube which will provide the transcript. The only functionality this doesn't offer is scrolling through the transcript and clicking a line to seek there in the video. If we provided a custom fullscreen layout, then this would be possible. For example, the Twitch video player provides similar functionality with their "theater mode" button which makes the video take up all available space in the browser window aside from a column on the right hand side for the chat. We could create a similar layout, but with a transcript instead of live stream chat.

Thanks @gumaerc. I like the idea. For now, I'm going to merge this PR. But I'll R&D / experiment on this and see how we could go with this.

PR #1373, opened 2024-03-20
Markdownify quiz choice

ibrahimjaved12:
Part of https://github.com/mitodl/hq/issues/3661

Adds `.RenderString` in `quiz_choice` to allow markdown rendering. The CSS changes were added to align everything centrally - the uneven spacing was specifically apparent with `code` blocks in MCQ options.

image
image
image

### How can this be tested?

Lecture 8 and Lecture 9 of `6.0001-fall-2016` have such contents... I'm not sure if they're available in `ocw-content-rc` right now, so you can also use this markdown:

```
### Getters and Setters

1. Which of the below is a getter method for the number of wheels?    

    ```plaintext
    ----------------------------------
    ----------- Given ------------
    ----------------------------------
    class Car(object):
        def __init__(self, w, d):
            self.wheels = w
            self.doors = d
            self.color = ""
    ----------------------------------

    (A)     def get_wheels():
                return wheels

    (B)     def get_wheels():
                return self.wheels

    (C)     def get_wheels(self):
                return wheels

    (D)     def get_wheels(self):
                return self.wheels
    ```
      
    {{< quiz_multiple_choice questionId="Q1_div" >}}{{< quiz_choices >}}{{< quiz_choice isCorrect="false" >}}  
    def get_wheels():    
    {{< /quiz_choice >}}  
    {{< quiz_choice isCorrect="false" >}}  
    def get_wheels():  
    {{< /quiz_choice >}}  
    {{< quiz_choice isCorrect="false" >}}  
    def get_wheels(self):    
    {{< /quiz_choice >}}  
    {{< quiz_choice isCorrect="true" >}}  
    def get_wheels(self):    
    {{< /quiz_choice >}}{{< /quiz_choices >}}  
    {{< quiz_solution >}}  
      
     

{{< resource uuid="c4e79feb-d329-f620-2a0b-250ea3c746aa" >}}
```

### DIFF
--- a/course-v2/layouts/partials/quiz_multiple_choice_choice.html
+++ b/course-v2/layouts/partials/quiz_multiple_choice_choice.html
@@ -1,16 +1,16 @@
 <div class="multiple-choice-div">
 	<label class="d-flex align-items-center">
 		<input type='radio' name={{ .questionId }} class="multiple-choice-radio">
-		{{ .choiceText | markdownify }}
-		
-		{{ if eq .correct "true" }}
-		<span class="toggle material-icons correctness-icon correctness-icon-correct">
-			check
-		</span>
-		{{ else }}
-		<span class="toggle material-icons correctness-icon correctness-icon-wrong">
-			close
-		</span>
-		{{ end }}
+			{{ .choiceText | markdownify }}
+			
+			{{ if eq .correct "true" }}
+				<span class="toggle material-icons correctness-icon correctness-icon-correct">
+					check
+				</span>
+			{{ else }}
+				<span class="toggle material-icons correctness-icon correctness-icon-wrong">
+					close
+				</span>
+			{{ end }}
 	</label>
 </div>
--- a/course-v2/layouts/partials/quiz_multiple_choice_choice.html
+++ b/course-v2/layouts/partials/quiz_multiple_choice_choice.html
@@ -1,16 +1,16 @@
 <div class="multiple-choice-div">
 	<label class="d-flex align-items-center">
 		<input type='radio' name={{ .questionId }} class="multiple-choice-radio">
-			{{ .choiceText | markdownify }}
-			
-			{{ if eq .correct "true" }}
-				<span class="toggle material-icons correctness-icon correctness-icon-correct">
-					check
-				</span>
-			{{ else }}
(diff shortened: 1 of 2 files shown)


course-v2/layouts/partials/quiz_multiple_choice_choice.html
```diff
@@ -1,14 +1,16 @@
 <div class="multiple-choice-div">
-	<label>
-		<input type='radio' name={{ .questionId }} class="multiple-choice-radio">{{.choiceText}}
+	<label class="d-flex align-items-center">
+		<input type='radio' name={{ .questionId }} class="multiple-choice-radio">
+		{{ .choiceText | markdownify }}
```
gumaerc, line 4: Why not use `.RenderString` here? The `markdownify` function is mostly still part of Hugo as a compatibility measure for older sites, and ultimately it calls `site.Home.RenderString` anyway. The `.RenderString` function is part of the `Page` context, so it does need to be called from a page, but in Hugo 0.111.0 the global `page` func was added. Directly on a page, you can call `.RenderString` because the context (`.`) *is* a `Page`. Here in this partial there is no passed in `Page` context, but you can call the new global `page` function to get that context, and then `.RenderString` will always be called in the context of the current `Page` being rendered. With `markdownify`, it's always run in the context of the "home" page. In this case I don't think that's a problem, but in the Hugo forum thread I posted above people mostly recommend using `.RenderString` wherever you can over `markdownify`. A lot of the code here in `ocw-hugo-themes` uses `markdownify` because a majority of it was written before `.RenderString` was even available. I think going forward, we should use `.RenderString` instead of `markdownify` wherever possible and potentially refactor old code to use it as well.

gumaerc (changes requested):
I was able to render the quiz with the appropriate indentation using this branch. However, I do have one suggestion:

course-v2/layouts/partials/quiz_multiple_choice_choice.html
ibrahimjaved12: Thanks. In the back of my mind I had a feeling I was missing something - I was actually trying `.RenderString` some days back for Course Collection. But I looked at the codebase, and `markdownify` was being used pretty much everywhere, so I decided to use it.

I have updated the PR with the changes

gumaerc (approved):
👍 LGTM!

Issue #1340, opened 2024-02-16
Perform Percy snapshotting in end to end tests


gumaerc:
### Description/Context
[Browserstack Percy](https://www.browserstack.com/docs/percy/overview/basics) is a product that takes snapshots of websites and performs automatic visual comparison, generating images that show the differences between two snapshots. We should add this to our end to end tests so visual comparisons can be made both from branch to branch and also from commit to commit if necessary.

### Plan/Design
 - [ ] Add the `@percy/cli` and `@percy/playwright` to dev dependencies
 - [ ] Add Percy snapshots to the appropriate tests:
   - [ ] The main `ocw-www` home page
   - [ ] The search page
   - [ ] The course home page
   - [ ] A course content page
 - [ ] Make sure the proper env variables are set up in RC / production before merging PR's related to this

gumaerc:
Here's some code that I wrote to do some experimental testing with Percy. This just takes a snapshot of the course home page:

```js
import playwright from "playwright"
import { env } from "../../env"
import { test, expect } from "@playwright/test"
import { getFirstAfter, CoursePage } from "../util"
const percySnapshot = require('@percy/playwright')

const browsers = {
  "firefox": playwright.firefox,
  "chromium": playwright.chromium
}

test("Loads the course home page", async ({ browserName }) => {
  const browser = await browsers[browserName].launch()
  const page = await browser.newPage()
  const course = new CoursePage(page, "course")
  await course.goto()
  await percySnapshot(page, `Loads the course home page in ${browserName}`)
  await browser.close()
})
```

One peculiarity that I found while experimenting with this is that you need to run the `browser.close()` function in order for Percy to consider the session "finished" and take the snapshot, otherwise it will wait indefinitely. All of our tests only use the `page` fixture, which doesn't include the `browser` object. If I used the `browser` fixture and run `browser.close()` on that, I would get an error from the next test that "the browser has been closed" even though each Playwright test is supposed to spin up its own browser window and close it at the end. So, I found that the best way to do this if you want a Percy snapshot in the test is to explicitly create a `browser` in the manner above. Also, we would need to create a new `package.json` script that runs the e2e tests with Percy:

```json
"test:e2e:percy": "yarn with-env --dev -- percy exec -- ./package_scripts/test_e2e.sh",
```

pdpinch:
If we had added the tests as described here, what would have happened in a case like #1379?

HussainTaj-arbisoft:
[...]

Issue #1271, opened 2023-10-24
Use a single source for test sites' content


HussainTaj-arbisoft:
# Description/Context

Our Playwright tests use a set of test sites. These test sites are located on RC as well as inside the `ocw-hugo-themes` repo, as markdown content. From what we see on RC, we were meant to keep these manually synced.

I believe that changes to the test sites should always be made on RC first. We should remove the need for a manual step, which I believe is the cause of this inconsistency. I propose that we use the GitHub content, created via Studio RC, as the source for the test sites in `ocw-hugo-themes`.

# Plan/Design

To achieve this objective, we will need to perform the following tasks.

- [ ] Sync/update the RC site with the content we currently have in the themes repo.
- [ ] Remove test sites from the themes repo and ignore them in source control.
- [ ] Update the `test:e2e` command to automatically pull the test sites when needed.
- [ ] Update GitHub actions to set up the appropriate secret required to pull the repos.

I don't know why we initially chose to duplicate the content. I might have missed something, so whoever works on this might need to refine the plan accordingly.

HussainTaj-arbisoft:
I just realized that the reason why we might duplicate content is to avoid conflicts when two different engineers work on the same set of tests.

In that case, what if instead of storing the entire content we only store the commit hash?

HussainTaj-arbisoft:
We had a short discussion on this.

The primary challenge with the proposed approach is the occasional DB sync with production. These syncs will end up destroying the studio DB and these test courses.

The current favored approach is to use the `themes` repo's test site content as the source, instead of the content repo's. To keep them in sync, it is proposed that whenever a change in the `themes` repo happens, we update the content repos via some workflow.

PR #1379, opened 2024-03-28
fix: expand description button style

HussainTaj-arbisoft:
Related to https://github.com/mitodl/ocw-hugo-themes/pull/1371 and https://github.com/mitodl/hq/issues/3858

In https://github.com/mitodl/ocw-hugo-themes/pull/1371 we fixed a markdown rendering issue. The fix changes the HTML layout of the `collapsed-description` div.

We used to have this layout
```html
<div>
  text text text <button></button>
</div>
```

The fix changes it to 
```html
<div>
  <p>text text text</p>
  <button></button>
</div>
```

This introduces a line break that changes how the "See more" button is placed.

Screenshot 2024-03-28 at 12 00 22 PM

> What it should look like

Screenshot 2024-03-28 at 12 00 41 PM

> What it looks like on RC right now

Screenshot 2024-03-28 at 12 04 25 PM

### How can this be tested?

1. Grab contents of [this course](https://github.mit.edu/ocw-content-rc/sts.050-spring-2016).
2. Navigate to your local `ocw-hugo-themes` setup.
3. Run
    ```
    yarn start course sts.050-spring-2016
    ```
4. Open http://localhost:3000/
5. Observe the homepage.
6. Expect to see the description (collapses and expanded) rendered correctly. Please see the screenshot for reference.


pdpinch:
Is there a way to add a test to this? Is a visual test necessary (like Percy) or can this be done with playwright?

HussainTaj-arbisoft:
> Is there a way to add a test to this? Is a visual test necessary (like Percy) or can this be done with playwright?

Yes, Playwright has [visual comparison](https://playwright.dev/docs/test-snapshots) tests that we can use. I'll write a quick test.

I think Percy would be better but I don't have any experience using it. @pdpinch, do we have Percy set up for this project?

HussainTaj-arbisoft:
@ibrahimjaved12, I'm going to try adding some visual tests. I'll let you know when it's ready for review again.

HussainTaj-arbisoft:
@ibrahimjaved12, @pdpinch, I experimented with adding Playwright visual tests. We require a more thorough plan to use visual tests effectively.

To create reliable visual tests, we need consistent environments. Playwright tests are run on three environments local, GitHub Actions, and Concourse, and we always use the latest browsers for these tests. Under these conditions, visual tests are not robust and tend to fail in one environment or the other.

@pdpinch, should I write an issue to set up visual testing for this project?

I read the Percy starter docs and I think Percy would be easier to integrate and maintain than playwright visual tests. 

---

For now, I have added a test to validate CSS for this change (though this is not as reliable as a visual test would be).

gumaerc (approved):
👍 LGTM

@HussainTaj-arbisoft  FYI, I created an issue already to integrate Percy into our testing: https://github.com/mitodl/ocw-hugo-themes/issues/1340. Percy works in conjunction with an end to end testing framework like Playwright, so we can write Playwright tests and send snapshots to Percy in whichever tests we decide need visual testing.

> To create reliable visual tests, we need consistent environments. Playwright tests are run on three environments local, GitHub Actions, and Concourse, and we always use the latest browsers for these tests. Under these conditions, visual tests are not robust and tend to fail in one environment or the other.

Currently, the end to end tests run in Concourse are run against sites generated from different content then that of Github actions or local development. In the past we have discussed moving to a single source for end to end test content: https://github.com/mitodl/ocw-hugo-themes/issues/1271. Even so, differences in environments could possibly end up making the results of the build different, so I see what you're saying here. However, when you are creating a Percy snapshot you can pass it a string to identify that snapshot. If we passed some kind of environment string here, then the snapshots would be unique per environment. Likely we would want to have Percy snapshotting feature flagged though, as the free account only allows you 5000 snapshots per month. If we were performing this every time a Github actions run is done as well as running them in Concourse, we would probably run out of snapshots pretty quickly. If we limited visual testing to the Concourse pipeline, then the free amount of snapshots would probably be fine.

@pdpinch For what it's worth, when I have performed Percy snapshot comparisons in the past I set that up manually following the process in https://github.com/mitodl/ocw-hugo-themes/issues/1340. In order to add visual tests for new functionality as a part of our workflow, we would need to add Percy snapshotting to the applicable tests like I describe in that issue. We would likely want to feature flag it and only put snapshots on some of the tests, ideally things like the home page, etc. If we decide that visual testing is valuable and want to pay for more snapshotting, then we could consider expanding it to more tests at that point.

PR #1383, opened 2024-04-03
Initial commit

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/3877

Adds `videojs` to `ocw-www` theme

- [ ] Desktop screenshots


ChristopherChudzicki:
@ibrahimjaved12 In regards to this weird breakage:

Screenshot 2024-04-08 at 2 54 55 PM

I looked into this for a while today. I finally realized this has to do with Webpack's  [SplitChunksPlugin](https://webpack.js.org/plugins/split-chunks-plugin/). We have the plugin configured like this:
```ts
    splitChunks: {
      minChunks:   2,
      cacheGroups: {
        common: {
          test:   /[\\/]node_modules[\\/]/,
          name:   "common",
          chunks: chunk => {
            const splitChunks = [entryNames.www, entryNames.courseV2]
            return splitChunks.includes(chunk.name)
          }
        }
      }
    }
```
The idea here is:
- Create an output called `common`. If a file satisfies condition `chunks`, then it goes in the "common" chunk. In particular, if something is used by BOTH `www` and `course-v2`, then it goes in `common`, and omitted from those two outputs.
- Consequently, the `www` theme needs to load both `www + common` assets, and `course-v2` needs to load both `course_v2 + common` assets.

Which we've been doing. For javascript.

But not for CSS. And that had been OK. There wasn't any CSS in the `common` output.

But now BOTH `www` and `course-v2` are requesting ` "video.js/dist/video-js.css"`. So now that goes into the `common` output. So now we DO need to include the common CSS.

For example:
```diff
--- a/course-v2/layouts/partials/extrahead.html
+++ b/course-v2/layouts/partials/extrahead.html
@@ -2,6 +2,7 @@
 {{- define "extrahead" -}}
 {{- $css_urls := slice (partial "get_asset_webpack_url.html" "course_v2.css") -}}
 {{- $layout := .Params.layout -}}
+{{- $css_urls = $css_urls | append (partial "get_asset_webpack_url.html" "common.css") -}}
 {{- if eq $layout "instructor_insights" -}}
     {{- $css_urls = $css_urls | append (partial "get_asset_webpack_url.html" "instructor_insights.css") -}}
 {{- end -}}
```
This fixes course-v2. Probably something similar for www? You may know better than me.

blarghmatey:
I disabled the pre-commit CI check on this repo since it doesn't have a valid pre-commit config at the moment.

ibrahimjaved12:
> I disabled the pre-commit CI check on this repo since it doesn't have a valid pre-commit config at the moment.

Thanks @blarghmatey

HussainTaj-arbisoft (approved):
Looks good to me! 👍 

I'm assuming we're leaving the www video resource page UI adjustments for later, right? I think some improvements can be made there.

PS: You can push an empty commit to trigger the pre-commit.ci. The pre-commit issue should be fixed by now.

```
git commit --allow-empty -m "trigger ci"
```

Edit: oh, well nevermind. The check's not required.

PR #1407, opened 2024-05-30
Presenting page links in navigation menu that are on the `ocw.mit.edu` domain

umar8hassan:
closes #4189

This PR intends to update presentation of external resources in left navigation menu. The resources which are hosted on `ocw.mit.edu` domain should be considered internal links though added as external resources. 
Screenshot 2024-05-31 at 4 57 00 PM

These links should be opened in the same window and external link arrow should not be displayed for such resources.

### How can this be tested?
1. Checkout to `master` branch in `ocw-studio` (recommended)
2. Use `OCW_HUGO_THEMES_BRANCH=umar/4189-presenting-external-resource-with-same-domain` in `ocw-studio` .env file
3. Start/Restart `ocw-studio`. check your env inside container with the following optional step to validate that correct configurations are loaded:
    - (optional) `docker-compose exec web bash -c "env | grep OCW_HUGO_THEMES_BRANCH"`
4. Update pipelines:
    - docker-compose exec web ./manage.py upsert_theme_assets_pipeline --themes-branch umar/4189-presenting-external-resource-with-same-domain
5. Open concourse interface: `http://concourse:8080/`
    - Unpause `ocw-themes-assets` pipeline for the above branch and trigger a build for it. Wait for it to complete.
6. Log in to `cow-studio` and create a new course.
7. Add two external resources, one with `cow.mit.edu` domain and one without it.
8. Link the external resources in menu items.
9. Publish course and wait for it to complete.
10. The course when published should display and open Left Navigation items as such:
    - resources on `ocw.mit.edu` domain should open in the same tab and should not contain arrow for new tab
    - resources not on `ocw.mit.edu` domain should open in a new tab and should contain arrow for new tab

### DIFF
--- a/base-theme/layouts/partials/external_resource_link.html
+++ b/base-theme/layouts/partials/external_resource_link.html
@@ -13,11 +13,21 @@
 {{- if and $isBroken $backupUrl -}}
   {{- $href = $backupUrl -}}
 {{- end -}}
-{{- partial "link" (dict
+{{- if in $href "ocw.mit.edu" -}}
+  {{- partial "link" (dict
       "href" $href
       "text" $text
       "class" $className
       "onClick" $onClick
       "target" "_blank"
-  )
--}}
+    )
+  -}}
+{{- else -}}
+  {{- partial "link" (dict
+    "href" $href
+    "text" $text
+    "class" $className
+    "onClick" $onClick
+    )
+  -}}
+{{- end -}}
--- a/base-theme/layouts/partials/external_resource_link.html
+++ b/base-theme/layouts/partials/external_resource_link.html
@@ -13,7 +13,7 @@
 {{- if and $isBroken $backupUrl -}}
   {{- $href = $backupUrl -}}
 {{- end -}}
-{{- if in $href "ocw.mit.edu" -}}
+{{- if not (in $href "ocw.mit.edu") -}}
   {{- partial "link" (dict
       "href" $href
       "text" $text
@@ -26,7 +26,7 @@
   {{- partial "link" (dict
     "href" $href
     "text" $text
-    "class" $className
+    "class" "text-dark nav-link"
     "onClick" $onClick
     )
   -}}
--- a/base-theme/layouts/partials/external_resource_link.html
+++ b/base-theme/layouts/partials/external_resource_link.html
@@ -26,7 +26,7 @@
   {{- partial "link" (dict
     "href" $href
     "text" $text
-    "class" "text-dark nav-link"
+    "class" "external-link"
     "onClick" ""
     )
   -}}
--- a/base-theme/layouts/partials/external_resource_link.html
+++ b/base-theme/layouts/partials/external_resource_link.html
@@ -28,7 +28,6 @@
     "text" $text
     "class" "text-dark nav-link"
     "onClick" ""
-    "target" ""
     )
   -}}
 {{- end -}}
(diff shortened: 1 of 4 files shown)


pdpinch:
How can we test this? Not sure if this is the best approach, but I'd suggest adding an ocw.mit.edu link to the test course and a playright test to validate the behavior.

umar8hassan:
> How can we test this? Not sure if this is the best approach, but I'd suggest adding an ocw.mit.edu link to the test course and a playright test to validate the behavior.

That's exactly what I have done. There was some issue while testing due to which this got delayed. But it should be good now. Will push updates and finalize PR first thing in the morning.

base-theme/layouts/partials/external_resource_link.html
```diff
@@ -13,11 +13,21 @@
...
--}}
+    )
+  -}}
+{{- else -}}
+  {{- partial "link" (dict
+    "href" $href
+    "text" $text
+    "class" "text-dark nav-link"
+    "onClick" $onClick
+    )
```
ibrahimjaved12, line 31: I tried all possible combinations of: (OCW link, non OCW link, live-qa link, with non-OCW warning, without non-OCW warning)

It failed in the case of: (OCW link, with non-OCW warning)
Technically it does not make sense to have this use-case in production. However, since non-OCW warning is selected by default, I think we could have cases where Course Authors could accidentally create an OCW link with the warning - in which case the redirect fails to occur, and no warning pop up appears as well.

Current behavior: If you try this in RC, currently a warning appears and the page redirects to OCW link if you click "Continue".

Possible options:

1. Bring back the warning in this specific case (with option to redirect). 
2. Remove warning pop up (even if selected from ocw-studio) and redirect in same tab, if it is truly OCW resource (the behavior from this PR already attempts this. However it does not work because of `"onClick" $onClick` on line 30, as it calls `event.preventDefault()`. Remove line number 30 and it will work.)
3. Set a warning in `ocw-studio` with the same URL filter. Eg.` It looks like the external resource you added is actually OCW. Are you sure you want to enable non-OCW content warning?`

Currently in production we have 1), but I think 2) may be preferable. 
If we do 2), then 3) becomes less important and could be skipped.
If we do 1), then 3) becomes more important.

ibrahimjaved12 (changes requested):

base-theme/layouts/partials/external_resource_link.html
pdpinch: @umar8hassan @ibrahimjaved12 what did you decided to do in this case?

ibrahimjaved12 (approved):
Looks good to me 👍🏻 
However we should wait for @pdpinch's approval on the approach

ibrahimjaved12:
While testing external resource for `www` for https://github.com/mitodl/hq/issues/4245, I noticed a styling side effect coming from this PR.

We are using class `nav-link` here for external resources. It works for left navbar, because all the items there already use `nav-link` - it makes the design clean. However, in the case of `linking` external resources, (in `pages`, or elsewhere), `nav-link` will add: `padding: 0.5rem 1rem` which makes it a bit awkward.
In addition to this, the `color` of the font also changes from some shade of dark blue to black. 

I have also attached a screenshot of before for comparison.

image

image

1. Make sure the additional paddings/margin is gone - make it same as before (while making sure the left navbar items stay consistent too)
2. Make the font color same as before. Also, font color on anchor links can sometimes give a false positive. For example, they may appear blue and change color to black after clicking on the link. But if you always open in new incognito tab -- this may go unnoticed because it would be always blue. It may be the inherited `visited` CSS property.

I'm still approving this because functionality wise everything is working, and the nature of this change is trivial. Make sure you do the fix before you merge. 👍🏻

PR #1414, opened 2024-06-20
Presenting external resource with same domain

umar8hassan:
closes https://github.com/mitodl/hq/issues/4189

This PR intends to update presentation of external resources in left navigation menu. The resources which are hosted on `ocw.mit.edu` domain should be considered internal links though added as external resources. 
Screenshot 2024-05-31 at 4 57 00 PM

These links should be opened in the same window and external link arrow should not be displayed for such resources.

### How can this be tested?
1. Checkout to `master` branch in `ocw-studio` (recommended)
2. Use `OCW_HUGO_THEMES_BRANCH=umar/4189-presenting-external-resource-with-same-domain` in `ocw-studio` .env file
3. Start/Restart `ocw-studio`. check your env inside container with the following optional step to validate that correct configurations are loaded:
    - (optional) `docker-compose exec web bash -c "env | grep OCW_HUGO_THEMES_BRANCH"`
4. Update pipelines:
    - docker-compose exec web ./manage.py upsert_theme_assets_pipeline --themes-branch umar/4189-presenting-external-resource-with-same-domain
5. Open concourse interface: `http://concourse:8080/`
    - Unpause `ocw-themes-assets` pipeline for the above branch and trigger a build for it. Wait for it to complete.
6. Log in to `cow-studio` and create a new course.
7. Add two external resources, one with `cow.mit.edu` domain and one without it.
8. Link the external resources in menu items.
9. Publish course and wait for it to complete.
10. The course when published should display and open Left Navigation items as such:
    - resources on `ocw.mit.edu` domain should open in the same tab and should not contain arrow for new tab
    - resources not on `ocw.mit.edu` domain should open in a new tab and should contain arrow for new tab

### DIFF
--- a/base-theme/layouts/partials/external_resource_link.html
+++ b/base-theme/layouts/partials/external_resource_link.html
@@ -13,11 +13,21 @@
 {{- if and $isBroken $backupUrl -}}
   {{- $href = $backupUrl -}}
 {{- end -}}
-{{- partial "link" (dict
+{{- if in $href "ocw.mit.edu" -}}
+  {{- partial "link" (dict
       "href" $href
       "text" $text
       "class" $className
       "onClick" $onClick
       "target" "_blank"
-  )
--}}
+    )
+  -}}
+{{- else -}}
+  {{- partial "link" (dict
+    "href" $href
+    "text" $text
+    "class" $className
+    "onClick" $onClick
+    )
+  -}}
+{{- end -}}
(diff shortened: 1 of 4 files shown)


ibrahimjaved12 (changes requested):
There's a slight problem here. `text-dark` class was probably originally only added to force the links in the left navbar to appear dark (different from the blue shade we get for anchor links otherwise).

I looked at the ticket to confirm. We probably want the linked external-resources to all act consistent color-wise.

image

Also, when you do this, could you please ensure the solution for this is in base-theme. (just like it currently is). Then we could apply these styling changes in `www` theme as well.

ibrahimjaved12 (approved):
👍🏻

PR #1432, opened 2024-08-02
Update OCW open graph preview logo

ibrahimjaved12:
### What are the relevant tickets?
No ticket

### Description (What does it do?)
Currently `open graph preview` for OCW site shows old logo. This PR updates to using the new logo, and also removes the old logos from assets.

### Screenshots (if appropriate):
image

This is the preview image we are going to use now:
image

### How can this be tested?
It would be difficult to test it locally (would have to use `ngrok`). It's a simple change. Let's test it on RC.

PR #1435, opened 2024-08-06
Update OCW open graph preview logo (use .png)

ibrahimjaved12:
No ticket

I put up a fix here https://github.com/mitodl/ocw-hugo-themes/pull/1432 but it used `.svg` over `.png` which has compatibility issues. This PR uses same logo (although in slightly different aspect ratio), and using `.png` instead.

These are general guidelines for OG images:
1. Use images that are at least 1200 x 630 pixels for the best display on high resolution devices."
2. Try to keep your images as close to 1.91:1 aspect ratio as possible to display the full image in News Feed without any cropping.
3. The preferred width of an image is 1500px

This PR is made keeping in mind these requirements.

### How can this be tested?
It would be difficult to test it locally (would have to use `ngrok`). It's a simple change. Let's test it on RC.

### Screenshots
Slack:
image
image
WhatsApp:
image


ibrahimjaved12:
This is mostly ready but I just need to confirm the image dimensions requirements.

ibrahimjaved12:
The current resolution is 1600 x 630, which comes to about 2.5 aspect ratio. We're good to go with this.

umar8hassan (approved):
Looks good. tested that logo is appearing in slack chat
Screenshot 2024-08-07 at 2 27 15 PM

Issue #1210, opened 2023-07-18
Change links to direct page, not to redirect


ChristopherChudzicki:
Several of our links actually link to a redirect:
- OCW main page Search links to `https://ocw.mit.edu/search?q=physics`, then redirects to `https://ocw.mit.edu/search/?q=physics` (slash after `search/`)
- Similar, course cards on the home page are missing trailing slash in their links, resulting in a redirect

It's unclear to me whether this affects the metrics that google uses, but certainly it would be a better user experience (on slow connections) to link directly to relevant pages.

PR #1437, opened 2024-08-09
Updated course card links to have trailing "/"

umar8hassan:
fixes https://github.com/mitodl/ocw-hugo-themes/issues/1210

Add trailing slash "/" to search and course links.

### How can this be tested?
- Inspect the course link in card at homepage, it should have trailing "/".

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ at the RC instances and temporarily disable CORS in your browser.
 
 ### CORS
 
-The search page at `/search` uses the `open-discussions` search API to source
+The search page at `/search/` uses the `open-discussions` search API to source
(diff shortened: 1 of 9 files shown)


ibrahimjaved12 (changes requested):
The functionality works as intended. However, I think we have to create the same change in a few more places.

If you run `yarn start course` and click the search icon on the top, it goes to `http://localhost:3000/search` which should've been `http://localhost:3000/search/`.

Looking at the codebase, there are 4 places where we are using:
`<a href="/search">`

This includes the Explore button as well.

image

image

We should investigate those tags in the codebase and fix them.

ibrahimjaved12 (approved):
Looks good to me 👍🏻

PR #1451, opened 2024-10-08
Remove the fallback logic of backup_url

ibrahimjaved12:
Refer to this comment: https://github.com/mitodl/hq/issues/5741#issuecomment-2402458356

This PR removes the fallback logic of `backup_url` if the ER is broken. The purpose is to let course authors decide themselves whether to use the (now) `wayback_url` or not directly from the `Add ER interface in Studio`.

### How can this be tested?
1. Make sure the ER's are working fine.


umar8hassan (approved):
Looks good

PR #1469, opened 2024-11-19
Fix issues with Mailchimp newsletter signup

pt2302:
Closes https://github.com/mitodl/hq/issues/5753.

This PR fixes two bugs with newsletter sign up. The first currently makes it impossible to sign up, due to a change in the Mailchimp form fields. The second is an issue with redirecting from the main OCW page to newsletter signup.

### How can this be tested?
Run `yarn start www`, and then navigate to `http://localhost:3000/newsletter`. Fill in the subscription form and verify that the subscription request goes through with no errors via the browser console (a successful subscription request can be checked on Mailchimp as well). 

Now, navigate to `http://localhost:3000` and enter an email address in the `Sign Me Up` form in the footer. Verify that clicking the button and hitting `enter/return` both navigate to `http://localhost:3000/newsletter/?EMAIL=`.


ibrahimjaved12 (approved):
The navigation and the sign up functionality now work as expected.

I received this mail after signing up:
image

Looks good to me.

PR #1473, opened 2024-12-03
Fix external links on mobile

pt2302:
Closes https://github.com/mitodl/hq/issues/6180.

This PR fixes an issue with opening external links caused by two conflicting external link modals (for mobile and non-mobile).

### How can this be tested?
Run `yarn start course 8.02-spring-2019` (or any other course with external links in the menu). Verify that the links work properly in mobile view. This should be tested with multiple browsers, including Chrome.


ChristopherChudzicki (approved):
Looks good to me.

Tested on firefox/safari/chrome mobile views.

pdpinch:
Would it make sense to add a test?

PR #1475, opened 2024-12-06
Update download page text when no resources found

pt2302:
Closes https://github.com/mitodl/hq/issues/6189 and closes https://github.com/mitodl/hq/issues/6190.

This PR makes two changes to the Course Download page when no resources are found for a course:
1. The `No Resources Found.` message is not displayed.
2. The message next to the `Download course` button is updated to the following:

**This package contains the same content as the online version of the course. Once downloaded, to open the homepage, click on the index.html file. For more help using these materials, read our [FAQs](https://mitocw.zendesk.com/hc/en-us/articles/4414681093659-I-have-downloaded-an-MIT-OpenCourseWare-course-but-I-can-t-access-the-materials-How-do-I-get-started-).**

### How can this be tested?

Run `yarn start course <course name>` for a course with resources, such as `21w.758-spring-2013` and then repeat for a course without resources, such as `10.01-spring-2020`. Confirm that the Download Course page at `localhost:3000/download` has been updated for courses with no resources.

### DIFF
--- a/course-v2/layouts/partials/resources_header.html
+++ b/course-v2/layouts/partials/resources_header.html
@@ -3,9 +3,7 @@
 {{- $downloadCourseUrl := partial "get_course_download_url.html" . -}}
 {{- $noResourcesFound := partial "show_no_resources_found.html" (dict "context" . "taxonomy" "learning_resource_types") -}}
 {{- $containsVideos := partial "contains_videos.html" (dict "context" . "taxonomy" "learning_resource_types") -}}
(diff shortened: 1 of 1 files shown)


pdpinch (approved):
TBH, I find this UX frustrating. I click download in the sidebar and then I have to click download again. The "download page" serves no value. 

But, since this is already done and it is an incremental improvement, I guess we should go ahead an merge it. But to do things right, we should rethink this UX.

PR #1505, opened 2025-02-03
Fix: Resource List Download Icon not visible

ibrahimjaved12:
None

We recently changed the way params were handled in the resource list partials for this issue: https://github.com/mitodl/hq/issues/5111

As a side effect, there was an inconsistency in the params passing, due to which `$downloadLink` comes out empty and no download icon appears for resource items in resource list view. There is a case sensitivity: `params` is being passed, but `Params` is being expected.

This PR changes `params` to `Params` making sure it is consistent across different partials.

Before:
image

After:
image

### How can this be tested?
1. `yarn start course 18.06sc-fall-2011`
2. Go to http://localhost:3000/download
3. Make sure the download icon appears and works as expected! (in both list view, and individual resource page)


umar8hassan (approved):
Looks good 👍

PR #1520, opened 2025-02-18
 Improve accessibility for external links

ibrahimjaved12:
Closes https://github.com/mitodl/hq/issues/6480

The voiced text is now consistently `opens in a new tab` irrespective of whether the ER opens the warning dialog or not.

### How can this be tested?
If you have Mac, please use its VoiceOver. Otherwise only inspect the aria-label attributes in `anchor` tags.
1. Checkout to this branch.
2. `yarn start course peter-test-2024-02-20`
3. Inspect the external links in the left navbar.

### DIFF
--- a/base-theme/layouts/partials/external_resource_link.html
+++ b/base-theme/layouts/partials/external_resource_link.html
@@ -3,14 +3,11 @@
 {{- $hasWarning := default true .has_external_license_warning -}}
 {{- $className := printf "%s %s" "external-link" (default "" .class) -}}
 {{- $onClick := "" -}}
-{{- $ariaLabel := "" -}}
+{{- $ariaLabel := printf "%s (opens in a new tab)" $text -}}
 {{- if $hasWarning -}}
   {{- $className = printf "%s %s" "external-link-warning" $className -}}
-  {{- $ariaLabel = printf "%s (opens warning dialog)" $text -}}
   {{/* Prevents external links from being clicked before corresponding JS is fully loaded. */}}
   {{- $onClick = "event.preventDefault()" -}}
-{{- else -}}
-  {{- $ariaLabel = printf "%s (opens in a new tab)" $text -}}
 {{- end -}}
 {{- if not (in $href "ocw.mit.edu") -}}
   {{- partial "link" (dict
--- a/base-theme/layouts/partials/external_resource_link.html
+++ b/base-theme/layouts/partials/external_resource_link.html
@@ -3,7 +3,8 @@
 {{- $hasWarning := default true .has_external_license_warning -}}
 {{- $className := printf "%s %s" "external-link" (default "" .class) -}}
 {{- $onClick := "" -}}
-{{- $ariaLabel := printf "%s (opens in a new tab)" $text -}}
+{{- $ariaLabel := $text | markdownify | plainify -}}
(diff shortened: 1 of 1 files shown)


pdpinch:
> I have provided a minor suggestion to improve code clarity.

Where? 

> The approach of setting the aria-label to "opens in a new tab" regardless of the warning dialog status is a good way to ensure a consistent user experience for screen reader users.

Do you think Gemini figured this out from the code only, or it is relying on Ibrahim's PR description?

ibrahimjaved12:
Where is the code review? What did you suggest? @gemini-code-assist

Please mention the change(s) explicitly.

ibrahimjaved12:
@gemini-code-assist This is already implemented in the PR, no.. ?

umar8hassan (commented):
@ibrahimjaved12 as discussed, In inspect i can see that the label is there.
Voiceover, does say `opens in a new tab` for an external link when I visit a page and leave voiceover to iterate over all of the text components. But if i navigate to an external link in navmenu with the keyboard, it only says `You are on a selectable text`.

I am not familiar with using Voiceover, but i think it'd be desirable to have the Voiceover say the same thing when we navigate with the keyboard. @ibrahimjaved12 is it possible?

@pdpinch we could use your opinion on this.

ibrahimjaved12:
@umar8hassan Hmm, I'm unable to replicate that on Chrome. I think that is more about how screen readers work, and how they may work with different browsers?
In essence, I normally just let it run through the page by itself, or navigate using tab.

But I also just used keyboard arrow keys (left and right) to navigate, and I got the same output.

pdpinch:
What do we need to do to resolve this and get this merged?

ibrahimjaved12:
@ChristopherChudzicki Could you please take a look at this PR and this comment: https://github.com/mitodl/ocw-hugo-themes/pull/1520#pullrequestreview-2644732910
This is the ticket context: https://github.com/mitodl/hq/issues/6480#issuecomment-2643707939
Basically it should always say in case of ERs: `(opens in a new tab)`

Can you please see if the current behavior is correct, because what I understand is VoiceOver acts slightly differently in different contexts- e.g., different browsers. It works fine for me in Chrome.

ChristopherChudzicki:
> as discussed, In inspect i can see that the label is there.
Voiceover, does say opens in a new tab for an external link when I visit a page and leave voiceover to iterate over all of the text components. But if i navigate to an external link in navmenu with the keyboard, it only says You are on a selectable text.

I just tried this in Chrome and Firefox, and voiceover always seems to read "Link Text (opens in a new tab)", whether I...
- navigate via tab
- navigate via keyboard arrows
- read the whole page 

This is what I would expect. So 👍 👍 

That said, I did notice one quirk. In Peter's test course, the page `/pages/this-is-a-page/` contains a link with bold text. The rendered aria-label is currently `Li***nk to P***BS (opens in a new tab)`, so voiceover reads "L i star star n k to P star star B S". I think this could be addressed via
```diff
-{{- $ariaLabel := printf "%s (opens in a new tab)" $text -}}
+{{- $ariaLabel := $text | markdownify | plainify -}}
+{{- $ariaLabel := printf "%s (opens in a new tab)" $ariaLabel -}}
```

umar8hassan (approved):
Approved in light of  Chris' Test results.

PR #1540, opened 2025-03-03
Add PostHog to the OCW public site

pt2302:
Closes https://github.com/mitodl/hq/issues/6689.

This PR adds PostHog integration to the OCW public site. Specifically, this PR adds the environment variables required for setting up PostHog, adds the PostHog library, initializing it for the online versions of `ocw-www` and `course-v2`, and adds tests for correct initialization and logging.

The capability for adding feature flags is also added in this PR. For any feature flags in the future, this can be added to the appropriate module via something like this

```
import { isFeatureEnabled } from './posthog'

if (isFeatureEnabled('show-learn-login')) {
 // show login
} else {
  // don't show login
}
```

### How can this be tested?
Set up PostHog using the new environment variables defined in the README. Set `POSTHOG_API_HOST`, `POSTHOG_ENABLED`, `POSTHOG_ENV`, and `POSTHOG_PROJECT_API_KEY` (obtaining the API key from the PostHog dashboard). Then run `yarn start www` and `yarn start course <course-id>` and verify that events like page views appear in the PostHog logs in the PostHog dashboard.

### DIFF
--- a/base-theme/assets/index.ts
+++ b/base-theme/assets/index.ts
@@ -11,15 +11,12 @@ import "video.js/dist/video-js.css"
 import { initExternalLinkModal } from "./js/external_link_modal"
 import { initQuotesModalHandler } from "./quote_card_modal_handler"
 import { initVideoDownloadPopup } from "./js/video_download_popup"
-import posthog from "posthog-js"
-
 export interface OCWWindow extends Window {
   $: JQueryStatic
   jQuery: JQueryStatic
   Popper: typeof Popper
   Sentry: typeof Sentry
   PDFObject: typeof PDFObject
-  posthog: typeof posthog
   initVideoJS: () => void
   initModalHandler: () => void
 }


base-theme/assets/index.ts
```diff
@@ -11,13 +11,15 @@ import "video.js/dist/video-js.css"
 import { initExternalLinkModal } from "./js/external_link_modal"
 import { initQuotesModalHandler } from "./quote_card_modal_handler"
 import { initVideoDownloadPopup } from "./js/video_download_popup"
+import posthog from "posthog-js"
```
gumaerc, line 14: I don't think we need any changes to this file now that www and course are initializing Posthog on their own. I tested removing these lines and everything still worked.

gumaerc (changes requested):
Everything seems to work great, just one thing I noticed:

gumaerc (approved):
👍

PR #1543, opened 2025-03-07
Replace headless:true with never render

ibrahimjaved12:
https://github.com/mitodl/hq/issues/6895

Removes `headless: true` as its behavior is now changed in Hugo in case of iterations. Look at the ticket for more details.

### How can this be tested?
1. Checkout to this branch
2. `yarn start www`
3. Go to About page... make sure `OCW Staff` and `Milestones` is visible in all screen sizes.


umar8hassan (approved):
Looks Good 👍

PR #1584, opened 2025-04-30
Adds new Learning Resource Types to Search

ibrahimjaved12:
https://github.com/mitodl/hq/issues/7140

Adds new Learning Resource Types to Search to make selection possible:
Editable files
Student work (course level only)
Podcasts

### How can this be tested?
Although this should have worked with the following steps, but it's not working for me right now:

* Checkout to this branch
* Get a fresh copy of `ocw-www` site. Just delete your existing one and run `yarn start www` it will prompt you to download, press `y` and enter.

* You will have to run an instance of browser with `CORS` disabled. If you're using Mac and have Google Chrome installed, run in terminal: 

`open -n -a /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --args --user-data-dir="/tmp/chrome_dev_test" --disable-web-security`

* Go to Search: http://localhost:3000/search/?type=course
Here under filters, you should see: `Podcasts`, `Student work`, These are course-level filters.
And then go to resources and check Resource Types filters:
image

The new Learning Resource Types should be clickable/selectable. 
Note: Also they're currently appearing in search because I had added them in my test course in RC: ibrahims-cat-course, so they have 1 instance from there.

What I propose is, we do a code-review, deploy it to RC and test there. If it doesn't work, then we can debug why it's not working.
This was the last PR of this kind: https://github.com/mitodl/ocw-hugo-themes/pull/1394


www/assets/js/lib/constants.ts
```diff
@@ -118,7 +119,9 @@ const RESOURCE_TYPES = [
   "Recitation Videos",
   "Recitations",
   "Simulation Videos",
+  "Editable files",
```
abeglova, line 122: Should this be "Editable Files" and  "Student Work" to match all the other resource types? I guess you will need to edit the tag in ocw-studio as well

abeglova (approved):
This works but the case should match other resource types

www/assets/js/lib/constants.ts
pdpinch: Thanks for catching this @abeglova. I agree, these should have initial caps for consistency. 

```suggestion
  "Editable Files",
```

And we do need to change this in ocw-studio as well. Hopefully we don't have too much data yet. 😳

www/assets/js/lib/constants.ts
```diff
@@ -118,7 +119,9 @@ const RESOURCE_TYPES = [
...
   "Simulations",
+  "Student work",
```
pdpinch, line 124:
```suggestion
  "Student Work",
```

pdpinch (changes requested):

PR #1602, opened 2025-06-19
fix: updated thumbnail url with content type

umar8hassan:
https://github.com/mitodl/hq/issues/7499

The search functionality referenced image thumbnails using `/images/` paths, which didn't align with the webpack build output structure.

Updated image paths in search code from `/images/` to `/static_shared/images/` to match the actual webpack output directory structure.

### How can this be tested?
This would be tested best on RC, because the search doesn't work properly in the Local setup.

### DIFF
--- a/www/assets/js/lib/constants.ts
+++ b/www/assets/js/lib/constants.ts
@@ -479,4 +479,4 @@ export const FACET_OPTIONS: Facets = {
   resource_type:       RESOURCE_TYPES
 }
 
-export const STATIC_THUMBNAIL_PATH_PREFIX = "/static_shared/images/"
+export const STATIC_THUMBNAIL_PATH_PREFIX = "/static_shared/images"


www/assets/js/lib/search.ts
```diff
@@ -152,7 +153,7 @@ export const getCoverImageUrl = (result: LearningResource) => {
       )}`
     }
   } else {
-    return `/images/${result.content_type}_thumbnail.png`
+    return `${STATIC_THUMBNAIL_PATH_PREFIX}/${result.content_type}_thumbnail.png`
```
ibrahimjaved12, line 156:
```suggestion
    return `${STATIC_THUMBNAIL_PATH_PREFIX}${result.content_type}_thumbnail.png`
```

www/assets/js/lib/search.test.tsx
```diff
@@ -36,7 +37,7 @@ describe("search library", () => {
         it(`should return correct image for result w/content type ${contentType}, image_src ${result.image_src}`, () => {
           const expectedSrc = hasImageSrc ?
             result.image_src :
-            `/images/${result.content_type}_thumbnail.png`
+            `${STATIC_THUMBNAIL_PATH_PREFIX}/${result.content_type}_thumbnail.png`
```
ibrahimjaved12, line 40:
```suggestion
            `${STATIC_THUMBNAIL_PATH_PREFIX}${result.content_type}_thumbnail.png`
```

PR #1606, opened 2025-07-02
Fix URLs for SEO in Offline Build

pt2302:
Closes https://github.com/mitodl/hq/issues/7693.

This PR fixes an issue where the downloadable versions of courses (offline build) display incorrect URLs for `og:url` and `canonical`, caused by combining two full URLs in both cases. This is fixed by using relative URLs instead.

### How can this be tested?
Make sure that there is a local copy of the course content for a given course, e.g., by running `yarn start course <course-id>`, which will automatically clone the relevant course content repo. Then, run `yarn build <path to local course repos>/<course id> <path to ocw-hugo-projects>/ocw-course-v2/config-offline.yaml`. Examine the source of any HTML page in the `<path to local course repos>/<course id>/dist` directory, and ensure that the URLs in `og:url` and `canonical` look correct. Verify that these URLs look correct in the online build as well.

### DIFF
--- a/base-theme/layouts/partials/seo.html
+++ b/base-theme/layouts/partials/seo.html
@@ -19,13 +19,13 @@
   {{/*  Facebook Open Graph  */}}
   <meta property="og:site_name" content="MIT OpenCourseWare">
   <meta property="og:type" content="website" />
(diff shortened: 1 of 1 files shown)


umar8hassan (approved):
Looks Good 👍

PR #1607, opened 2025-07-03
chore: added youtube player for chalk radio on `about us` page

umar8hassan:
https://github.com/mitodl/hq/issues/6943

This PR adds a YouTube player for Chalk Radio, replacing the Simplecast player on the `About Us` page on the OCW home.

### How can this be tested?
1. Switch to this branch and run the homepage with: `yarn start www`
2. Go to `http://localhost:3000/about/` and scroll half way down to verify that it has YouTube player for Chalk radio as shown in the screenshot above
3. Play the video to verify that it works

### DIFF
--- a/www/layouts/about/section.html
+++ b/www/layouts/about/section.html
@@ -339,7 +339,7 @@
               behind some of MIT’s most interesting courses, from nuclear physics to film appreciation
               to ethical responsibilities of computing.
             </p>
-            {{ partial "simplecast.html" "a9349695-05d2-4150-bba6-271cbf875ba0" }}
+            {{ partial "youtube_player.html" (dict "youtubeKey" "8kFFg5jAoQc" "startTime" "" "endTime" "" "downloadLink" "" "transcriptLink" "" "captionsLocation" "" "isOffline" false) }}
           </li>
           <li>
             <span class="font-weight-bold">Open Textbooks</span>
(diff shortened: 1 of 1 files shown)


ibrahimjaved12 (changes requested):
Could you please attach transcript and captions with the video.

Also, could you please define them, and the video data inside variables and use them to make the code more general-purpose? Maybe something like `$featured_video`?

These are the files- fetched from 3play:
[Transcript and Captions.zip](https://github.com/user-attachments/files/21117544/Transcript.and.Captions.zip)

umar8hassan:
> Could you please attach transcript and captions with the video.
> 
> Also, could you please define them, and the video data inside variables and use them to make the code more general-purpose? Maybe something like `$featured_video`?
> 
> These are the files- fetched from 3play: [Transcript and Captions.zip](https://github.com/user-attachments/files/21117544/Transcript.and.Captions.zip)

I have updated the Video ID to the latest video in the playlist.

ibrahimjaved12 (changes requested):
Could we use an `iframe` here instead?

Since we are not leveraging the videojs features that we usually use it for (themes, captions, transcripts etc.) we are essentially using it here like an `iframe`. But when we load the About Us page using `videojs`, 6mb of additional imports are downloaded.

image

ibrahimjaved12:
I think this size is too small?
image

pdpinch:
Good to know the footprint of video.js. I'm fine with saving some bytes, but remember that there is a maintenance overhead for special cases. Hopefully, this iframe approach is simpler to maintain too.

PR #1644, opened 2025-09-22
fix: accessibility issues

zawan-ila:
https://github.com/mitodl/hq/issues/8512
https://github.com/mitodl/hq/issues/8641

This PR makes the following changes:-
1. Change aria-label of the Give Now button on the Homepage(HP) to `Give Now-Donate to MIT open course ware` to ensure that the aria-label contains the visible text
2. Removes the hover state on LRT pills on the Course home page (CHP)
3. Makes focus more prominent for the Search and Explore buttons on the home page
4. Adds focus state for the share your story button.

### How can this be tested?
For both the offline and online builds of the Homepage and Course Home Page, verify that the changes mentioned in the description above have been correctly implemented.

### DIFF
--- a/www/assets/css/home.scss
+++ b/www/assets/css/home.scss
@@ -250,7 +250,7 @@ $searchbox-height: 50px;
 
 .home-search-box button.submit:focus,
 .btn.explore:focus {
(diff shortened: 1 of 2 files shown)


umar8hassan (approved):
Looks Good 👍

PR #1743, opened 2026-02-23
feat: Lighthouse Green performance and best practices

umar8hassan:
https://github.com/mitodl/hq/issues/10183

This is in follow-up to https://github.com/mitodl/ocw-hugo-themes/pull/1742 and aims to improve the Lighthouse score for performance and best practices.

### Netlify
https://ocw-hugo-themes-course-v3-pr-1743--ocw-next.netlify.app/

### How can this be tested?
run lighthouse using Chrome devtools and verify that score for each mentioned pages on the acttached are green


zawan-ila (changes requested):

zawan-ila:
@umar8hassan It seems that a number of my earlier suggestions are still pending. Footer is using the mit-red logo, instead of the black one. The hover over `Download Course` is broken. The og:image tags are using explicit dimensions when we do not force any particular dimensions in studio.

You have also undone some changes from https://github.com/mitodl/ocw-hugo-themes/pull/1741, which cause a regression on that front.

zawan-ila:
@umar8hassan I don't see any updates since my last review. I believe that you have swapped the labels accidentally. Please let me know if that is not the case.

zawan-ila (changes requested):
Mostly looks good. 

1. I think we should match the styling of the mit logo in the footer with learn. Currently, it appears to be much smaller.
2. The heading level changes make the headings on the course home page much darker (see image below) than they were earlier. We should try to preserve the styles, and if that is infeasible, make a note of it in the description and revisit it afterwards, preferably with input from design.

---------
Image

zawan-ila (approved):
Please remove the Appzi script changes. We should also load it locally so the environment matches production as closely as possible.

Removing it from the local environment may silence Lighthouse locally, but it won’t reflect what happens in production.

PR #1791, opened 2026-04-15
feat: added image picture partial and migrated existing images to it

umar8hassan:
https://github.com/mitodl/hq/issues/10884

Added picture partial for Fastly based optimization with original-image fallback behavior.

### Netlify
https://ocw-hugo-themes-course-v3-pr-1791--ocw-next.netlify.app/

### How can this be tested?
1. Open the Netlify preview in browser: DevTools → **Network** tab → filter **Img**.
2. On the course homepage, the course hero image request URL should contain `format=webp&quality=60&width=<breakpoint>`.
3. On a resource image page (e.g. `/resources/15-s21iap14/`), the displayed image should similarly use the WebP srcset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ Use the shared `picture_element.html` partial for content images that should be 
 ) }}
 ```
 
-The partial renders WebP `srcset` candidates at `480w`, `800w`, `1280w`, and `1920w`, with the original image URL kept on the fallback `<img>`. Root-relative and absolute URLs are optimized; relative paths are emitted as a plain `<img>` so offline bundles continue to use original local assets. Fastly IO currently has WebP enabled for OCW; add other formats only after the service defaults support them.
+The partial renders optimized `srcset` candidates at `480w`, `800w`, `1280w`, `1920w`, `2560w`, and `3840w` using `format=auto&quality=75`. `format=auto` lets Fastly content-negotiate the best format (WebP, AVIF, or original) based on the browser's `Accept` header — no explicit format needs to be hard-coded. The extended breakpoints cover 2× and 3× retina displays. Root-relative and absolute URLs are optimized; relative paths (used in offline builds) are emitted as a plain `<img>` unchanged.
 
 Choose `sizes` based on the rendered slot width, not the source image width. Use `100vw` for full-width content images, `50vw` for two-column desktop layouts with a `100vw` mobile fallback, and fixed values such as `200px` only when the component has a fixed image slot.
 
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ Use the shared `picture_element.html` partial for content images that should be 
 ) }}
 ```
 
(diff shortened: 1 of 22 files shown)


zawan-ila (changes requested):
Mostly looks good. A couple comments:-

1. We should add support in the picture_element partial for devices with different pixel intensities. This is noted in the RFC as well as an AC in the linked ticket
2. The quality being set to 60 seems a bit low to me. By default, fastly appears to use 85. What do you think about setting it to 75? We can always tune it later if need be.

zawan-ila (changes requested):
@umar8hassan I initially suggested handling the device resolutions explicitly because I thought it was necessary, but browsers already take care of this automatically through the srcset and sizes attributes. Given that, I think we should revert the picture_element changes introduced in https://github.com/mitodl/ocw-hugo-themes/pull/1791/changes/2a2946c35982b1beba8ef40789a4d4a819133984 since the previous implementation was simpler and already covered the required use cases.

Secondly, this breaks some styles e.g check out the president message section of the ocw [about](https://ocw-hugo-themes-www-pr-1791--ocw-next.netlify.app/about/) page. I suspect this may be due to the markup structure changing from img to picture > img, causing some existing selectors/styles not to apply as expected. If that assumption is correct, one possible solution would be to avoid using the picture element entirely and instead apply srcset and sizes directly on the img element, since both attributes are natively supported there as well.

zawan-ila (commented):
@umar8hassan Please take a look at the second paragraph from my [review](https://github.com/mitodl/ocw-hugo-themes/pull/1791#pullrequestreview-4249969656) above. As I note, this introduces some styling problems. For an example, please look at the president's message in the netlify about [page](https://ocw-hugo-themes-www-pr-1791--ocw-next.netlify.app/about/) build and compare it to the live about page.

zawan-ila (changes requested):
Surprisingly, this still seems to have some styling issues. As noted in the testing instructions, go to the [resource page](https://ocw-hugo-themes-course-v3-pr-1791--ocw-next.netlify.app/resources/15-s21iap14/). Notice how the image is much smaller than it is on main.

zawan-ila (commented):
Can you please elaborate why we remove the srcset from the images in the most recent commit? The goal of the PR is to add srcsets so that we load appropriately sized images on different screen sizes? Is it not possible to fix the UI issues while preserving the srcset?

umar8hassan:
> Can you please elaborate why we remove the srcset from the images in the most recent commit? The goal of the PR is to add srcsets so that we load appropriately sized images on different screen sizes? Is it not possible to fix the UI issues while preserving the srcset?

I had reverted it to facilitate the different sized images, like on resource pages, without any size indication which defaulted to `100vw` in picture partial. This resulted into browser intrinsically correcting density, thus causing different images than their natural size.

zawan-ila (commented):
I think we should limit the width to 1920 pixels. I doubt that most of our images are large enough to support 2560 or 3840 pixels effectively, and the RFC also specifies a maximum width of 1920 pixels.

zawan-ila (approved):

PR #1792, opened 2026-04-15
feat: updated canonical url for course-v3 and added noindex setting

umar8hassan:
https://github.com/mitodl/hq/issues/10893

Added canonical url and no-indexing settings for course-v3 theme

### Netlify
TBA

### How can this be tested?
TBA

### DIFF
--- a/base-theme/layouts/partials/head.html
+++ b/base-theme/layouts/partials/head.html
@@ -1,5 +1,5 @@
 {{- $gtmId := getenv "GTM_ACCOUNT_ID" -}}
-{{- $noindex := getenv "NOINDEX" -}}
+{{- $noindex := partial "noindex.html" . -}}
 
 <head>
   {{ if $gtmId }}
@@ -12,7 +12,7 @@
     <!-- End Google Tag Manager -->
   {{ end }}
   {{ block "extrahead" . }}{{ end }}
-  {{ if eq $noindex "true" }}
+  {{ if $noindex }}
   <meta name="robots" content="noindex">
   {{ end }}
   <meta charset="utf-8">
--- a/base-theme/layouts/partials/head.html
+++ b/base-theme/layouts/partials/head.html
@@ -1,5 +1,5 @@
 {{- $gtmId := getenv "GTM_ACCOUNT_ID" -}}
-{{- $noindex := partial "noindex.html" . -}}
+{{- $noindex := getenv "NOINDEX" -}}
 
 <head>
   {{ if $gtmId }}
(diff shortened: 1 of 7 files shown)


zawan-ila (changes requested):
We have already got the "NOINDEX" env variable. Can't we leverage that for v3 too, instead of introducing a "COURSE_V3_NOINDEX". Even currently, we are leveraging it on  https://live-qa.ocw.mit.edu/ocw-course-v3/courses/mas-836-sensor-technologies-for-interactive-environments-spring-2011/ to set the pages to noindex

For the canonical url, we may also be able to leverage existing functionality (SITEMAP_DOMAIN is used for the current v2 theme) instead of introducing newer variables.

umar8hassan:
> We have already got the "NOINDEX" env variable. Can't we leverage that for v3 too, instead of introducing a "COURSE_V3_NOINDEX". Even currently, we are leveraging it on https://live-qa.ocw.mit.edu/ocw-course-v3/courses/mas-836-sensor-technologies-for-interactive-environments-spring-2011/ to set the pages to noindex
> 
> For the canonical url, we may also be able to leverage existing functionality (SITEMAP_DOMAIN is used for the current v2 theme) instead of introducing newer variables.

`NOINDEX` is being used in course-v2 and we shouldn't change its value. same goes for `SITEMAP_DOMAIN`.

zawan-ila:
> NOINDEX is being used in course-v2 and we shouldn't change its value. same goes for SITEMAP_DOMAIN

@umar8hassan These are env variables and can be set to different values for different themes in the relevant concourse pipelines.

The current variables that you've added `COURSE_v3_NOINDEX` will need to be added in the environment of the concourse jobs building the sites (unless you only want to use the defaults). What I am suggesting is that instead of these new vars, you specify the cureently existing ones (e.g `NOINDEX`) with the appropriate values across themes (v2/v3). That way we do not end up with a bunch of extra variables that do almost the same thing.

zawan-ila (changes requested):
The canonical url calculation is incorrect, primarily because you use short_id instead of, say, something like the site_url_path.

That said, once https://github.com/mitodl/ocw-studio/pull/3032 is merged, we should simply be able to use `.Relpermalink` and it should all work out as expected. The only change required would be swapping out `$sitemapDomain` for `COURSE_V3_CANONCICAL_DOMAIN` in the course-v2 [partial](https://github.com/mitodl/ocw-hugo-themes/blob/main/base-theme/layouts/partials/get_canonical_url.html).

I'd suggest making the change above once https://github.com/mitodl/ocw-studio/pull/3032 is merged.

PR #1797, opened 2026-05-06
fix: issues with learn-deployed v3 theme

zawan-ila:
https://github.com/mitodl/hq/issues/11124

This PR introduces two related improvements:

1. **_Update get_download_url for ZIP packages_**
Refactors the get_download_url logic to support the Learn deployment of Course V3 under the /course/o/... path.

2. **_Fix resource URL handling for subdirectory deployments_**
Resource URLs are currently generated as absolute paths relative to the domain root. While this works for root-level deployments (e.g., /courses/12-108-fall-2004), it breaks when the site is hosted under a subdirectory (e.g., /ocw-course-v3/courses/... or /courses/o/...). This PR updates the resource URL partial to respect the site’s base URL, ensuring assets are correctly resolved in both root and subdirectory deployments.
### How can this be tested?
This is kind of ugly since we need a way to emulate the learn [deploy](https://rc.learn.mit.edu/courses/o/7-016-introductory-biology-fall-2018) somehow. Do let me know if you know of a cleaner way.

1. First, apply this [patch](https://gist.github.com/zawan-ila/2ef4e71c1c82bb95b756a9d2ef881e50). This ensures that the webpack assets load correctly when we deploy to something like `localhost:8000/courses/o/6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016`. Secondly, it provides a way to pass the baseURL argument to the build.
2. Now run a course build with `yarn build <path_to_course_content> <path_to_v3_projects_config> <baseURL(e.g /courses/o/6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016)>`
3. Now the build output is in the dist folder of the content repo.
4. To the build output, add the website assets (images etc.). This can be done by downloading the offline zip package of the course from prod, and copy pasting the files from there.
4. Create a new directory `build` and inside it create the directory courses/o/6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016 `mkdir -p courses/o/6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016 `
5. Inside the `courses/o/6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016 ` folder, copy the output from the dist folder.
5. Inside the `build` directory, run a http server `python3 -m http.server`
6. Go to `localhost:8000/courses/o/6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016 `
7. Verify that all issues identified in the linked issue description are fixed and the acceptance criteria is fulfilled.
8. Also, run a quick check to ensure that all links, assets etc. work correctly, and there are no issues due to incorrect urls.

### DIFF
--- a/course-v3/layouts/partials/get_course_download_url.html
+++ b/course-v3/layouts/partials/get_course_download_url.html
@@ -1,4 +1,3 @@
-{{- $domain := strings.TrimSuffix "/" (getenv "STATIC_API_BASE_URL") -}}
 {{- $trimmedBaseUrl := strings.TrimPrefix "/" (strings.TrimSuffix "/" site.BaseURL) -}}
 {{- $shortId := site.Data.course.site_short_id -}}
-{{- return printf "%v/%v/%v.zip" $domain $trimmedBaseUrl $shortId -}}
+{{- return printf "/%v/%v.zip" $trimmedBaseUrl $shortId -}}
--- a/course-v3/layouts/partials/get_course_download_url.html
+++ b/course-v3/layouts/partials/get_course_download_url.html
@@ -1,3 +1,3 @@
-{{- $trimmedBaseUrl := strings.TrimPrefix "/" (strings.TrimSuffix "/" site.BaseURL) -}}
+{{- $trimmedBasePath := strings.TrimSuffix "/" (urls.Parse site.BaseURL).Path -}}
 {{- $shortId := site.Data.course.site_short_id -}}
-{{- return printf "/%v/%v.zip" $trimmedBaseUrl $shortId -}}
+{{- return printf "/%v/%v.zip" $trimmedBasePath $shortId -}}
--- a/course-v3/layouts/partials/get_course_download_url.html
+++ b/course-v3/layouts/partials/get_course_download_url.html
@@ -1,3 +1,3 @@
-{{- $trimmedBasePath := strings.TrimSuffix "/" (urls.Parse site.BaseURL).Path -}}
+{{- $trimmedBasePath := strings.TrimSuffix "/" (strings.TrimPrefix "/" (urls.Parse site.BaseURL).Path) -}}
 {{- $shortId := site.Data.course.site_short_id -}}
 {{- return printf "/%v/%v.zip" $trimmedBasePath $shortId -}}
--- a/course-v3/layouts/partials/resource_url.html
+++ b/course-v3/layouts/partials/resource_url.html
@@ -1,0 +1,17 @@
+{{- $url := .url -}}
+{{/*
+  Here we are disassembling the URL and returning a URL rooted at RESOURCE_BASE_URL.
+  Resource paths from Studio can include a leading course path segment
+  (for example, courses/foo), which is stripped before prefixing.
+  We also prepend Hugo's baseURL path segment before final URL construction.
+*/}}
+{{- if .url -}}
+  {{- $prefix := getenv "RESOURCE_BASE_URL" | default "" -}}
+  {{- $path := (urls.Parse .url).Path -}}
+  {{- $path = strings.TrimPrefix "/" $path -}}
+  {{- $path = replaceRE "^courses/[^/]+/?" "" $path -}}
+  {{- $basePath := (urls.Parse site.BaseURL) -}}
+  {{- $path = printf "%s/%s" (strings.TrimSuffix "/" $basePath.Path) (strings.TrimPrefix "/" $path) -}}
+  {{- $url = printf "%s/%s" (strings.TrimSuffix "/" $prefix) (strings.TrimPrefix "/" $path) -}}
+{{- end -}}
+{{- return $url -}}


course-v3/layouts/partials/get_course_download_url.html
```diff
@@ -1,4 +1,3 @@
-{{- $domain := strings.TrimSuffix "/" (getenv "STATIC_API_BASE_URL") -}}
 {{- $trimmedBaseUrl := strings.TrimPrefix "/" (strings.TrimSuffix "/" site.BaseURL) -}}
 {{- $shortId := site.Data.course.site_short_id -}}
-{{- return printf "%v/%v/%v.zip" $domain $trimmedBaseUrl $shortId -}}
\ No newline at end of file
+{{- return printf "/%v/%v.zip" $trimmedBaseUrl $shortId -}}
```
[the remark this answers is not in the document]
zawan-ila: Addressed in https://github.com/mitodl/ocw-hugo-themes/pull/1797/changes/f5d816cd60b794d72aa2b30e800ec074ae3c3f80

course-v3/layouts/partials/resource_url.html
```diff
@@ -0,0 +1,17 @@
...
+{{- if .url -}}
+  {{- $prefix := getenv "RESOURCE_BASE_URL" | default "" -}}
+  {{- $path := (urls.Parse .url).Path -}}
+  {{- $path = strings.TrimPrefix "/" $path -}}
+  {{- $path = replaceRE "^courses/[^/]+/?" "" $path -}}
+  {{- $basePath := (urls.Parse site.BaseURL) -}}
+  {{- $path = printf "%s/%s" (strings.TrimSuffix "/" $basePath.Path) (strings.TrimPrefix "/" $path) -}}
+  {{- $url = printf "%s/%s" (strings.TrimSuffix "/" $prefix) (strings.TrimPrefix "/" $path) -}}
+{{- end -}}
+{{- return $url -}}
```
pt2302, line 17: The partial unconditionally strips the original `courses/<slug>/` prefix from the resource path and replaces it with `site.BaseURL`'s path. This produces incorrect URLs when `RESOURCE_BASE_URL` points to a different location, because files on that host live at their original paths, not at the site's deployed path.

```suggestion
{{- $url := .url -}}
{{- if .url -}}
  {{- $prefix := getenv "RESOURCE_BASE_URL" | default "" -}}
  {{- $path := (urls.Parse .url).Path -}}
  {{- if eq $prefix "" -}}
    {{- $path = strings.TrimPrefix "/" $path -}}
    {{- $path = replaceRE "^courses/[^/]+/?" "" $path -}}
    {{- $basePath := (urls.Parse site.BaseURL).Path -}}
    {{- $url = printf "%s/%s" (strings.TrimSuffix "/" $basePath) (strings.TrimPrefix "/" $path) -}}
  {{- else -}}
    {{- $url = printf "%s%s" (strings.TrimSuffix "/" $prefix) $path -}}
  {{- end -}}
{{- end -}}
{{- return $url -}}
```

To reproduce the issue with the PR, run `yarn build:course-v3-prefixed 18.086-spring-2006 --base-url-prefix /ocw-course-v3/courses/o --serve`, and visit `http://localhost:8000/ocw-course-v3/courses/o/18-086-mathematical-methods-for-engineers-ii-spring-2006/resources/hw1_086_s06/`. The download link according to this partial resolves to `https://live-qa.ocw.mit.edu/ocw-course-v3/courses/o/7055eb8ec14e963aecb348c8246578a8_hw1_086_s06.pdf`, but it should be `https://live-qa.ocw.mit.edu/courses/18-086-mathematical-methods-for-engineers-ii-spring-2006/7055eb8ec14e963aecb348c8246578a8_hw1_086_s06.pdf`. Note that the partial is stripping out the actual course name from the URL, resulting in a broken link.

course-v3/layouts/partials/get_course_download_url.html
```diff
@@ -1,4 +1,3 @@
-{{- $domain := strings.TrimSuffix "/" (getenv "STATIC_API_BASE_URL") -}}
-{{- $trimmedBaseUrl := strings.TrimPrefix "/" (strings.TrimSuffix "/" site.BaseURL) -}}
+{{- $trimmedBasePath := strings.TrimSuffix "/" (urls.Parse site.BaseURL).Path -}}
 {{- $shortId := site.Data.course.site_short_id -}}
-{{- return printf "%v/%v/%v.zip" $domain $trimmedBaseUrl $shortId -}}
\ No newline at end of file
+{{- return printf "/%v/%v.zip" $trimmedBasePath $shortId -}}
```
pt2302, line 3:
```suggestion
{{- return printf "%v/%v.zip" $trimmedBasePath $shortId -}}
```

This would lead to a leading double-slash.

pt2302 (changes requested):
This PR doesn't appear to be working for deployments where `RESOURCE_BASE_URL` is non-empty; see the comments for details.

course-v3/layouts/partials/get_course_download_url.html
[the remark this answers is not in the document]
zawan-ila: This has been addressed in https://github.com/mitodl/ocw-hugo-themes/pull/1797/changes/ac67b8dc3bb4670a4c8d13d466f4c4ee15111692 The earlier fix was incorrect

course-v3/layouts/partials/get_course_download_url.html
zawan-ila: addressed in https://github.com/mitodl/ocw-hugo-themes/pull/1797/changes/ac67b8dc3bb4670a4c8d13d466f4c4ee15111692

course-v3/layouts/partials/resource_url.html
zawan-ila: > This PR doesn't appear to be working for deployments where RESOURCE_BASE_URL is non-empty

Can you try with `rc.learn.mit.edu` as the RESOURCE_BASE_URL? I think we should keep the `RESOURCE_BASE_URL` field simple. It effectively acts as a prefix for generated paths, and the description in the env file defines it that way as well.

The proposed change would fix the case for `live-qa.ocw.mit.edu`, but it’s easy to imagine another deployment using a different path prefix and running into the same issue again. Since we can simply set it to `rc.learn.mit.edu`—which follows the v3 conventions and already works fine—I don’t think we should add unnecessary complexity.

As far as I know, the only real use case for `RESOURCE_BASE_URL` is local testing.

> Note that the partial is stripping out the actual course name from the URL, resulting in a broken link.

The partial is generating the url just fine. It is just that the url 301s to one without the course name. I strongly suspect this to be due to various Fastly VCL [configs](https://github.com/mitodl/ol-infrastructure/tree/main/src/ol_infrastructure/applications/ocw_site/snippets) we have for ocw, which I believe to be for historical reasons. Swapping the RESOURCE_BASE_URL to `rc.learn.mit.edu` should fix the broken urls you are seeing.

pt2302: Yes, setting `RESOURCE_BASE_URL=https://rc.learn.mit.edu` and running with `yarn build:course-v3-prefixed 18.086-spring-2006 --base-url-prefix /courses/o --serve ` works fine.

pt2302 (approved):
Looks good to me!

PR #1826, opened 2026-07-07
feat: line-based truncation for the course-v3 course description

zawan-ila:
https://github.com/mitodl/hq/issues/11894

Switches the course-v3 home page course description from character-based to line-based truncation.

**Before:** the template decided to collapse when the (link-stripped) markdown source exceeded 320 characters, then rendered two copies of the description — one cut with `truncate 320 ""` and one full, hidden with `d-none` — and JS swapped the two divs.

**After:**
- The description is rendered and clamped to 5 lines with `-webkit-line-clamp` (native ellipsis). The line count lives in one SCSS variable, `$course-description-max-lines`.
- JS (`initCourseDescriptionExpander`) shows the Show more / Show less toggle only when the text actually overflows the clamp (`scrollHeight > clientHeight`), and re-checks via `ResizeObserver` as the viewport changes, so the button appears/disappears as the text re-wraps.
- The toggle updates `aria-expanded`, and the button is rendered with `aria-controls` pointing at the description.
- Hiding the button swaps `d-flex` together with `d-none`

See the netlify v3 [build](https://ocw-hugo-themes-course-v3-pr-1826--ocw-next.netlify.app/courses/o/15-s21-nuts-and-bolts-of-business-plans-january-iap-2014/)

### How can this be tested?
- Run a local build of any v3 course `yarn build:course-v3-prefixed <content-repo-directory> --serve`
- Verify that the course description is truncated based on the number of lines and shows an ellipsis. Toggling the show more button works as expected
- For the course you just built, in the content data.json file, change the description to something that'd fit in less than 5 lines. Build the course again and verify that the description renders properly and the show more button is hidden.

### Additional Context
- The value for $course-description-max-lines has been chosen somewhat arbitrarily and will be confirmed with appropriate stakeholders.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/course-v3/assets/css/course-home.scss
+++ b/course-v3/assets/css/course-home.scss
@@ -60,15 +60,10 @@
   }
 }
 
-#collapsed-description {
-  & p:last-of-type {
-    display: inline;
-  }
-}
+// Number of lines shown for the collapsed course description.
+$course-description-max-lines: 5;
 
-#collapsed-description,
-#expanded-description,
-#full-description {
+#course-description-text {
   color: #000;
 
   font-size: 0.875rem;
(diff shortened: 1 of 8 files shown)


umar8hassan (approved):
Looks Good 👍 

Tests may need some improvements. In `course_expander.ts`, typeof `ResizeObserver !== "undefined"` is always false in tests as jsdom does not implement `ResizeObserver`, meaning every test silently runs the else fallback (course_expander.ts:81-88). And even that fallback is never triggered or asserted (no test dispatches a resize event or checks the listener was registered)

PR #1830, opened 2026-07-14
feat: enable posthog event capture by default

zawan-ila:
## Related Tickets
https://github.com/mitodl/hq/issues/12194

## Description

Adds a `POSTHOG_UI_HOST` environment variable so PostHog's toolbar/dashboard links can resolve correctly even when `POSTHOG_API_HOST` points at a reverse proxy, and enables event capture by default.

## Testing instructions
-  Set the following variables in .env
  ```
  POSTHOG_ENABLED=true
  POSTHOG_PROJECT_API_KEY=<your_project_api_key>
  POSTHOG_API_HOST=https://ph.ol.mit.edu
  POSTHOG_UI_HOST=https://us.posthog.com
  ```
 - Run a v3 build yarn start course <content_repo> with the COURSE_HUGO_CONFIG_PATH pointing to v3
 - Visit the built site, navigate around and after a while verify from the posthog dashboard that pageview events are logged for your navigations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,6 @@ To further explain the various environment variables and what they do:
 | `POSTHOG_PROJECT_API_KEY` | `www`, `course` | `api-key` | API key for PostHog |
 | `POSTHOG_UI_HOST` | `www`, `course` | `https://us.posthog.com` | PostHog UI host, used when `POSTHOG_API_HOST` is a reverse proxy. When empty, PostHog falls back to the API host |
 | `POSTHOG_COURSE_V3_ENABLED` | `course` | `true` | Whether PostHog analytics are enabled for the `course` theme |
-| `POSTHOG_COURSE_V3_PROJECT_API_KEY` | `course` | `api-key` | API key for the PostHog project used by the `course` theme |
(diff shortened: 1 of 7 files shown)


pdpinch:
Do we need to support two posthog projects? 

I know we had added an ocw-only project to the v2 theme, but we never turned it on. Since we never used it, I think it would be okay to use the “learn” posthog project everywhere.

zawan-ila:
> I know we had added an ocw-only project to the v2 theme, but we never turned it on

@pdpinch but we [may intend to](https://github.com/mitodl/hq/issues/12184#issuecomment-4893685006)?

pdpinch:
Yes, I'd like to have a feature flag, but I think it could be in the Learn account.

zawan-ila:
@copilot remove the apikey setting from initposthog arguments. Remove POSTHOG_COURSE_V3_PROJECT_API_KEY entirely and use POSTHOG_PROJECT_API_KEY across themes as the key.

umar8hassan (approved):
Looks Good 👍

PR #1843, opened 2026-07-31
fix: skip caption/transcript partials when the value is not an array

umar8hassan:
https://github.com/mitodl/hq/issues/12635

Paired with mitodl/ocw-studio#3123. Either side stops the build failing on its own, and both are worth having: that one keeps the bad shape out of new frontmatter, this one covers content that is already published and would otherwise need a rebuild first.

`video_caption_tracks.html` and `video_transcript_links.html` assume their input is always an array of `{file, language}` objects. When a video has no caption or transcript linked, or its linked resource has no uploaded file, published frontmatter carries the CMS relation object `{"content": ..., "website": ...}` instead. Ranging a map yields its values rather than objects, so `.language` aborts the entire site build, not just the affected page:

```
video_caption_tracks.html:22:15: executing "partials/video_caption_tracks.html"
at <.language>: can't evaluate field language in type interface {}
```

- `base-theme/layouts/partials/video_caption_tracks.html`, `base-theme/layouts/partials/video_transcript_links.html`: skip a value that is not an array, and note in each why the check has to stay. These two are the only consumers that iterate the field, so this covers `video.html`, `video_embed.html`, the base-offline copy and `video_v3.html` through them.

### Netlify
[course-v2](https://ocw-hugo-themes-course-v2-pr-1843--ocw-next.netlify.app/)
[course-v3](https://ocw-hugo-themes-course-v3-pr-1843--ocw-next.netlify.app/)

### How can this be tested?

**Regression.** Open any video lecture on course-v2 or course-v3 above. Caption tracks and transcript download links should be unchanged, with English still selected by default and other languages still listed.

**The fix.** It is not reachable from the test course, whose video fixtures all carry the resolved array shape. To exercise it, render either partial with the relation-object shape:

```
{{ $bad := dict "content" (slice) "website" "somesite" }}
{{ $tracks := partial "video_caption_tracks.html" (dict "context" . "captionsFile" $bad) }}
```

Before this change that fails the build with the error above. After it, the partial returns an empty slice and the page renders with no caption tracks. Same for `video_transcript_links.html` with `transcriptFile`.

### DIFF
--- a/base-theme/layouts/partials/video_caption_tracks.html
+++ b/base-theme/layouts/partials/video_caption_tracks.html
@@ -14,8 +14,15 @@
 {{- $captionsFile := .captionsFile -}}
 {{- $tracks := slice -}}
 
-{{- if $captionsFile -}}
-  {{/* video_captions_resources is always an array of {file, language, locale?} objects */}}
+{{/*
+  The slice check is load-bearing, do not remove it. When a video has no
+  caption linked, or its linked resource has no uploaded file, older content
+  carries the CMS relation object {"content": ..., "website": ...} here
+  instead of an array. Ranging a map yields its values rather than objects,
+  so `.language` below then aborts the whole site build with
+  "can't evaluate field language in type interface {}".
+*/}}
+{{- if and $captionsFile (reflect.IsSlice $captionsFile) -}}
   {{/* First pass: check whether any English track exists */}}
   {{- $hasEnglish := false -}}
   {{- range $captionsFile -}}
--- a/base-theme/layouts/partials/video_caption_tracks.html
+++ b/base-theme/layouts/partials/video_caption_tracks.html
@@ -6,7 +6,9 @@
 
   Input: dict with:
     - "context" : page context (for resource_url resolution)
-    - "captionsFile" : video_captions_resources, an array of {"file": ..., "language": ..., "locale"?: ...}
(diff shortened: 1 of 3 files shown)


umar8hassan:
Lighthouse and Netlify checks/deployments are failing with errors in ocw-www due to ongoing work on RC. But it should be ready for review. Meanwhile, I'll fix it.

zawan-ila (commented):
Since there are only 5 websites that have this problem (reading from the issue description), I think publishing them after https://github.com/mitodl/ocw-studio/pull/3123 should be enough, and we don't need this.

At the very least, I think we should remove the comments and the tests, since they won't really be relevant after those 5 sites are published. And would just be confusing from then on for anybody going through the code.

umar8hassan:
> Since there are only 5 websites that have this problem (reading from the issue description), I think publishing them after [mitodl/ocw-studio#3123](https://github.com/mitodl/ocw-studio/pull/3123) should be enough, and we don't need this.
> 
> At the very least, I think we should remove the comments and the tests, since they won't really be relevant after those 5 sites are published. And would just be confusing from then on for anybody going through the code.

The guard was removed in a recent PR, along with a data change in Studio, which led to this scenario and failed mass build. Having a test would make sure it doesn't happen again.

